PR #2, opened 2021-03-03
Quantum Kernel

attp:
Class (`quantum_kernel.py`) that calculates a quantum kernel for use in scikit-learn machine learning algorithms for classification (`qsvc.py`) and regression (`qsvr.py`) algorithms.

To do:
- [x] determine interface for `qsvc` and `qsvr`
- [x] write test cases for `qsvc` and `qsvr`
- [x] update tutorial


qiskit_machine_learning/algorithms/regressors/qsvr.py
```diff
@@ -0,0 +1,42 @@
...
+    r"""Quantum Support Vector Regressor.
+
+        **Example**
+        .. code-block::
+            qsvr = QSVR(feature_map=map, quantum_instance=backend)
+            qsvr.fit(sample_train,label_train)
+            qsvr.predict(sample_test)
+    """
+
+    def __init__(self,
```
woodsp-ibm, line 33: Similar comments to the QSVC in terms of docstrings, params and consistency with setter/getters etc

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,322 @@
...
+    to m-dimension space. :math:`\langle x, y \rangle` denotes the dot product.
+    Usually m is much larger than n.
+
+    The quantum kernel algorithm calculates a kernel matrix, given datapoints x and y and feature
+    map f, all of n dimension. This kernel matrix can then be used in classical machine learning
+    algorithms such as support vector classification, spectral clustering or ridge regression.
+    """
+
+    def __init__(self,
+                 feature_map: QuantumCircuit,
```
woodsp-ibm, line 51: Often we will default something like this (the feature_map in this case) to None and internally create a default. We state what the default is of course. If there is no useful./reasonable default then requiring one seems fine. But it also relates to the comments I made about SVC above.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,322 @@
...
+                 enforce_psd: bool = False,
+                 batch_size: int = 1000,
+                 quantum_instance: Optional[
+                     Union[QuantumInstance, BaseBackend, Backend]] = None) -> None:
+        """
+        Args
+            feature_map:       # parameterized circuit to be used as the feature map
```
woodsp-ibm, line 58: Why do we have comment markers (the #) in the descriptions of the params here. Normally it would be more like `feature_map: Parameterized circuit to be used.....` etc

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,322 @@
...
+    def feature_map(self) -> QuantumCircuit:
+        """ Returns feature map """
+        return self._feature_map
+
+    @feature_map.setter
+    def feature_map(self, feature_map: QuantumCircuit):
+        """ Sets feature map """
+        self._feature_map = feature_map
+        if not hasattr(feature_map, 'ordered_parameters'):
+            self.feature_map.ordered_parameters = sorted(feature_map.parameters,
```
woodsp-ibm, line 78: Parameters now have a fixed order in Terra's parameterized circuits such that I think we can drop the sorted here and rely on the sorting in parameterized circuits now - which is better. @Cryoris can correct me. Also the ordered_parameters field, as far as I am aware, has now been deprecated,

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,322 @@
...
+        if len(x) != self._feature_map.num_parameters:
+            raise ValueError("x and class feature map incompatible dimensions.\n" +
+                             "x has %s dimensions, but feature map has %s." %
+                             (len(x), self._feature_map.num_parameters))
+
+        q = QuantumRegister(self._feature_map.num_qubits, 'q')
+        c = ClassicalRegister(self._feature_map.num_qubits, 'c')
+        qc = QuantumCircuit(q, c)
+
+        x_dict = dict(zip(self._feature_map.ordered_parameters, x))
```
woodsp-ibm, line 129: See ordering comment above

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,322 @@
...
+
+        Raises:
+            AlgorithmError:
+                - A quantum instance or backend has not been provided to the class
+            ValueError:
+                - x_vec and/or y_vec are not one or two dimensional arrays
+                - x_vec and/or y_vec have incompatible dimension with feature map
+        """
+        if self._quantum_instance is None:
+            raise AlgorithmError("A QuantumInstance or Backend "
```
woodsp-ibm, line 191: There is a QiskitMachineLearningError - it may be preferable to use that instead of the Algorithm one

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,322 @@
...
+        if x_vec.ndim == 1:
+            x_vec = np.reshape(x_vec, (-1, 2))
+
+        if y_vec is not None and y_vec.ndim > 2:
+            raise ValueError("y_vec must be a 1D or 2D array")
+
+        if y_vec is not None and y_vec.ndim == 1:
+            y_vec = np.reshape(y_vec, (-1, 2))
+
+        # TODO: update number of qubits --> should allow setting the number of params too!
```
woodsp-ibm, line 213: Will you be addressing this TODO?

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,322 @@
...
+            result = results.get_counts(idx)
+
+            kernel_value = result.get(measurement_basis, 0) / sum(result.values())
+        return kernel_value
+
+    def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
+        r"""
+        Construct kernel matrix for given data and feature map
+
+        If y is None, self inner product is calculated.
```
woodsp-ibm, line 170: If `y_vec` is None, would be more precise, 

Having said that are there better names that can be given to the params other than x_vec and y_vec which really don't say a lot. But then looking at kernel examples to sklearn they use X and Y so maybe things are fine if this is an understood use.

qiskit_machine_learning/kernels/quantum_kernel.py
attp: To be discussed with @stefan-woerner who added the TODO.

qiskit_machine_learning/kernels/quantum_kernel.py
attp: @stefan-woerner What do you think? I'm not sure what a good default feature map could be.

qiskit_machine_learning/kernels/quantum_kernel.py
attp: Changed the comment to `y_vec`, but leaving the variable names as they are, since we want this to be easy for people familiar with sklearn to use.

qiskit_machine_learning/kernels/quantum_kernel.py
attp: Changed from `AlgorithmError` to `QiskitMachineLearningError`

attp:
@stefan-woerner previously you mentioned that the `QSVR` and `QSVC` classes should perhaps directly take a `QuantumKernel` in their constructor, rather than a `feature_map / quantum_instance`. we should decide on this. I don't have a strong opinion about this either way.

qiskit_machine_learning/kernels/quantum_kernel.py
stefan-woerner: We can use the `ZZFeatureMap` as default, that's the one from the original paper.

qiskit_machine_learning/kernels/quantum_kernel.py
stefan-woerner: This implicitly assumes that the number of qubits equals the number of parameters (as in the ZZFeatureMap). However, that doesn't always need to be the case. Here we should adjust as follows:
1. If `self._feature_map.num_parameters == x_vec.shape[1]` nothing will be done (cf. check in line 210).
2. if 1. does not hold, then we can check wether the feature map actually allows to set the `num_qubits` or not. If not, raise Exception, if yes, adjust and assume it sets the number of parameters correctly, if htis is not true, raise an exception.

attp: Turns out this assignment isn't required in the current Qiskit, (`self._feature_map.num_qubits = x_vec.shape[1]`) and subsequent checks for `self._feature_map.num_parameters` and `x_vec.shape[1]`, `y_vec.shape[1]` are sufficient.

qiskit_machine_learning/kernels/quantum_kernel.py
attp: Default `ZZFeatureMap` added with default 2 parameters.

stefan-woerner:
@attp let's leave as is with feature map and quantum instance.

qiskit_machine_learning/algorithms/__init__.py
```diff
@@ -44,14 +45,16 @@
...
+from .classifiers import QSVC, VQC
+from .regressors import QSVR
 from .distribution_learners import (DiscriminativeNetwork,
                                     GenerativeNetwork,
                                     NumPyDiscriminator,
                                     PyTorchDiscriminator,
                                     QuantumGenerator, QGAN)
 
 __all__ = [
+    'QSVC',
```
stefan-woerner, line 57: please add QSVR

attp:
> @attp let's leave as is with feature map and quantum instance.

Now that I'm adding documentation and tests, it makes more sense to initialise `QSVC` and `QSVR` with a `QuantumKernel` rather than a `feature_map` and `quantum_instance`.

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -0,0 +1,43 @@
...
+    r"""Quantum Support Vector Classifier.
+
+        **Example**
+        .. code-block::
+            qsvc = QSVC(feature_map=map, quantum_instance=backend)
+            qsvc.fit(sample_train,label_train)
+            qsvc.predict(sample_test)
+    """
+
+    def __init__(self,
```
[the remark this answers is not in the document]
attp: @woodsp-ibm I've updated the QSVC to allow no QuantumKernel in its constructor, and then call the superclass constructor with an default empty QuantumKernel. 

However, from the [documentation](https://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html), I can't see a way to update the kernel in a SVC object after it's been constructed. I've tried to using a `@quantum_kernel.setter` but it doesn't work, as shown in the `test_change_kernel` in test_qsvc. 

I'm assuming I'm missing something really basic in Python here. Any ideas?

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -0,0 +1,62 @@
...
+    @property
+    def quantum_kernel(self) -> QuantumKernel:
+        """ Returns quantum kernel """
+        return self._quantum_kernel
+
+    @quantum_kernel.setter
+    def quantum_kernel(self, quantum_kernel: QuantumKernel):
+        """ Sets quantum kernel """
+        self._quantum_kernel = quantum_kernel
+        self.kernel = self._quantum_kernel
```
stefan-woerner, line 62:
```suggestion
        self.kernel = self._quantum_kernel.evaluate
```

qiskit_machine_learning/algorithms/classifiers/qsvc.py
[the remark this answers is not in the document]
stefan-woerner: see my comment below (you need to pass `evaluate` to sklearn)

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -0,0 +1,63 @@
...
+
+    .. code-block::
+
+        qsvc = QSVC(quantum_kernel=qkernel)
+        qsvc.fit(sample_train,label_train)
+        qsvc.predict(sample_test)
+    """
+
+    def __init__(self, *args,
+                 quantum_kernel: Optional[QuantumKernel] = None, **kwargs):
```
woodsp-ibm, line 44: I assume *args and **kwargs are to allow you to pass in parameters as supported by the sklearn SVC? In which case it might be helpful to state this.

Can you also do unit tests to ensure that when passing in extra positional args and keyword kwargs that it indeed does what is expected.

qiskit_machine_learning/algorithms/regressors/qsvr.py
```diff
@@ -0,0 +1,63 @@
...
+
+    .. code-block::
+
+        qsvr = QSVR(quantum_kernel=qkernel)
+        qsvr.fit(sample_train,label_train)
+        qsvr.predict(sample_test)
+    """
+
+    def __init__(self, *args,
+                 quantum_kernel: Optional[QuantumKernel] = None, **kwargs):
```
woodsp-ibm, line 44: Similar comment to QSVC in regards of args/kwargs

qiskit_machine_learning/algorithms/regressors/qsvr.py
```diff
@@ -0,0 +1,63 @@
...
+
+from ...kernels.quantum_kernel import QuantumKernel
+
+logger = logging.getLogger(__name__)
+
+
+class QSVR(SVR):
+    r"""Quantum Support Vector Regressor.
+
+    This class shows how to use a quantum kernel for classification. The class extends
```
woodsp-ibm, line 28: Would we want to make this text a bit more different that the classifier text (at present its just the final link below to sklearn user guide whose difference would not be clear in the docs either without hovering over the text or clicking on it . Might it be better here to start out `This class shows how to use a quantum kernel for regression` i.e. change to classification to regression at least.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,327 @@
...
+    Usually m is much larger than n.
+
+    The quantum kernel algorithm calculates a kernel matrix, given datapoints x and y and feature
+    map f, all of n dimension. This kernel matrix can then be used in classical machine learning
+    algorithms such as support vector classification, spectral clustering or ridge regression.
+    """
+
+    def __init__(self,
+                 feature_map: Optional[QuantumCircuit] = None,
+                 enforce_psd: bool = False,
```
woodsp-ibm, line 52: In the prior QSVM we had enforce_psd default as True so it would work by default with the qasm simulator. I would have thought perhaps True would be a better default here too. It may be worth stating here on the docstring that this is only done on when sampling using a qasm sim or real device i.e. when not using a statevector sim.

qiskit_machine_learning/algorithms/classifiers/qsvc.py
woodsp-ibm: That works. You could also have an evaluate method in this class, that you give to SVC, where this just forwards the call to self._kernel.evaluate to do the real work. That way when self.kernel is updated it still goes to the right place because of the forwarding.

stefan-woerner (approved):
lgtm

woodsp-ibm (approved):

PR #7, opened 2021-03-11
Adjust size of extreme bins in discretize_and_truncate

gabrieleagl:
### Summary
`discretize_and_truncate` currently has an odd behaviour on the first and last bin, that result to be smaller than the others.

### Details and comments
`discretize_and_truncate` splits the domain `(bounds[j, 1], bounds[j, 0])` in `(2 ** prec)` bins of equal size.

In the previous implementation, the first bin was centered in `bounds[j, 0]` and the last in `bounds[j, 1]`. So, half of the first bin fell out of the bounds, and therefore the number of samples in the first bin was misleadingly low. Similarly for the last bin.

I propose to change the logic, so that all bins fall completely within the bounds; notice that this way, the "label" (central point) of the first bin is no longer `bounds[j, 0]`, but rather `bounds[j, 0] + something`; in the same way, the label of the last bin becomes `bounds[j, 1] - something`.

@Zoufalc


gabrieleagl:
@Zoufalc I'm realizing right now that if the original domain was discrete, say integer numbers `[j_1, .., j_N]`, with this modification we would "relabel" outputs as `[j_1+someting, ..., j_N-something]`. What do you think?

Zoufalc:
Thank you  @gabrieleagl for raising this point.
The relabeling is something that we should avoid from my point of view.

Rather than changing the grid, I would prefer if lines 180 & 182 in `dataset_helper.py` are being changed, i.e. that values are not cut off exactly at the min/max of the bounds but symmetrically around the min and max of the bounds according to the grid resolution.

gabrieleagl:
Shall we rename the parameter `bounds` into `extreme_bins`, `extreme_bin_labels`, or similar, in this case?

Zoufalc:
You can rename it to discretization_bounds

gabrieleagl:
I have opened a new pull request for this; refer to https://github.com/Qiskit/qiskit-machine-learning/pull/17

PR #10, opened 2021-03-13
Setup not including top folder files

manoelmarques:
### Summary

Wheel file was not including the top files.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
-recursive-include qiskit_machine_learning
+recursive-include qiskit_machine_learning *.txt
 graft test
 prune docs
 prune tools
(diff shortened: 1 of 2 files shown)


mtreinish:
The issue here and in the similar prs is the use of find_namespace_packages, that should only be used in a namespace package (like aqua) where the root of the namespaceis in another package. You should be using find_packages in packages that are self contained like this.

manoelmarques:
I thought according that: "...Full support for PEP 420 via find_namespace_packages(), which is also backwards compatible to the existing find_packages() for Python >= 3.3." at https://setuptools.readthedocs.io/en/latest/setuptools.html

So I thought find_namespace_packages was the way to go forwards but I will go back to find_packages.

woodsp-ibm (approved):

PR #12, opened 2021-03-16
[WIP] Adding Torch connector

stefan-woerner:
### Summary
This PR adds the `TorchConnector` to use a (Quantum) `NeuralNetwork` form Qiskit Machine Learning as a PyTorch `Module`. Further, it adds corresponding tutorials.


qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -0,0 +1,101 @@
...
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""A connector to use Qiskit (Quantum) Neural Networks as PyTorch modules."""
+
+import numpy as np
+
+from torch import Tensor
```
woodsp-ibm, line 17: I would assume, that since torch is not included conditionally, and must exist, that we will not by default be importing this class elsewhere? We just need to ensure that things work as expected when torch is not installed and things do not fail completely to load up etc.

Is there an expectation to convert the qgan pytorch discriminator to somehow using this connector? But my reading is that this is more the other way around in that whatever networks are provided here can be used more directly inside classical Torch code.

qiskit_machine_learning/connectors/__init__.py
```diff
@@ -0,0 +1,35 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2019, 2021.
```
woodsp-ibm, line 3: I guess we should try with these dates to have the first date correct. I think just 2021 suffices for this file as its completely new right.

qiskit_machine_learning/connectors/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+   :nosignatures:
+
+   TorchConnector
+
+"""
+
+from .torch_connector import TorchConnector
+
+
+__all__ = ["TorchConnector"]
```
woodsp-ibm, line 35: See my comment below in terms of conditional import. If there is more than one connector from different frameworks here then we need to be careful that using the one you want does not fail because some other connector is loaded and its dependent pkg is not installed.

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -43,19 +43,21 @@ def __init__(self, circuit: QuantumCircuit,
...
+                * 'int' (default): an integer corresponding to the bitstring, e.g. '01010' -> 10
                 * 'str': a bitstring of type str, e.g. '01010'
-                * 'int': an integer corresponding to the bitstring, e.g. 10
+                * 'tuple': a tuple of binary values, e.g. (0, 1, 0, 1, 0)
                 * a custom callable that takes a sample of type 'tuple' and maps it to some other
                 output, output should be hashable for sparse representation of probabilities
                 and probability gradients.
             dense: Whether to return a dense (array with 'output_shape') or sparse (dict)
                 probabilities. Dense probabilities require "interpret == 'int'" where the integer
                 will be the index in the array of probabilities.
```
woodsp-ibm, line 54: Code below suggests that 'callable' is also valid, given certain output from it, for dense. That would be nice to state here too

qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -0,0 +1,101 @@
...
+                    if len(weights_grad.shape) == 1:
+                        weights_grad = weights_grad.reshape(1, len(weights_grad))
+                    weights_grad = grad_output.float() @ Tensor(weights_grad)
+
+            # return gradients for the first two arguments and None for the others
+            return input_grad, weights_grad, None
+
+    def __init__(self, nn: NeuralNetwork):
+        """Initializes the TorchConnector."""
+        super(TorchConnector, self).__init__()
```
woodsp-ibm, line 88: This seems like it should not be needed and seems like old style of calling super too

tutorials/01_neural_networks.ipynb
```diff
@@ -0,0 +1,957 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "engaging-webster",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "\n",
+    "from qiskit import Aer, QuantumCircuit\n",
```
woodsp-ibm, line 30: I guess this is a case where we really want to use Aer (for its snapshot capability i.e. using AerPauliExpectatio) and using BasicAer would not suffice.

tutorials/01_neural_networks.ipynb
```diff
@@ -0,0 +1,957 @@
...
+      "«     ┌──────────────────────────────┐┌─┴─┐\n",
+      "«q_2: ┤ P(2.0*(π - x[1])*(π - x[2])) ├┤ X ├\n",
+      "«     └──────────────────────────────┘└───┘\n"
+     ]
+    }
+   ],
+   "source": [
+    "# specify the feature map\n",
+    "fm = ZZFeatureMap(num_qubits, reps=2)\n",
+    "print(fm.draw())"
```
woodsp-ibm, line 360: Might want to look at the 'mpl' option since it gives nicer output in these notebookss

qiskit_machine_learning/connectors/__init__.py
stefan-woerner: done.

tutorials/01_neural_networks.ipynb
stefan-woerner: done.

qiskit_machine_learning/connectors/torch_connector.py
stefan-woerner: added logic to check whether torch has been installed.

adekusar-drl (approved):
Thanks for taking care the review! Looks good!

manoelmarques (approved):

PR #13, opened 2021-03-16
[WIP] initial version of classifier and regressor

beichensinn:


woodsp-ibm:
Please make sure all files here have the required copyright header in them - this includes the init files too.

qiskit_machine_learning/utils/loss_functions/loss.py
```diff
@@ -0,0 +1,111 @@
+from abc import ABC, abstractmethod
+import numpy as np
+
+from ...exceptions import QiskitMachineLearningError
+
+
+class Loss(ABC):
```
woodsp-ibm, line 7: Do we need a whole package for one file. If these were all in separate files I could see it. But with everything in one file I would have a thought a loss_functions.py file directly in utils would have sufficed. Unless there is some view to future expansion that would see more coming to this loss_funtions folder.

beichensinn: Yes. it is our plan to add more loss in this folder.

test/test_readme_sample.py
```diff
@@ -47,24 +49,42 @@ def print(*args):
 
         # Use Wine data set for training and test data
         feature_dim = 4  # dimension of each data point
-        _, training_input, test_input, _ = wine(training_size=12,
-                                                test_size=4,
+        # sample_train, training_input, test_input, class_labels
```
woodsp-ibm, line 52: This code, ie the body of the test from here ` # --- Exact copy of sample code ----------------------------------------` to the similar marker, is supposed to be a direct copy/paste of the sample code that is shown in the readme. The rationale being to make sure that the sample code we show continues to work.

We can change the readme to whatever sort of sample we thing might best showcase things and be a simple sample a user can run. Whatever it is though should just be copy/paste here in this test and the two should not diverge.

If you feel you need some other better/different test cases for VQC thats fine. This though was specific to making sure we test the readme sample that it continues to work.

adekusar-drl: That's right, once this test works, I will update readme.

stefan-woerner (approved):
lgtm

PR #14, opened 2021-03-18
Enable SparseArray for Neural Networks

stefan-woerner:
### Summary
Enables `SparseArray` as possible return type for `NeuralNetwork`. This use used and needed particularly for `CircuitQNN` where the output can be exponential but very sparse arrays that wouldn't scale otherwise.

### Details and comments

TODOs
- [x] add `sparse` to setup.py
- [x] unit tests


adekusar-drl (approved):
Looks good, thanks!

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -42,26 +48,23 @@ def __init__(self, circuit: QuantumCircuit,
...
             gradient: The gradient converter to be used for the probability gradients.
             quantum_instance: The quantum instance to evaluate the circuits.
 
         Raises:
-            QiskitMachineLearningError: if an incorrect value for `interpret` or `output_shape`
-                is passed.
+            QiskitMachineLearningError: if `interpret` is passed without `output_shape`.
         """
 
         # TODO: currently cannot handle statevector simulator, at least throw exception
```
woodsp-ibm, line 70: I see in the _sample method below a test for statevector - does this mean this TODO is outdated and can be removed. Should the test be done earlier, so it can be caught earlier rather than later, by that I mean check when the QI is provided or is there other function where this is still useful.

I did not see any setter for QI like we often have to allow it to be changed.

adekusar-drl (approved):

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -176,10 +154,17 @@ def _sample(self, input_data: np.ndarray, weights: np.ndarray) -> np.ndarray:
         self.quantum_instance.backend_options['memory'] = orig_memory
 
         # return samples
-        return np.array([self._interpret_bitstring(b) for b in result.get_memory()])
+        memory = result.get_memory()
+        samples = np.zeros((1, *self.output_shape))
+        for i, b in enumerate(memory):
+            sample: Union[int, Tuple[int, ...]] = int(b, 2)
+            if self._interpret:
+                sample = self._interpret(cast(int, sample))
```
woodsp-ibm, line 162: I see this pattern in here a bit. It might have been easier to have set the method on the class called _interpret, when given an interpret of None (default), to some "dummy callback" then simply returned the same value as it was given that way the self._interpret could have always been called and any one doing any future changes could not forget to do the if test before calling it as that would not longer be needed,

qiskit_machine_learning/neural_networks/neural_network.py
```diff
@@ -27,12 +28,13 @@ class NeuralNetwork(ABC):
     batched inputs. This is to be implemented by other (quantum) neural networks.
     """
 
-    def __init__(self, num_inputs: int, num_weights: int,
+    def __init__(self, num_inputs: int, num_weights: int, dense: bool,
```
woodsp-ibm, line 31: Should we define a default for dense so that for most users it works without needing to set this, though that would mean changing the order and moving it to end if output_shape has no meaningful default and must always be given.

qiskit_machine_learning/neural_networks/sampling_neural_network.py
```diff
@@ -26,23 +27,24 @@ class SamplingNeuralNetwork(NeuralNetwork):
...
+                 output_shape: Union[int, Tuple[int, ...]]) -> None:
         """
 
         Args:
             num_inputs: The number of input features.
             num_weights: The number of trainable weights.
-            output_shape: The shape of the output.
-            return_samples: Determines whether the network returns a batch of samples or a sparse
-                vector (dictionary) of probabilities in its forward pass. In case of probabilities,
+            dense: Returns whether the output is dense or not.
```
woodsp-ibm, line 37: Similar comment to dense above for the base class; and maybe likewise for return_samples.

tutorials/conf.py
```diff
@@ -70,7 +70,7 @@
 html_sourcelink_suffix = ''
 exclude_patterns = ['_build', '**.ipynb_checkpoints']
 
-cell_timeout = int(os.getenv('QISKIT_CELL_TIMEOUT', 180))
+cell_timeout = int(os.getenv('QISKIT_CELL_TIMEOUT', 360))
```
woodsp-ibm, line 73: Since I assume these will be built by the main qiskit tutorials process then I guess this change would have to be reconciled  with what is expected there too.

pbark (approved):

qiskit_machine_learning/neural_networks/neural_network.py
stefan-woerner: We have currently 2 implementations, for one it is always dense, for the second both makes sense. I'd like to keep like that for now.

qiskit_machine_learning/neural_networks/circuit_qnn.py
stefan-woerner: removed TODO and added QI setter.

qiskit_machine_learning/neural_networks/circuit_qnn.py
stefan-woerner: adjusted accordingly

qiskit_machine_learning/neural_networks/sampling_neural_network.py
stefan-woerner: We have only one example so far, let's not fix a default here but in the implementation.

woodsp-ibm (approved):

PR #17, opened 2021-03-22
Adjust size of extreme bins in discretize_and_truncate (2)

gabrieleagl:
### Summary
`discretize_and_truncate` currently has an odd behaviour on the first and last bin, that result to be smaller than the others.

### Details and comments
`discretize_and_truncate` splits data into `(2 ** num_qubits[j])` bins of equal size and equally spaced, where the first bin is centered in `(bounds[j, 0]` and the last in `bounds[j, 1])`.

In the previous implementation, data were cut so that only values between `bounds[j, 0]` and `bounds[j, 1]` survived. So, half of the first bin fell out of the bounds, and therefore the number of samples in the first bin was misleadingly low. Similarly for the last bin.

Code was modified as discussed in https://github.com/Qiskit/qiskit-machine-learning/pull/7


woodsp-ibm:
@Zoufalc can I ask you to take a look over this please

Zoufalc (changes requested):
Thanks for the changes. The code looks good. 
Please add a note in the docstring which explains that the bounds are now not really the bounds anymore but actually represent the bin centers. 
I would also suggest using the name `discretization_bin_centers` instead of `discret_bounds`. Otherwise, the users might be confused when data points that are lower than the bound are being included in their experiment.

gabrieleagl:
Modified the paramater name to `min_max_bin_centers` as agreed offline

Zoufalc (approved):
Thank you @gabrieleagl!

woodsp-ibm (dismissed):

woodsp-ibm:
@Zoufalc This method is used from QGAN where it seems bounds is parameter  of its constructor and this bounds ends up getting passed to this newly named parameter (which was bounds before this adjustment). Does QGAN need to change in anyway given this now seems to have a different interpretation of what it means?

Also, is this method principally for QGAN as in arguably, given this code had now been released, existing names/behavior ought to be deprecated and phased out rather than simply being changed. I guess if this is only used by QGAN and the impact is low we might be able to avoid that; on QGAN itself, if that needs changing, then thats less obvious. Maybe even there the bounds param, given it was optional, is not often used? Ideally we deprecate though.

adekusar-drl:
@woodsp-ibm, @Zoufalc what's the status of this PR?

Zoufalc:
@woodsp-ibm Sorry for the delayed reply.
The new name is reflective of what is actually happening so the qGAN code doesn't need to be changed.
I see your point. I would also be fine with keeping the parameter name 'bounds' as is. @gabrieleagl what is your opinion here? Do you think that the renaming of the parameter is vital or would it be sufficient to include the shifting changes?

gabrieleagl:
Hello  @Zoufalc, @woodsp-ibm, I personally prefer the new name, as it is more representative.

As far as deprecation is concerned, I believe it's not really necessary, because the method is only used for the qGAN, and arguments are typically called positionally.
Anyway, if you really think it's better to have the deprecation for consistency, I can do that, which means:
- making `min_max_bin_centers` and `num_qubits` optional
- adding an additional optional parameter `bounds` at the end (so that the order is respected, in case of positional calls)
- checking that `num_qubits` is always values, and raise an exception otherwise
- checking `min_max_bin_centers` or `bounds` is valued
- adding a deprecation message if `bounds` is valued.
If we take this path, I would need your help on which exceptions to raise, and how to create the deprecation message.

Let me know!

Zoufalc:
Either way is fine with me. For me, the most important point was to fix the behavior of the binning.

qiskit_machine_learning/datasets/dataset_helper.py
```diff
@@ -19,16 +19,27 @@
...
+
+    For each dimension k, the domain is split into (2 ** num_qubits[k]) bins equally spaced and
+    equally sized, each centered in min_max_bin_centers[k, 0], ..., min_max_bin_centers[k, 1]. Bins
+    have size equal to
+    (min_max_bin_centers[k, 1] - min_max_bin_centers[k, 0])/(2 ** num_qubits[k] - 1).
+    Notice that:
+    * Every sample in data that falls out of the bins is discarded.
+    * The leftmost bin extends both to the left and to the right around its center,
+    therefore min_max_bin_centers[k, 0] is not the left bound for truncation, but only the center of
+    the leftmost bin. Similar considerations hold for min_max_bin_centers[k, 1] on the right.
```
adekusar-drl, line 36:
```suggestion
    For each dimension k, the domain is split into ``(2 ** num_qubits[k])`` bins equally spaced and
    equally sized, each centered in ``min_max_bin_centers[k, 0], ..., min_max_bin_centers[k, 1]``. Bins
    have size equal to
    ``(min_max_bin_centers[k, 1] - min_max_bin_centers[k, 0]) / (2 ** num_qubits[k] - 1)``.
    Notice that:
    
    * Every sample in data that falls out of the bins is discarded.
    * The leftmost bin extends both to the left and to the right around its center,
    
    therefore ``min_max_bin_centers[k, 0]`` is not the left bound for truncation, but only the center of
    the leftmost bin. Similar considerations hold for ``min_max_bin_centers[k, 1]`` on the right.
```

qiskit_machine_learning/datasets/dataset_helper.py
```diff
@@ -19,16 +19,27 @@
...
 
     Args:
         data (list or array or np.array): training data (int or float) of dimension k
-        bounds (list or array or np.array):  k min/max data values
-            [[min_0,max_0],...,[min_k-1,max_k-1]] if univariate data: [min_0,max_0]
+        min_max_bin_centers (list or array or np.array):  k min/max data values
+            [[min_center_0, max_center_0],...,[min_center_k-1, max_center_k-1]].
+            If univariate data: [min_center_0, max_center_0]
```
adekusar-drl, line 42:
```suggestion
        min_max_bin_centers (list or array or np.ndarray):  ``k`` min/max data values
            ``[[min_center_0, max_center_0],...,[min_center_k-1, max_center_k-1]]``.
            If univariate data: ``[min_center_0, max_center_0]``
```

qiskit_machine_learning/datasets/dataset_helper.py
```diff
@@ -19,16 +19,27 @@
...
         num_qubits (list or array or np.array): k numbers of qubits to determine
             representation resolution, i.e. n qubits enable the representation of 2**n
             values [num_qubits_0,..., num_qubits_k-1]
```
adekusar-drl, line 45: Could you please reformat the whole docstring to highlight code/math as I suggested?

adekusar-drl (commented):
Looks good, a few suggestion on the docstring formatting.

woodsp-ibm:
As far as the deprecation is concerned; which since this is a public API we should in principle do, the simplest way is to keep the old method as-is and just add another new method as the 'replacement' with this new interface.  In the old method you would add a deprecation warning indicating the new method which should be used instead; of course wherever we call the old one we would want to change over to the new one - just in QGAN I guess. Here is an example of a deprecation warning in code https://github.com/Qiskit/qiskit-terra/blob/68b5ede97b6bf5454250927fb522be217f3f3b7e/qiskit/circuit/library/probability_distributions/uniform.py#L57-L60

Zoufalc (approved):

adekusar-drl (approved):

Issue #36, opened 2021-04-03
test_size parameter for the iris dataaset has been hardcoded to 1


charmerDark:
### Information

- **Qiskit Machine Learning version**:
0.1
- **Python version**:
3.8.6
- **Operating system**:
Windows
### What is the current behavior?
upon calling the iris() function from datasets module, the size of the testinput is always a single data point. This is because the test_size parameter for the train_test_split() function in used has been hardcoded to 1 instead of the parameter test_size of the iris() function

### Steps to reproduce the problem

```
from qiskit.ml.datasets import iris

datapoints, training_input, test_input, class_labels=iris(120,30,4)
print(test_input)

```
clearly gives a single element in the test_input. This can also be replicated by using any other value for test_size

### What is the expected behavior?

We get a test_input of the given size

### Suggested solutions

I believe this is an easy fix changing the test_size attribute in the train_test_split() function within datasets.iris

PR #37, opened 2021-04-03
update iris.py to fix test_size error

charmerDark:
Changed the value of the test_size attribute on the train_test_split() function fro the hardcoded 1 that causes bugs to the given parameter.
#36 should be fixed by this

### Summary

fix to #36 that should patch up the issue by now allowing train_test_split() to  use the test_size argument passed to the iris class


adekusar-drl:
@charmerDark You should update the test as well: `test/datasets/test_iris.py`.

adekusar-drl:
@woodsp-ibm @stefan-woerner Please take a look.

woodsp-ibm (approved):
It seems fine to me. I assume the json ref data changed since the test size, as per the test, is passed in as 3 and is actually used now and not forced to 1.

PR #40, opened 2021-04-09
Code refactoring

dlasecki:
### Summary
Had some time so decided to address issues suggested by Sonar Lint. Mostly reducing the Cognitive Complexity of the code. Let me know what you think.


qiskit_machine_learning/datasets/gaussian.py
```diff
@@ -22,138 +22,102 @@
...
+        _plot(n, label_train, samples_train, training_size)
+
+    return training_feature_array, training_label_array, test_feature_array, test_label_array
+
+
+def _generate_class_labels(n):
+    if n > 25:
+        # TODO change to it a more complex labeling system e.g. AA, AB, AAA etc. or just to C1,
+        #  C2, C3 etc.
+        raise Exception("To many classes requested. Maximum is 26")
```
woodsp-ibm, line 66: Should raise something like ValueError not a raw Exception. Also its informative to have the failing value included in the message.

qiskit_machine_learning/algorithms/distribution_learners/qgan/pytorch_discriminator.py
```diff
@@ -119,10 +119,9 @@ def get_label(self, x, detach=False):  # pylint: disable=arguments-differ
             torch.Tensor: Discriminator output, i.e. data label
         """
 
+        # TODO handle tensor case
```
woodsp-ibm, line 122: You have added TODOs where there were none before, I guess presuming pass was somehow a placeholder for doing something?

qiskit_machine_learning/datasets/gaussian.py
```diff
@@ -22,138 +22,102 @@
...
+
+    for t_r in range(training_size + test_size):
+        for feat in range(n):
+            for sample, randomized_vector in zip(samples_train, randomized_vectors):
+                sample[t_r][feat] = _calc_random_normal(randomized_vector, feat,
+                                                        sigma) if n == 2 \
+                    else _calc_random_normal_2(randomized_vector, feat, sigma)
+                # TODO unify both functions in the future when other n supported
+
+    _update_label_train(n, label_train, test_size, training_size)
```
woodsp-ibm, line 41: While I see the intent I just wonder if it goes too far. When we have method that is only called from one place and its a few lines long, for me this makes things harder to see as I have to go back and forth to methods to see what was laid out in front of me clearly before. But I do agree that there was room for improvement

qiskit_machine_learning/algorithms/distribution_learners/qgan/pytorch_discriminator.py
dlasecki: I suppose yes but I thought it would be best to eliminate the pass until this case is implemented and handled. Added a TODO so that we do not forget.

qiskit_machine_learning/datasets/gaussian.py
dlasecki: I agree, will fix that.

qiskit_machine_learning/algorithms/distribution_learners/qgan/pytorch_discriminator.py
woodsp-ibm: I think I am saying that it could well have been handled, that would have been my reading - i.e. it was explicit in there as not needing to do anything for that case - i.e. pass.

qiskit_machine_learning/datasets/gaussian.py
woodsp-ibm: FYI There was another exception in there that could have the value in the message too - I just did not call it out explicitly,

qiskit_machine_learning/datasets/gaussian.py
dlasecki: I often refer to the book by Robert C. Martin "Clean code". The essence of what he suggests there is summarized in this article (https://medium.com/swlh/clean-code-writing-functions-or-methods-4e6e53ff4ac2):

"The first rule of writing functions is functions should be small. The second rule for writing functions is that they should be even smaller. Functions that have 200 or 300 lines are kind of messy functions. They should not be longer than 20 lines and mostly less than 10 lines. Functions should be making as few arguments as possible, ideally none. Small functions are easy to understand and their intention is clear." 

But of course the line between what should be a separate function and what should not will always be blurry.

adekusar-drl:
@dlasecki @woodsp-ibm should we have an issue for this PR?

qiskit_machine_learning/datasets/gaussian.py
```diff
@@ -22,138 +22,102 @@
...
+
+
+def _calc_random_normal(randomized_vector, feat, sigma):
+    center = 1 / 2
+    return algorithm_globals.random.normal(pow(-1, randomized_vector[feat]) * center, sigma, None)
+
+
+def _calc_random_normal_2(randomized_vector, feat, sigma):
+    coefficient = 2 * (1 + 2 * randomized_vector[feat])
+    return algorithm_globals.random.normal(coefficient * np.pi / 6, sigma, None)
```
dlasecki, line 123: Would there be a way to generalize and unify these methods? What is the rule behind specifying the center of a distribution based on the number of qubits considered? @adekusar-drl

woodsp-ibm:
>  should we have an issue for this PR?

We might suggest, if you had an idea that you wanted to do, before creating a PR open an Issue so it can be described/outlined discussed ahead of doing a PR. Since there is a PR for the work an issue is not really required.

qiskit_machine_learning/datasets/gaussian.py
adekusar-drl: @dlasecki the means are different here as I see, so different distributions. This code was here way before me, so I don't know what the initial intention was behind it.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -223,106 +238,108 @@ def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
                     "x_vec and class feature map have incompatible dimensions.\n" +
                     "x_vec has %s dimensions, but feature map has %s." %
                     (x_vec.shape[1], self._feature_map.num_parameters)) from AttributeError
-
         if y_vec is not None and y_vec.shape[1] != self._feature_map.num_parameters:
             raise ValueError("y_vec and class feature map have incompatible dimensions.\n" +
```
dlasecki, line 242: Is there a reason why the blocks for `x_vec` and `y_vec` are slightly different?

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -223,106 +238,108 @@ def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
...
                              "y_vec has %s dimensions, but feature map has %s." %
                              (y_vec.shape[1], self._feature_map.num_parameters))
```
dlasecki, line 244: Is there a reason why the blocks for `x_vec` and `y_vec` are slightly different?

adekusar-drl: @dlasecki I don't really get the question. 
If you mean this block: https://github.com/Qiskit/qiskit-machine-learning/blob/master/qiskit_machine_learning/kernels/quantum_kernel.py#L218:L230 then it looks good to me.

dlasecki: Sorry, my line selection did not render well. I was hoping to remove some code duplication there and have been wondering about the missing `try` for `y_vec`. But now I noticed the previous `if  `statement that explains it.

qiskit_machine_learning/datasets/gaussian.py
dlasecki: I see. We would ideally like to support an arbitrary number of qubits (not just 2 and 3), so I have been trying to figure out which direction would be suggested regarding distributions. Currently, the upper one is used in case of 2 qubits and the bottom one in case of 3 qubits...

adekusar-drl:
Closing this PR as an outdated. Many of the refactorings in this PR were implemented in other PRs. Thanks!

PR #55, opened 2021-04-16
Check if torch is installed on tests

### LINKED ISSUE #54: Tests (Torch): test_batch_gradients fails if pytorch not installed
Most ``test_torch_connector.py`` tests are wrapped in logic to skip the test if ``pytorch`` isn't installed, but not ``test_batch_gradients()``. Thus, tests would be properly skipped for all torch tests, except for ``test_batch_gradients``.

https://github.com/Qiskit/qiskit-machine-learning/blob/7824bacf7c3288aa5ab8ca1f9d1ae7177a8e81ee/test/connectors/test_torch_connector.py#L466-L529

For logic that should be there, ee e.g. https://github.com/Qiskit/qiskit-machine-learning/blob/7824bacf7c3288aa5ab8ca1f9d1ae7177a8e81ee/test/connectors/test_torch_connector.py#L450-L464

manoelmarques:
### Summary

Fixes #54

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -143,6 +143,14 @@ jobs:
         with:
           name: ml${{ matrix.python-version }}
           path: ./m${{ matrix.python-version }}/*
+      - name: Machine Learning Unit Tests without torch/torchvision under Python ${{ matrix.python-version }}
+        run: |
+          pip uninstall -y torch torchvision
+          if [ "${{ github.event_name }}" == "schedule" ]; then
+              export QISKIT_TESTS="run_slow"
+          fi
+          stestr --test-path test run
+        shell: bash
   Tutorials:
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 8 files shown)


drewrisinger (approved):
* [x] Diff LGTM
* [x] CI Tests pass

Looks like your solution was to wrap the test in a try/catch block. I was wondering if a fixture something like https://github.com/Qiskit/qiskit-terra/blob/0fc83f44a6e80969875c738b2cee7bc33223e45f/qiskit/test/decorators.py#L66-L76
would be a better way to do it, but that could be a later step.

woodsp-ibm:
I noticed, via the CI change to uninstall what we install as extras, that we `torchvision` as an extra requires. From what I can tell torchvision is only used in the notebook tutorial, not in the core library code. But if the user installs torch via the `pip install qiskit-machine-learning[torch]` they will get torchvision that the library does not use/need. I think the extra requires should be limited to what the library needs/can use i.e. torch and should not install torchvision that maybe a user does not want and certainly does not need for the core library. We can install torchvision in the CI for running notebooks (or any other library that the notebook examples may use in the future that the core does include/need).

manoelmarques:
I removed torchvision from extra requires.

woodsp-ibm (approved):

PR #64, opened 2021-04-20
Extend docstrings of loss functions

### LINKED ISSUE #49: Extend loss function doc strings
In the current release loss functions docstring documentation is very scarse. It should be extended and enhanced.

beichensinn:
### Summary
Added docstrings.
Closes #49.


qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -20,24 +20,57 @@
...
+
+        Returns:
+            a float value of the loss function
+        """
         return self.evaluate(predict, target)
 
     @abstractmethod
-    def evaluate(self, predict, target):
-        """ evaluate """
+    def evaluate(self, predict: np.ndarray, target: np.ndarray):
```
woodsp-ibm, line 38: I see in the derived classes they document this returning a float and the meaning of the float. In which case I think the signature here should be updated to return a float ie `def evaluate(self, predict: np.ndarray, target: np.ndarray) -> float:` and document here the meaning of that return.

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -20,24 +20,57 @@
...
+            target: a numpy array of the true values
+
+         An abstract method for evaluating the loss function
+         """
         raise NotImplementedError
 
     @abstractmethod
-    def gradient(self, predict, target):
-        """ gradient """
+    def gradient(self, predict: np.ndarray, target: np.ndarray):
```
woodsp-ibm, line 49: Similar comment on return value.

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -20,24 +20,57 @@
...
+            predict: a numpy array of predicted values using the model
+            target: a numpy array of the true values
+
+         An abstract method for computing the gradient
+         """
         raise NotImplementedError
 
     @staticmethod
-    def _validate(predict, target):
+    def _validate(predict: np.ndarray, target: np.ndarray):
```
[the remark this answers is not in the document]
beichensinn: updated

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -47,9 +80,24 @@ def _validate(predict, target):
 
 
 class L1Loss(Loss):
-    """ L1Loss """
+    """
+    L1Loss:
+        This class computes the L1 loss: sum |target - predict|
+    """
+
+    def evaluate(self, predict: np.ndarray, target: np.ndarray):
```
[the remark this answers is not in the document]
beichensinn: updated

qiskit_machine_learning/utils/loss_functions/loss_functions.py
beichensinn: updated

qiskit_machine_learning/utils/loss_functions/loss_functions.py
beichensinn: updated

qiskit_machine_learning/utils/loss_functions/loss_functions.py
woodsp-ibm: For both `evaluate` and `gradient` we expect them to return a float right so the caller of a concrete sub-class knows what to expect. Hence in the abstract methods here the signature should show the float return typehint and the Returns in the docstring document the meaning of the float. The docstrings here would then be complete meaning no docstring should be needed on the derived classes - which at present all have a copy/paste of the same Returns text when they should all just simply have no docstring i.e inherit the one from the base class.

adekusar-drl:
@woodsp-ibm if you could take a look again at the PR.

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -20,37 +20,80 @@
...
+
+        Returns:
+            a float value of the loss function
+        """
         return self.evaluate(predict, target)
 
     @abstractmethod
-    def evaluate(self, predict, target):
-        """evaluate"""
+    def evaluate(self, predict: np.ndarray, target: np.ndarray):
```
woodsp-ibm, line 39:
```suggestion
    def evaluate(self, predict: np.ndarray, target: np.ndarray) -> float:
```
The typehint should show it returns a float

Also there should be a Returns: entry in the docstring that states the meaning of the return valie

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -20,37 +20,80 @@
...
+
+        Raises:
+            QiskitMachineLearningError: shapes of predict and target do not match
+        """
         raise NotImplementedError
 
     @abstractmethod
-    def gradient(self, predict, target):
-        """gradient"""
+    def gradient(self, predict: np.ndarray, target: np.ndarray):
```
woodsp-ibm, line 53:
```suggestion
    def gradient(self, predict: np.ndarray, target: np.ndarray) -> float:
```
The typehint should show the return type

And likewise there should be a Returns: entry in the docstring that states the meaning of the return value

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -20,37 +20,80 @@
...
+            target: a numpy array of the true values
+
+        Raises:
+            QiskitMachineLearningError: shapes of predict and target do not match.
+        """
         raise NotImplementedError
 
     @staticmethod
-    def _validate(predict, target):
+    def _validate(predict: np.ndarray, target: np.ndarray):
```
woodsp-ibm, line 67:
```suggestion
    def _validate(predict: np.ndarray, target: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
```
The typehint should have the return type. The docstring has a Returns entry so thats good.

I must admit this method is kinda confusing. If it just needs to validate the shapes why does it need to return them since the caller already has them? Now I see it makes a numpy array of them, but then the typehint says it takes a numpy array and not something like a list where I could see it adjusting the types as needed as well as doing some validation, with the (possibly) adjusted values returned,

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -20,37 +20,80 @@
...
+                f"Shapes don't match, predict: {predict.shape}, target: {target.shape}!"
             )
         return predict, target
 
 
 class L1Loss(Loss):
-    """L1Loss"""
+    """
+    L1Loss:
+        This class computes the L1 loss: sum |target - predict|
```
woodsp-ibm, line 93: Duplicating the name of the class into the docstring is not necessary; It should be a simple sentence summary, as you have there already, followed by further paragraph(s) of text describing things for more complicated classes that need it = which I do not think is the case here.
```suggestion
    This class computes the L1 loss: sum |target - predict|
```

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -62,16 +105,20 @@ def evaluate(self, predict, target):
...
         predict, target = self._validate(predict, target)
 
         return np.sign(predict - target)
 
 
 class L2Loss(Loss):
-    """L2Loss"""
+    """
+    L2Loss:
+        This class computes the L2 loss: sum (target - predict)^2
```
woodsp-ibm, line 117:
```suggestion
    This class computes the L2 loss: sum (target - predict)^2
```

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -81,31 +128,38 @@ def evaluate(self, predict, target):
...
         predict, target = self._validate(predict, target)
 
         return 2 * (predict - target)
 
 
 class CrossEntropyLoss(Loss):
-    """CrossEntropyLoss"""
+    """
+    CrossEntropyLoss:
+        This class computes the cross entropy loss: -sum target * log(predict)
```
woodsp-ibm, line 140:
```suggestion
    This class computes the cross entropy loss: -sum target * log(predict)
```

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -81,31 +128,38 @@ def evaluate(self, predict, target):
...
         predict, target = self._validate(predict, target)
 
         return predict * np.sum(target) - target
 
 
 class CrossEntropySigmoidLoss(Loss):
-    """This is used for binary classification"""
+    """
+    CrossEntropySigmoidLoss:
+        This class computes the cross entropy sigmoid loss.
```
woodsp-ibm, line 157:
```suggestion
    This class computes the cross entropy sigmoid loss.
```

beichensinn: updated

qiskit_machine_learning/utils/loss_functions/loss_functions.py
beichensinn: updated

qiskit_machine_learning/utils/loss_functions/loss_functions.py
beichensinn: updated

qiskit_machine_learning/utils/loss_functions/loss_functions.py
beichensinn: updated

qiskit_machine_learning/utils/loss_functions/loss_functions.py
beichensinn: updated

qiskit_machine_learning/utils/loss_functions/loss_functions.py
beichensinn: updated

stefan-woerner (dismissed):
lgtm

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -13,46 +13,95 @@
...
 
 class Loss(ABC):
     """
-    Abstract base class for Loss.
+    Abstract base class for computing Loss.
+
     """
 
-    def __call__(self, predict, target):
+    def __call__(self, predict: Union[int, np.ndarray], target: Union[int, np.ndarray]) -> float:
```
woodsp-ibm, line 29: So this is now either an integer or an numpy array.  The args below mention just the numpy array not a single value. I see the single value is reformed to a numpy array. Does that mean the numpy arrays are arrays of ints too? I.e the predicted and target values are all ints. Maybe the Args docstring can be improved.

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -13,46 +13,95 @@
...
-    def evaluate(self, predict, target):
-        """evaluate"""
+    def evaluate(self, predict: Union[int, np.ndarray], target: Union[int, np.ndarray]) -> float:
+        """
+        An abstract method for evaluating the loss function
+
+        Args:
+            predict: a numpy array of predicted values using the model
+            target: a numpy array of the true values
+
```
woodsp-ibm, line 48: It would be good to have a Returns here to state what the float return is

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -13,46 +13,95 @@
...
-    def gradient(self, predict, target):
-        """gradient"""
+    def gradient(self, predict: Union[int, np.ndarray], target: Union[int, np.ndarray]) -> float:
+        """
+        An abstract method for computing the gradient
+
+        Args:
+            predict: a numpy array of predicted values using the model
+            target: a numpy array of the true values
+
```
woodsp-ibm, line 62: Again it would be good to have a Returns statement here for the float return

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -13,46 +13,95 @@
...
+        Raises:
+            QiskitMachineLearningError: shapes of predict and target do not match.
+        """
         raise NotImplementedError
 
     @staticmethod
-    def _validate(predict, target):
+    def _validate(
+        predict: Union[int, np.ndarray], target: Union[int, np.ndarray]
+    ) -> Tuple[np.ndarray, np.ndarray]:
```
woodsp-ibm, line 71: So if I pass in an 'int' and an 'int' as predict and target, I will not get back an Tuple of ints rather a Tuple of arrays? I guess its doing a type cast as well as validate that the other methods rely on. Not sure if it would be helpful to expose the duality in the name say `_cast_validate` or `_to_array_validate` - its a private method so maybe its less important.

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -13,46 +13,95 @@
...
-                f"Shapes don't match, predict: {predict.shape}, " f"target: {target.shape}!"
+                f"Shapes don't match, predict: {predict.shape}, target: {target.shape}!"
             )
         return predict, target
 
 
 class L1Loss(Loss):
-    """L1Loss"""
+    """
+    L1Loss:
```
woodsp-ibm, line 96: The name of the class will be in the docs from the class itself, it does not need to be put into the docstring as it will not really add any value. All it needs is the description of what this class is about ie. the `L1Loss:` bit can simply be removed. Same in the other classes in this module.

qiskit_machine_learning/utils/loss_functions/loss_functions.py
adekusar-drl: > 
> 
> So this is now either an integer or an numpy array. The args below mention just the numpy array not a single value. I see the single value is reformed to a numpy array. Does that mean the numpy arrays are arrays of ints too? I.e the predicted and target values are all ints. Maybe the Args docstring can be improved.

No more ints, just numpy arrays.

qiskit_machine_learning/utils/loss_functions/loss_functions.py
adekusar-drl: > 
> 
> It would be good to have a Returns here to state what the float return is

Added a Returns section.

qiskit_machine_learning/utils/loss_functions/loss_functions.py
adekusar-drl: > 
> 
> Again it would be good to have a Returns statement here for the float return

Same here, added Returns.

qiskit_machine_learning/utils/loss_functions/loss_functions.py
adekusar-drl: > 
> 
> So if I pass in an 'int' and an 'int' as predict and target, I will not get back an Tuple of ints rather a Tuple of arrays? I guess its doing a type cast as well as validate that the other methods rely on. Not sure if it would be helpful to expose the duality in the name say `_cast_validate` or `_to_array_validate` - its a private method so maybe its less important.

Now only numpy arrays, renamed to `_validate_shapes`. I hope it is good enough now.

qiskit_machine_learning/utils/loss_functions/loss_functions.py
adekusar-drl: > 
> 
> The name of the class will be in the docs from the class itself, it does not need to be put into the docstring as it will not really add any value. All it needs is the description of what this class is about ie. the `L1Loss:` bit can simply be removed. Same in the other classes in this module.

Agree, docstrings are updated.

woodsp-ibm (approved):

PR #73, opened 2021-04-26
Compute gradients with respect to input data only if they are required

### LINKED ISSUE #44: Compute gradients only when they are required

Compute gradient w.r.t. input data only when it is required. Now all gradients are computed at every time.

adekusar-drl:
### Summary
This PR introduces a new property `input_gradients` in the `NeuralNetwork` class and corresponding logic to handle optional computation of gradients with respect to input data in the subclasses: `CircuitQNN` and `OpflowQNN`. The PR closes #44.

### Details and comments
By default input gradients are not computed and the property `input_gradients` is set `False`. The only case where they are required is `TorchConnector`.


qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -167,6 +177,18 @@ def quantum_instance(self, quantum_instance) -> None:
         elif len(self._circuit.clbits) == 0:
             self._circuit.measure_all()
 
+    @property
+    def input_gradients(self) -> bool:
```
woodsp-ibm, line 181: Is this purely an internal performance optimization or something that a end user creating the neural network might want to set too. Either way I think it should have a reno to note the performance aspect etc.

qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -192,6 +192,10 @@ def __init__(self, neural_network: NeuralNetwork,
         """
         super().__init__()
         self._neural_network = neural_network
+
+        # turn on input gradients if they are off by default
+        self._neural_network.input_gradients = True
```
woodsp-ibm, line 197: Do we expect users to possible re-use neural networks. As soon as you use it with the torch connector here it is going to have its flag forced True - so it it started out False the network will no longer behave as it did before if re-used elsewhere - if I understand things

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -68,51 +68,61 @@ def __init__(self, circuit: QuantumCircuit,
...
-                         output_shape_)
-
     def _compute_output_shape(self, interpret, output_shape, sampling) -> Tuple[int, ...]:
         """Validate and compute the output shape."""
 
         # this definition is required by mypy
         output_shape_: Tuple[int, ...] = (-1,)
         if sampling:
-            num_samples = self.quantum_instance.run_config.shots
+            num_samples = self._quantum_instance.run_config.shots
```
woodsp-ibm, line 125: Is there a rationale to change this from accessing the quantity via the public property to using the private variable that is currently backing it? Same for `self.circuit` change below

qiskit_machine_learning/neural_networks/opflow_qnn.py
```diff
@@ -134,7 +155,15 @@ def _backward(self, input_data: Optional[np.ndarray], weights: Optional[np.ndarr
 
         # iterate over rows, each row is an element of a batch
         batch_size = input_data.shape[0]
-        grad_all = np.zeros((batch_size, *self.output_shape, self.num_inputs + self.num_weights))
+        if self._input_gradients:
+            grad_all = np.zeros((batch_size,
+                                 *self._output_shape,
+                                 self._num_inputs + self._num_weights))
+        else:
```
woodsp-ibm, line 162: Pretty minor comment: Since the the body of the if's is so similar, only in one case it has the extra self._num_inputs would it better to set up the third parameter value based on the if and just have the one np.zeros block. Saves someone checking back and forth across the if/else looking for the difference.

qiskit_machine_learning/neural_networks/circuit_qnn.py
adekusar-drl: This is an internal performance optimization. A user may play with it, but at their own risk. A reno file is added.

qiskit_machine_learning/connectors/torch_connector.py
adekusar-drl: Yes, a network maybe and probably should be re-used. The problem is that it is hard to trace when a network is not used any more and gradients may be turned off again. Management of gradients is up to a upper level class in the hierarchy, e.g. a classifier, regressor, or torch connector. But even in this case it is not obvious. I was thinking of this problem when I was writing the code. Unfortunately I don't have a good solution here.

qiskit_machine_learning/neural_networks/circuit_qnn.py
adekusar-drl: Only one: if possible, reduce a number of abstractions. In case of `self._*` I know this is a plain old private class member. In case of a property you never know what is built in it, e.g. properties of `BlueprintCircuit`, this is an example of how properties should NOT be designed in my opinion.

qiskit_machine_learning/neural_networks/opflow_qnn.py
adekusar-drl: Agree. Looks better.

qiskit_machine_learning/neural_networks/circuit_qnn.py
woodsp-ibm: Arguably the getter, like it is here, should not have any side-effects, unlike a setter where sometimes other things are invalidated or changed via the property setter that will not happen if you use the variable that is backing it. I have seen things fail because of this when the property should ideally have been used consistently. If a property exists my choice is normally to use it unless we have to avoid any side-effect.

qiskit_machine_learning/neural_networks/circuit_qnn.py
woodsp-ibm: Ok, being public it will show up in API reference hence why I asked.

woodsp-ibm: Would `backward_gradients` be a more specific name here rather than input_gradients. The text seems to suggest that it only applies to backwards hence the question

adekusar-drl: `backward_gradients` does not work. There are two types: gradients wrt input and wrt weights. Both are computed in the backward pass. Weights are always computed.

qiskit_machine_learning/neural_networks/circuit_qnn.py
adekusar-drl: A lesson learnt from `BlueprintCircuit` is that even in a getter can be some logic with side effects, e.g. `parameters`. Properties are for public access, internally, in the code, there should not be a difference and plain member access is just simpler.

woodsp-ibm: With a mix you just have to be more careful evidently - and people changing the code have to take care. As to simpler, arguably I think its simpler if they all go through the getter/setter, rather than a mix. As I mentioned I have seen things go wrong because of direct access when side-effect was desired. Seems you have too but in an opposite way. Either way with the code doing it 2 different ways one just has to take more care.

adekusar-drl: Oh... I see your point now. Why I did not notice this earlier. There should be a consistent way of accessing class members. I will review the code! Thanks for spotting it!

adekusar-drl: Now class member access is consistent.

woodsp-ibm (approved):

Issue #70, opened 2021-04-26
Error in gradient calculation in qGAN using Gradient() 


bonavia1997:
### Information

- **Qiskit Machine Learning version**: 0.1.0 
- **Python version**: 3.8.8
- **Operating system**: Windows 10 

### What is the current behavior?

I was trying to apply "Phase shift rule" for gradient calculation in quantum GAN using the tutorial :
https://qiskit.org/documentation/tutorials/machine_learning/04_qgans_for_loading_random_distributions.html.

To do so, I replaced the line :
```python 
qgan.set_generator(generator_circuit=g_circuit, generator_init_params=init_params)
```
by
```python 
qgan.set_generator(generator_circuit=g_circuit, generator_init_params=init_params, generator_gradient = Gradient())
```
with `qiskit.aqua.operators.gradients.Gradient`.

However, while running 
```python 
result = qgan.run(quantum_instance)
```  
with ` reps = 2` (2 blocks in the quantum generator), the code shows me an error message below :

```bash
ValueError                                Traceback (most recent call last)
~\Anaconda3\lib\site-packages\qiskit\aqua\components\neural_networks\quantum_generator.py in loss(self, x, weights)
    382             # pylint: disable=no-member
--> 383             loss = (-1) * np.dot(np.log(x).transpose(), weights)
    384         except Exception:  # pylint: disable=broad-except
<__array_function__ internals> in dot(*args, **kwargs)
ValueError: shapes (4,1) and (6,4) not aligned: 1 (dim 1) != 6 (dim 0)
```

Please also find the full error message below 

 <img src="https://github.com/bonavia1997/datarep/blob/main/error_message.png?raw=true">.

### Steps to reproduce the problem
Please also find the code [here](https://github.com/bonavia1997/datarep/blob/main/qgan_phase_shift.ipynb). 

In the tutorial [04_qgans_for_loading_random_distributions.ipynb](https://qiskit.org/documentation/tutorials/machine_learning/04_qgans_for_loading_random_distributions.html),
 
1) Import gradient object :  
```python 
from qiskit.opflow import Gradient
```

2) Replace the line  :
```python 
qgan.set_generator(generator_circuit=g_circuit, generator_init_params=init_params)
```
by

```python  
qgan.set_generator(generator_circuit=g_circuit, generator_init_params=init_params, generator_gradient = Gradient())`
```

3)  Replace the line : 

```python
[...]

PR #77, opened 2021-04-27
Fix qGAN gradient behavior

Zoufalc:
### Summary

Address Issue #70 by fixing the dimensions of the gradients & include an additional gradient test for the statevector backend.


test/algorithms/distribution_learners/qgan/test_qgan.py
```diff
@@ -266,14 +266,18 @@ def test_qgan_training_run_algo_numpy_multivariate(self):
...
         """Test QGAN training with analytic gradients"""
-        self.qgan.set_generator(self.generator_circuit)
-        numeric_results = self.qgan.run(self.qi_qasm)
-        self.qgan.set_generator(self.generator_circuit,
-                                generator_gradient=Gradient('param_shift'))
-        analytic_results = self.qgan.run(self.qi_qasm)
-        self.assertAlmostEqual(numeric_results['rel_entr'],
-                               analytic_results['rel_entr'], delta=0.1)
-
+        for backend in ['qasm', 'sv']:
```
woodsp-ibm, line 269: It would be preferable to use `ddt` here so the backend ends up being passed to test so we run two independent tests instead of one that will stop at the first failure in the loop .

woodsp-ibm:
@Zoufalc Are you able to sign the CLA here. When code was moved the CLA database should have moved with it and everyone that originally signed it still should have been signed up here. But some people ended up not being recognized as having signed the CLA, and you were among them it seems. But hopefully it should be as simple as re-signing it again and it will get you recognized once more as having signed it.

qgan_gradient_fix-6d2dd8f1b471d8c0.yaml
```diff
@@ -0,0 +1,3 @@
+fixes:
```
woodsp-ibm, line 1: This reno note needs to be moved to replace the generic template one of the same note in the releasenotes/notes folder

Zoufalc: Sure, will do.

woodsp-ibm (approved):

PR #79, opened 2021-04-30
Added getter/setter, exception handling for OpflowQNN, CircuitQNN 

### LINKED ISSUE #75: Verify `CircuitQNN` and `OpflowQNN` work correctly when no `QuantumInstance` is passed.
The parameter `quantum_instance` in the `CircuitQNN` and `OpflowQNN` may have value of `None`. In such cases there should be additional checks:

- Add `if` statements to be sure code runs smoothly when no quantum instance is passes and it is not required at this moment.
- If the value is required, then raise an appropriate `QiskitMachineLearning` exception.
- Add getters/setters for quantum instance.

stephenmurray2:
### Summary

This PR closes #75:
- Added getter and setter to `OpflowQNN`
- In `CircuitQNN`, raises errors in methods for which a `quantum_instance` is required

### Details and comments

My first PR in this repo, still figuring things out :)


qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -69,11 +69,14 @@ def __init__(self, circuit: QuantumCircuit,
...
         if isinstance(quantum_instance, (BaseBackend, Backend)):
             quantum_instance = QuantumInstance(quantum_instance)
-        self._quantum_instance = quantum_instance
-        self._sampler = CircuitSampler(quantum_instance, param_qobj=False, caching='all')
+        if quantum_instance:
+            self._quantum_instance = quantum_instance
+            self._sampler = CircuitSampler(quantum_instance, param_qobj=False, caching='all')
+        else:
+            self._quantum_instance = None
+            self._sampler = None
```
woodsp-ibm, line 80: This is probably done simpler as
```
        if quantum_instance is not None:
            self._sampler = CircuitSampler(quantum_instance, param_qobj=False, caching='all')
```

Also the test should be against None explicitly and not try to treat the object as boolean. I see some original code ought to be done that way too.

I would ask now what happens in the case of None and self._sampler is not created. I imagine this ought to be done in the quantum_instance setter. Maybe it should always be done in the setter so that there is a new circuit sampler running on the instance just being provided rather than the old one running on whatever the quantum instance was before.

And again on the existing code the quantum_instance setter should have the same signature as the constructor i.e. they should be more like this https://github.com/Qiskit/qiskit-terra/blob/3b86cda6923809067588a706fda7a411e8c05264/qiskit/algorithms/variational_algorithm.py#L84-L95 which you can see logic in the setter similar to the constructor here in wrapping the backend

qiskit_machine_learning/neural_networks/opflow_qnn.py
```diff
@@ -80,6 +80,16 @@ def __init__(self, operator: OperatorBase,
         super().__init__(len(self._input_params), len(self._weight_params),
                          sparse=False, output_shape=output_shape)
 
+    @property
```
woodsp-ibm, line 83: I think the comments here are similar. The setter should allow Backend... and wrap it like in the constructor. Also I think it again needs to take care of CircuitSampler, that is setup in the constructor if the quantum instance is updated we need to redo that I think off the new quantum instance

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -208,6 +213,10 @@ def _sample(self, input_data: Optional[np.ndarray], weights: Optional[np.ndarray
 
     def _probabilities(self, input_data: Optional[np.ndarray], weights: Optional[np.ndarray]
                        ) -> Union[np.ndarray, SparseArray]:
+
+        if not self._quantum_instance:
```
woodsp-ibm, line 217: Should be `if self._quantum_instance is not None:` for an explicit test

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -112,6 +115,8 @@ def _compute_output_shape(self, interpret, output_shape, sampling) -> Tuple[int,
         # this definition is required by mypy
         output_shape_: Tuple[int, ...] = (-1,)
         if sampling:
+            if not self._quantum_instance:
```
woodsp-ibm, line 118: `if self._quantum_instance is not None:`

woodsp-ibm (commented):
Welcome and thanks for looking at this. I made some comments in regards of the aspects I know about this code.

qiskit_machine_learning/neural_networks/circuit_qnn.py
adekusar-drl: I agree with @woodsp-ibm:

- There should be a new test that tests `CircuitQNN` with `quantum_instance` set to `None`.
- In the setter of `quantum_instance` we should recreate the sampler.
- Type hints should be consistent. I know, there's no type hint in the setter, but this is a mistake. Ideally, the code in the constructor and in the setter should not be duplicated.

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -208,6 +213,10 @@ def _sample(self, input_data: Optional[np.ndarray], weights: Optional[np.ndarray
 
     def _probabilities(self, input_data: Optional[np.ndarray], weights: Optional[np.ndarray]
                        ) -> Union[np.ndarray, SparseArray]:
+
+        if not self._quantum_instance:
+            raise QiskitMachineLearningError(
+                'Evaluation of probabilities requires a quantum instance!')
```
adekusar-drl, line 219: If don't miss anything such checks should be in `_sample()`, `_probabilities()`, and `_probability_gradients()`. I guess `self._sampler.convert()` will raise an exception if `q_i` is `None`.

adekusar-drl (commented):
Thanks a lot for taking care of this, really appreciated!

stephenmurray2:
@woodsp-ibm @adekusar-drl Thanks so much for the feedback, will address these comments shortly

qiskit_machine_learning/neural_networks/circuit_qnn.py
stephenmurray2: @woodsp-ibm should it be
```python
if self._quantum_instance is None:
     raise QiskitMachineLearningError("Quantum instance required")
     
if self._quantum_instance.is_statevector:
     raise QiskitMachineLearningError("Sampling does not work with statevector simulator!")
...
```
(so we avoid executing the remainder of the function if no quantum instance)

woodsp-ibm: From my understanding there is _sample and _probabilities. The _sample method, for sampling, does not work with statevector, but either require a quantum_instance hence the split in the testing.

stephenmurray2: @woodsp-ibm thanks, will address this shortly

stefan-woerner (dismissed):
lgtm

qiskit_machine_learning/neural_networks/circuit_qnn.py
woodsp-ibm: Would it not be better, now that the quantum instance setter takes care of creating a CircuitSampler etc, that the constructor here leveraged that code i.e used the setter directly. It might mean moving the code around a little to set self._circuit earlier so that logic around measurements will again be handled by the setter and not needed in the constructor. Better from the viewpoint that we just have one copy of the code to maintain/fix etc rather than what seems to be duplicate of code in constructor and the setter, that could be avoided.

qiskit_machine_learning/neural_networks/opflow_qnn.py
```diff
@@ -114,6 +114,24 @@ def _construct_gradient_operator(self):
         except (ValueError, TypeError, OpflowError, QiskitError):
             logger.warning("Cannot compute gradient operator! Continuing without gradients!")
 
+    @property
+    def quantum_instance(self) -> QuantumInstance:
+        """Returns the quantum instance to evaluate the circuit."""
+        return self._quantum_instance
+
+    @quantum_instance.setter
```
woodsp-ibm, line 122: I think the same applies here where this setter and code in the constructor are duplicated. Or almost. In the setter here if its not passed a QuantumInstance, but rather a [Base]Backend it does not make a QuantumInstance out of it, like I think it should, as such the getter will return a [Base]Backend which is not what we want I think. I note that CircuitSampler, like we normally do, allows a QuantimInstance or {Base]Backend to be passed in so the code works in that regard, but it should be wrapped to a QuantumInstance if its not already, like the equivalent code in the constructor does.

Arguably there is a bug in the constructor code that the quantum_instance test should be explicit to be is not None, like is done here.

adekusar-drl:
@stephenmurray2 Would you be able to finish the PR? There's just one open comment left.

stephenmurray2:
@adekusar-drl yes! will address today

qiskit_machine_learning/neural_networks/opflow_qnn.py
woodsp-ibm: In case the 'I think the same applies here....' is not clear I had made it while appending a comment to one or the review bits above on circuit_qnn here https://github.com/Qiskit/qiskit-machine-learning/pull/79#discussion_r644339176  I see these comments are far apart now; at least on the page - but they were done at the same time as you can see from the timestamps.

stephenmurray2: I noticed that we were assigning `_sampler` in the constructor, but `_circuit_sampler` in the setter, which I think was an error. In the refactored version we assign just `self._circuit_sampler` as a `CircuitSampler` instance, let me know if that's incorrect. 

Also I assume we wish to similarly wrap a `Backend` in a `QuantumInstance` within the setter of `circuit_qnn.py` as well, so I made the change there as well

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -191,18 +181,25 @@ def quantum_instance(self) -> QuantumInstance:
...
-
-        # add measurements in case non are given
-        if quantum_instance.is_statevector:
-            if len(self._circuit.clbits) > 0:
-                self._circuit.remove_final_measurements()
-        elif len(self._circuit.clbits) == 0:
-            self._circuit.measure_all()
+        if quantum_instance is not None:
+            if isinstance(quantum_instance, (BaseBackend, Backend)):
+                quantum_instance = QuantumInstance(quantum_instance)
```
woodsp-ibm, line 193: He, if we have a backend, as getter returns just a QuantumInstance type we should store the one we create here in self._quantum_instance so the getter can pass it back. Eg. see this https://github.com/Qiskit/qiskit-terra/blob/d2098068d31f4c04e7dee228a27951a3dde3b4f9/qiskit/algorithms/variational_algorithm.py#L97-L103

woodsp-ibm (commented):
Thanks for making the change to use the setter. I think things look much cleaner now. Just a couple of small points.

Also I think we should have reno to describe this new feature of these networks. I.e. the getter/setter allowing quantum instance to be set/updated later.

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -147,7 +137,7 @@ def _compute_output_shape(self, interpret, output_shape, sampling) -> Tuple[int,
 
         # this definition is required by mypy
         output_shape_: Tuple[int, ...] = (-1,)
-        if sampling:
+        if sampling and self._quantum_instance is not None:
```
woodsp-ibm, line 140: The `_compute_output_shape` is called in constructor and if a valid quantum instance is given then the this code is run - the resultant shape is then passed to super constructor.

Now if QuantumInstance is None this code will not be executed. Should we be executing this later and somehow setting/updating the base class that stores output shape once a valid quantum instance is given via its setter.

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -191,18 +181,25 @@ def quantum_instance(self) -> QuantumInstance:
         return self._quantum_instance
 
     @quantum_instance.setter
-    def quantum_instance(self, quantum_instance) -> None:
+    def quantum_instance(
+        self, quantum_instance: Optional[Union[QuantumInstance, BaseBackend, Backend]]
+    ) -> None:
```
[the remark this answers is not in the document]
stephenmurray2: @woodsp-ibm if the argument must be a valid `quantum_instance`, should we set `self._sampler` in the constructor? so:
```python
     self._quantum_instance = None
     self._sampler = None
     if quantum_instance is not None:
         self.quantum_instance = quantum_instance
```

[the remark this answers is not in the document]
woodsp-ibm: All instance variables should ideally be defined in the constructor - allows the class/instance vars to be quickly understood since its all in one place for readability. So yes they should be defined in the constructor - even when the code was like it was before. Often an IDE will highlight - mine does - if an instance var you are referring to is not defined in the constructor.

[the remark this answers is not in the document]
stephenmurray2: @woodsp-ibm ok thanks. Will address shortly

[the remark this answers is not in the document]
woodsp-ibm: @stephenmurray2 Ok, we have a release next week that we want to have this in, so it really needs to be finished off this week which hopefully fits in with your 'shortly' timeframe.

qiskit_machine_learning/neural_networks/circuit_qnn.py
stephenmurray2: I don't quite follow... do you mean we should call `_compute_output_shape` within the setter (in addition to within the constructor?)

woodsp-ibm: There is an aspect  of `_compute_output_shape` that requires a QuantumInstance. In the constructor if that is None, and the method is called from there, the output shape it returns, if it should have gone via the QuantumInstance is not None path it will not. Hence once we do have a valid QuantumInstance I was thinking that method should be called and output_shape in the base class updated. Since passing in valid quantum instance is done later via the setter then it would be called from there - and I think we should do it even if there was a previous valid QuantumInstance since shots, which is used to compute output shape, may have been changed.

I think calling the compute_output_shape from both places is ok. In the constructor the output_shape gets passed up via the super.init, and if sampling and no QuantInst then it will be whatever the default is returned from that method which hopefully satisfies the base NeuralNetwork class that ultimately stores it. Later when the QI is updated, we can recompute the output shape from the setter, and update the base class. Since there is no setter we probably need to update _output_shape of the base class directly - though I note when going through the constructor, there is some sort of validity check too. Maybe thats not needed - if we think it is then possible make a private method to validate/set the output shape, that the constructor calls, and that we can call from a sub-class, in this case in the QI setter to update the shape.

Does that help?

stephenmurray2:
@woodsp-ibm I'm having some issues getting unit tests to pass for `test_circuit_qnn`. I even checked out an old commit which was passing on CI (`ccf4748`) and I ran `make test` and I am getting some failures. Any ideas?

woodsp-ibm:
> and I ran make test and I am getting some failures. Any ideas?

Did you remove a logging warning about output shape? This PR #127 was done, while this has been in progress, to help users understand what happens around output shape behavior. The unit tests seem related to that from what I can see which check that the warning gets emitted as expected.

I fixed the merge conflict between the file in that PR and here, which included the warning into the code here, and as far as I recall that was working fine then.

stephenmurray2:
@woodsp-ibm I see that the version of qiskit-terra has been bumped to `0.18.0`, but this version isn't yet available pip. Do I have to build it from the source?

woodsp-ibm:
In general for development the CI here, and even locally, we run/test against the latest Terra from their main branch. When we checked recently there are some failures against the current release hence we bumped the dependency version in readiness for the release. You can simply clone and install Terra from its repo - this has more info in case its of use https://qiskit.org/documentation/contributing_to_qiskit.html#installing-from-source

qiskit_machine_learning/neural_networks/circuit_qnn.py
stephenmurray2: Ok thanks, so to make sure I understand:

We add a new function to the `NeuralNetwork` class, which handles the validation of the output shape:
```python
    def _validate_output_shape(self, output_shape):
        if isinstance(output_shape, int):
            output_shape = (output_shape,)
        if not np.all([s > 0 for s in output_shape]):
            raise QiskitMachineLearningError(
                f"Invalid output shape, all components must be > 0, but got: {output_shape}."
            )
```
 and within the constructor of `NeuralNetwork`, we just replace that validation with a call to this function. Within `CircuitQNN`, we add the following at the end of the QI setter:
 ```python
         output_shape = self._compute_output_shape(self._interpret, self._output_shape, self._sampling)
        self._validate_output_shape(output_shape)
        self._output_shape = output_shape
 ```
 is this on the right track?

woodsp-ibm: Seems like thats the right direction. Maybe the method should return output shape so its that which is set into the instance. Of course that only happens when its an `int` but it will allow the base class constuctor and you to use it the same way as seems your intent.

stephenmurray2:
> In general for development the CI here, and even locally, we run/test against the latest Terra from their main branch. When we checked recently there are some failures against the current release hence we bumped the dependency version in readiness for the release. You can simply clone and install Terra from its repo - this has more info in case its of use https://qiskit.org/documentation/contributing_to_qiskit.html#installing-from-source

Thanks, checking out the Qiskit Terra repo and installing the package from the source resolved the issue 👍

adekusar-drl:
@woodsp-ibm please take a look, the code is heavily revised, you may consider it as a brand new version.

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -153,24 +151,33 @@ def _construct_gradient_circuit(self):
...
+                result = np.array(ret)
+                if len(result.shape) == 0:
+                    output_shape_ = (num_samples, 1)
+                else:
+                    output_shape_ = (num_samples, *result.shape)
             else:
-                output_shape_ = (num_samples, *result.shape)
+                output_shape_ = (num_samples, 1)
         else:
             if interpret:
```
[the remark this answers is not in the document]
adekusar-drl: Updated `if interpret`

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -243,12 +297,22 @@ def set_interpret(self, interpret, output_shape=None):
         """Change 'interpret' and corresponding 'output_shape'. If self.sampling==True, the
         output _shape does not have to be set and is inferred from the interpret function.
         Otherwise, the output_shape needs to be given."""
-        self._interpret = interpret if interpret else lambda x: x
+
+        # save original values
+        self._original_output_shape = output_shape
```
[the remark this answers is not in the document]
adekusar-drl: Agree. Type hints are added.

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -208,24 +219,67 @@ def weight_params(self) -> List:
         """Returns the list of trainable weights parameters."""
         return self._weight_params
 
+    @property
+    def interpret(self) -> Optional[Callable[[int], Union[int, Tuple[int, ...]]]]:
+        """Returns interpret function to be used by the neural network.
+        If no interpret function was passed at the construction time, then an identity function
+        is used by this neural network and returned by this property."""
```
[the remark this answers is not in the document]
adekusar-drl: You are right, the behavior is complex now. I'll try to get back to the code to review it and simplify it.

adekusar-drl:
@woodsp-ibm could please take a look?

qiskit_machine_learning/neural_networks/circuit_qnn.py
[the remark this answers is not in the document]
woodsp-ibm: In terms of this getter, I did not see anything in regards of the recent set of changes that have any simplification. It seems that this `If no interpret function was passed at the construction time, then an identity function  is used by this neural network and returned by this property.` is still not quite right. It seems that until such time as a QuantumInstance is passed, whereby set_interpret will run, that this will return None (and not an identity fn). I imagine this getter could return an identity fn if self._interpret was None here. It seems that for now though the code and docstring do not quite align.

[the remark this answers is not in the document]
adekusar-drl: I have updated the docstrings. I hope, it is clearer now.

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -77,7 +77,8 @@ def __init__(
             interpret: A callable that maps the measured integer to another unsigned integer or
                 tuple of unsigned integers. These are used as new indices for the (potentially
                 sparse) output array. If this is used, and ``sampling==False``, the output shape of
-                the output needs to be given as a separate argument.
+                the output needs to be given as a separate argument. If no interpret function is
+                passed, then an identity function will used by this neural network.
```
dlasecki, line 81:
```suggestion
                passed, then an identity function will be used by this neural network.
```

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -208,24 +220,67 @@ def weight_params(self) -> List:
...
         self._quantum_instance = quantum_instance
 
-        # add measurements in case non are given
-        if quantum_instance.is_statevector:
-            if len(self._circuit.clbits) > 0:
-                self._circuit.remove_final_measurements()
-        elif len(self._circuit.clbits) == 0:
-            self._circuit.measure_all()
+        if self._quantum_instance is not None:
+            # add measurements in case non are given
```
dlasecki, line 270:
```suggestion
            # add measurements in case none are given
```

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -239,16 +294,41 @@ def input_gradients(self, input_gradients: bool) -> None:
...
+
+        Args:
+            interpret: A callable that maps the measured integer to another unsigned integer or
+                tuple of unsigned integers. See constructor for more details.
+            output_shape: The output shape of the custom interpretation, only used in the case
+                where an interpret function is provided and ``sampling==False``. See constructor
+                for more details.
+
+        Returns:
+            None.
```
dlasecki, line 314: Not necessary because of the typehint. Please check other places.

qiskit_machine_learning/neural_networks/opflow_qnn.py
```diff
@@ -118,18 +104,18 @@ def _construct_gradient_operator(self):
...
         if isinstance(op, ComposedOp):
-            return self._get_output_shape_from_op(op.oplist[0].primitive)
+            return self._compute_output_shape(op.oplist[0].primitive)
         # this "if" statement is on purpose, to prevent sub-classes.
         # pylint:disable=unidiomatic-typecheck
         if type(op) == ListOp:
             shapes = []
             for op_ in op.oplist:
-                shape_ = self._get_output_shape_from_op(op_)
+                shape_ = self._compute_output_shape(op_)
```
dlasecki, line 118:
```suggestion
                shapes = [self._compute_output_shape(op_) for op_ in op.oplist]
```

test/neural_networks/test_circuit_qnn.py
```diff
@@ -300,6 +328,90 @@ def test_circuit_qnn_gradient(self, config):
...
+        (False, False, STATEVECTOR, 1, 1),
+        (False, False, STATEVECTOR, 1, 2),
+        (False, False, STATEVECTOR, 2, 1),
+        (False, False, STATEVECTOR, 2, 2),
+        (False, False, QASM, 0, 1),
+        (False, False, QASM, 0, 2),
+        (False, False, QASM, 1, 1),
+        (False, False, QASM, 1, 2),
+        (False, False, QASM, 2, 1),
+        (False, False, QASM, 2, 2),
```
dlasecki, line 379: So many lines of data could be extracted to a separate file.

dlasecki (changes requested):

qiskit_machine_learning/neural_networks/circuit_qnn.py
adekusar-drl: Removed the return statements.

qiskit_machine_learning/neural_networks/opflow_qnn.py
adekusar-drl: Updated. Let's focus on the changes introduced in the PR.

test/neural_networks/test_circuit_qnn.py
adekusar-drl: Agree, but left as it is now. There should a consolidated change across all tests. And I'm not sure moving out such things to a separate file will solve the issue. Test parameters will be hidden, so I'm not convinced. Could be a dedicated function with a lot of nested loops better. But to be discussed, I think.

qiskit_machine_learning/neural_networks/circuit_qnn.py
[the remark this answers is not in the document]
woodsp-ibm: I think something is missing

>  """Returns interpret function to be used by the neural network. If it is not set in
        the constructor or can not be implicitly derived (e.g. a quantum instance is not), then
        ``None`` is returned."""
        
from this `(e.g. a quantum instance is not)` I imagine perhaps `(e.g. a quantum instance is not provided)`

test/neural_networks/test_circuit_qnn.py
woodsp-ibm: Since it looks like the whole data set there for the test is easily computed, just combinations of the parameters, it seems like some sort of simplification should be easily doable in the future at some point if desired.

woodsp-ibm (approved):

qiskit_machine_learning/neural_networks/circuit_qnn.py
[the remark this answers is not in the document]
adekusar-drl: You are right. The intention was to type "None", but mechanically typed in "not". But I like "not provided" more, so updated.

adekusar-drl:
@dlasecki thanks a lot!
@stephenmurray2 thanks for looking into it. I'm sorry for the "goodfirstissue" label, apparently it was misleading. This issue was not a very good first issue. Appreciated!

PR #80, opened 2021-05-04
Make use of the `score()` methods from Scikit-Learn in NNClassifier and NNRegressor

adekusar-drl:
### Summary
This PR makes use of exiting classes,  `ClassifierMixin` and `RegressorMixin`, from Scikit-Learn for score calculations.

### Details and comments
In the PR we extend `NeuralNetworkClassifier` and `NeuralNetworkRegressor` from the corresponding mixin classes from Scikit-Learn and replace our implementations of the `score` methods with the methods from the base classes. Thus, we reduce the amount of code.

### DIFF
--- a/.pylintdict
+++ b/.pylintdict
@@ -77,6 +77,7 @@ rangle
 readme
 regressor
 regressors
+scikit
 scipy
 semidefinite
 setter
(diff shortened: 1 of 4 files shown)


adekusar-drl:
@stefan-woerner for your attention. The changes are minimal here.

woodsp-ibm:
Should this have a reno yaml file. The method signatures are compatible and I maybe they return the same number so that users would not notice the difference in implementation? Now the scikit learn methods do have one extra param so maybe its helpful to have the change as a reno?

adekusar-drl:
The `score` methods return the same numbers, basically we have the identical implementations except the optional parameter `sample_weights`. That's why I did not change the tests. I added a reno saying that scores are computed by sklearn now.

woodsp-ibm (approved):

PR #81, opened 2021-05-05
Switch to using black for code formatting

manoelmarques:
### Summary
Follows the changes made in Qiskit/qiskit-terra#6361

### DIFF
--- a/.github/actions/run-tests/action.yml
+++ b/.github/actions/run-tests/action.yml
@@ -24,8 +24,11 @@ inputs:
     required: true
 runs:
   using: "composite"
-  steps: 
-    - run: |
+  steps:
+    - name: Run Unit Tests
+      env:
+        PYTHONWARNINGS: default
+      run: |
         # run slow tests only on scheduled event or if input flag is set
         if [ "${{ inputs.event-name }}" == "schedule" ] || [ "${{ inputs.run-slow }}" == "true" ]; then
           export QISKIT_TESTS="run_slow"
@@ -33,6 +36,5 @@ runs:
         if [ "${{ inputs.python-version }}" == "3.7" ]; then
           export PYTHON="coverage3 run --source qiskit_machine_learning --parallel-mode"
         fi
-        export PYTHONWARNINGS=default
         stestr --test-path test run 2> >(tee /dev/stderr out.txt > /dev/null)
       shell: bash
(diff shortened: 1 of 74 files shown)


adekusar-drl:
This is a disaster! I can't say we had a very well formatted code, but this is terrible.
instead of
```python
                    input_grad = grad_output.transpose(0, 1) @ input_grad.transpose(0, 1)
```
now:
```python
                    input_grad = grad_output.transpose(0, 1) @ input_grad.transpose(
                        0, 1
                    )
```

Or:
instead of
```python
                    weights_grad = weights_grad.to_dense()  # this should be eventually removed
```
it is
```python
                    weights_grad = (
                        weights_grad.to_dense()
                    )  # this should be eventually removed
```

Or:
```python
            parameterized_circuit = self._quantum_instance.transpile(parameterized_circuit)[0]
```
replaced with 
```python
            parameterized_circuit = self._quantum_instance.transpile(
                parameterized_circuit
            )[0]
```

and so on and so on...

adekusar-drl:
Any ideas why `max-line-length=105`?

manoelmarques:
Terra changed to 105 from 100 in their PR. I don't know why but would think it has to do on how black formatted code which may have added more lines.

woodsp-ibm (approved):

woodsp-ibm:
This issue in Terra has a note as to the line length of 105 https://github.com/Qiskit/qiskit-terra/pull/5883

I agree it does not do a great job in some places - I made a comment in the equivalent PR in Nature.

adekusar-drl:
Thanks for pointing out to the Nature's PR. I agree with @mtreinish in general. Well, of course we get rid of any formatting questions but at what a cost. 
I'd rather increase max line length to 140-150, we have wide screens for many years, and change the style.

adekusar-drl:
> 
> 
> Terra changed to 105 from 100 in their PR. I don't know why but would think it has to do on how black formatted code which may have increased de size of the code in one line.

This is the answer from https://github.com/Qiskit/qiskit-terra/pull/5883:
> Black treats maximum-line-length "like a speed limit" ie if the least bad option is to exceed the line length by a little, then it will sometimes do so (I kept black using line length 100 and set the pylint/pycodestyle line length to 105 to accommodate this -- there are currently no places where this actually comes up in qiskit)

Issue #82, opened 2021-05-05
Strings like 'l2' and 'l1' would be preferable as 'L2' and 'L1"


woodsp-ibm:
In skimming over the changes that black did I noticed the strings 'l1' and 'l2'. Normally for names lower case l is avoided as it can be mistaken for capital I and even for numeric 1 depending on font. I think it would be preferable to have these as uppercase L, since the docstring even talks about L2 loss where L is in capitals. This is in code like NetworkNeuralClassifier/Regressor among others. And where we test the value, if we choose to accept lower case for compatibility, then maybe do an upper() so it can be compared to 'L2' etc.

adekusar-drl:
I agree in general, but we have other loss functions coded as lower case strings: `cross_entropy`, `cross_entropy_sigmoid`, so I think we need a consistent solution. I suggest replace `l1` with `absolute_error` and `l2` with `squared_error`.
@beichensinn what do you think?

woodsp-ibm:
The strings across the API are case insensitive as it does a lower() on what is passed in to compare against constants in the code in lowercase. If upper() was done instead it would just mean the constants all became upper internally and maybe we'd show them that way in the docstring too. It would not change the API in this case; but changing to different strings means deprecating the l1 and l2 so they'd have to stay in the code for a while. Of course doing this upper() instead, and having it as 'L1' internally and in the docs, the API, would not prevent users from using the 'l1' etc in lowercase in their code and thus have the potential readability problem there still, that different strings would prevent.

mattwright99:
@woodsp-ibm @adekusar-drl I am new to qiskit contributions but I am interested in picking up this issue. Would that be ok?

I think using `upper()` and comparing to `'L1'` internally is a good idea since it would make the code more readible and the docstrings more clear. I think it could also be beneficial to add support for `'absolute_error'` and `'squared_error'` in addition to `'L1'` and `'L2'` respectively.
I also saw a comment in #92 suggesting that we should extract the `if...else` block to a seperate class and I could see that being in the scope of this PR. Do you agree? If so, could you provide some more information on that (e.g. what class should I look at)?
Thanks!

adekusar-drl:
@mattwright99 Surely, you can work on this issue!
I'd prefer to keep string constants in lower case just to keep them more readable in the documentation. I suggest to gracefully deprecate `l1` and `l2` in favor of 'absolute_error' and 'squared_error' cause this would solve the issue completely.
[...]

Issue #63, opened 2021-04-19
Review and add support for batches in `NeuralNetworkClassifier` and `NeuralNetworkRegressor`


adekusar-drl:
### What is the expected enhancement?
In the current implementation there are a few TODOs left in the code that mention batch support. These places should be reviewed and batch support should be enabled.

PR #92, opened 2021-05-10
Cache results of `forward` and introduce a super class for ML models.

### LINKED ISSUE #58: Compute forward pass only once. Re-use the results of forward in backward
Now in `NeuralNetworkClassifier/Regressor` forward pass is computed at least two times, in `objective` and `objective_grad`. This should be changed in a way forward pass is computed only once and the results are stored for re-use in the backward pass.

adekusar-drl:
### Summary
This PR closes #58 by caching `forward` results and introduces a base class `TrainableModel` for ML models and `ObjectiveFunction` as a base class for objective functions optimized in classifier/regressor.

### Details and comments
Since classes `NeuralNetworkClassifier` and `NeuralNetworkRegressor` optimize similar objective functions and the goal of this PR is to compute `forward` pass only once, it is reasonable to pull out these objective functions of the classifier/regressor and add an ability to cache last results of `forward` in the base class `ObjectiveFunction` of the objective functions. In the same way we introduce a base class `TrainableModel` for ML models (classifier and regressor) to reduce the amount of common code and place this common code in the base model class. Names of the base classes are open for discussion.

New classes in the PR:

- `TrainableModel` a base/super class for the NN classifier and regressor. Both of extend this class. This class defines Scikit-Learn interface.
- `ObjectiveFunction` for optimization in the ML models. It has `objective` and `gradient` methods. There are also `BinaryObjectiveFunction`, `MultiClassObjectiveFunction`, and `OneHotObjectiveFunction`.


qiskit_machine_learning/algorithms/trainable_model.py
```diff
@@ -0,0 +1,160 @@
...
+        else:
+            loss = loss.lower()
+            if loss == "l1":
+                self._loss = L1Loss()
+            elif loss == "l2":
+                self._loss = L2Loss()
+            elif loss == "cross_entropy":
+                self._loss = CrossEntropyLoss()
+            elif loss == "cross_entropy_sigmoid":
+                self._loss = CrossEntropySigmoidLoss()
```
dlasecki, line 76: Shall we extract it to a separate class that would handle providing a correct Loss object based on the string? I can imagine that this string-based convention might reappear in many classes and then each of them needs this long block of if statements which leads to code duplication.

qiskit_machine_learning/algorithms/objective_functions.py
```diff
@@ -0,0 +1,158 @@
...
+class OneHotObjectiveFunction(ObjectiveFunction):
+    """
+    An objective function for one hot encoding representation of the output,
+    e.g. classes like ``[1, 0, 0]``, ``[0, 1, 0]``, ``[0, 0, 1]``.
+    """
+
+    def objective(self, weights: np.ndarray) -> float:
+        val = 0.0
+        probs = self._neural_network_forward(weights)
+        # TODO: do batch eval
```
dlasecki, line 146: Was it intended to address the TODO in the current PR?

qiskit_machine_learning/algorithms/objective_functions.py
```diff
@@ -0,0 +1,158 @@
...
+        for i in range(len(self._X)):
+            val += self._loss(probs[i], self._y[i])
+        return val
+
+    def gradient(self, weights: np.ndarray) -> np.ndarray:
+        grad = np.zeros(self._neural_network.num_weights)
+        y_predict = self._neural_network_forward(weights)
+        # TODO: do batch eval for backward instead of zip()
```
dlasecki, line 154: Was it intended to address the TODO in the current PR?

qiskit_machine_learning/algorithms/objective_functions.py
```diff
@@ -0,0 +1,158 @@
...
+        probs = self._neural_network_forward(weights)
+        for i in range(len(self._X)):
+            for y_predict, prob in enumerate(probs[i]):
+                val += prob * self._loss(y_predict, self._y[i])
+        return val
+
+    def gradient(self, weights: np.ndarray) -> np.ndarray:
+        num_classes = self._neural_network.output_shape[0]
+        grad = np.zeros((1, self._neural_network.num_weights))
+        # TODO: do batch eval instead of zip()
```
dlasecki, line 129: Was it intended to address the TODO in the current PR?

qiskit_machine_learning/algorithms/trainable_model.py
adekusar-drl: In general, I agree. But I'd rather wait till the time this situation occurs or address it when #82 will be in progress.

qiskit_machine_learning/algorithms/objective_functions.py
adekusar-drl: No, there's issue #63 to review all of these TODOs.

qiskit_machine_learning/algorithms/objective_functions.py
adekusar-drl: As above, there's #63.

adekusar-drl:
> There are a number of new classes added here and the documentation, nor init file, have been updated to add them as docs or **all**. Shouldn't they be more exposed this way - although others to re-use them as needed etc. Any maybe a reno would be good as I imagine this is intended as a performance improvement

Agree, these are the things I usually forget to do.

qiskit_machine_learning/algorithms/objective_functions.py
```diff
@@ -0,0 +1,158 @@
...
+        Computes and caches the results of the forward pass. Cached values may be re-used in
+        gradient computation.
+
+        Args:
+            weights: an array of weights to be used in the forward pass.
+
+        Returns:
+            The result of the neural network.
+        """
+        data_id = str(id(weights)) + str(id(self._X))
```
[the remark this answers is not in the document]
adekusar-drl: I'd expect a comparison by id to be faster than a direct comparison of arrays. Well, it is not the case now, we have just a few qubits, but in case of classic ML, arrays maybe huge.
It is also a good point whether we need to add `_X` to the comparison. Honestly, I added it only for safety reasons. In general, a comparison of weights is good enough, but just in case. So, if you don't mind I'd keep it.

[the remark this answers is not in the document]
woodsp-ibm: Sure, I was just trying to understand things from whats there. Whether its worth a comment in the code to that effect, i.e. safety just in case, to help any other new readers of the code I'll leave to you.

woodsp-ibm (approved):

PR #102, opened 2021-05-17
Review batch support in NNClassifier/Regressor

### LINKED ISSUE #63: Review and add support for batches in `NeuralNetworkClassifier` and `NeuralNetworkRegressor`

adekusar-drl:
### Summary
This is a continuation of the batch support work. There are a few batch related TODOs left in the objective functions used by `NeuralNetworkClassifier` and `NeuralNetworkRegressor`. Right now these functions don't make use of the batch support added in the neural network implementation and compute all elements in a for loop. The goal of this this PR is to review these functions, add batch support and review loss function implementations as they don't support batches as well.
This PR closes #63.

### Details and comments
Changes:

- Loss functions now takes an array of the shape `(N, *)`, where `N` is the number of samples and return `(N, 1)` for values and a corresponding `(N, *)` array for gradients. Implementation is vectorized.
- Objective functions are vectorized as much as possible.
- Added a test for L1 and L2 loss functions, the test is based on comparison with pytorch. CrossEntropy is not added since it is implemented in pytorch differently (not a one hot encoded function). A test for cross entropy and cross entropy sigmoid to be added separately (this is not a goal of the PR).


test/utils/loss_functions/test_loss_functions.py
```diff
@@ -0,0 +1,81 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2018, 2021.
```
woodsp-ibm, line 3:
```suggestion
# (C) Copyright IBM 2021.
```
Since its a brand new test

woodsp-ibm:
Should this need have a reno at all around the changes - I notice a couple of TODOs get removed, or is this all internal that an end user will not really notice.

adekusar-drl:
> 
> 
> Should this need have a reno at all around the changes - I notice a couple of TODOs get removed, or is this all internal that an end user will not really notice.

Reno is not required, all the changes are internal and a user should not notice anything (unless I introduced a bug).

adekusar-drl:
As agreed, I added a few comments in my changes, @stefan-woerner please take a look.

.pylintdict
```diff
@@ -96,6 +98,8 @@ otimes
 ovo
 ovr
 param
+parameterize
+parameterized
```
woodsp-ibm, line 102: I get that 'paramterize(d)` is valid too - but I think we have usually tried to standardize to the one way, which is  'parametrized` without the e after the t as you see below here.

stefan-woerner (approved):
lgtm

PR #103, opened 2021-05-19
Make sparse dependency optional

### LINKED ISSUE #67: Make `sparse` dependency optional
Currently we use `sparse` for multidimensional sparse arrays/tensors. This library depends on `numba` which, in turns, has compiled code with support/wheels for various machines but not all of them, e.g. if someone has been using a Raspberry Pi 4 and llvmlite, which is installed as part of `numba`, and its not supported there and QML fails to install. This issue propose to make `sparse` as an optional dependency with the default value of `sparse=False`. If someone wanted to use `sparse=True` QML would raise an exception saying it needed to be installed, rather than QML failing to get installed at all.

manoelmarques:
### Summary
Fixes #67

### DIFF
--- a/qiskit_machine_learning/neural_networks/circuit_qnn.py
+++ b/qiskit_machine_learning/neural_networks/circuit_qnn.py
@@ -16,7 +16,22 @@ from numbers import Integral
 from typing import Tuple, Union, List, Callable, Optional, cast, Iterable
 
 import numpy as np
-from sparse import SparseArray, DOK
+
+try:
+    from sparse import SparseArray, DOK
+
+    _HAS_SPARSE = True
+except ImportError:
+    _HAS_SPARSE = False
+
+    class SparseArray:  # type: ignore
+        """Empty SparseArray class
+        Replacement if sparse.SparseArray is not present.
+        """
+
+        pass
+
+
 from scipy.sparse import coo_matrix
 
 from qiskit import QuantumCircuit
@@ -24,6 +39,7 @@ from qiskit.circuit import Parameter
 from qiskit.opflow import Gradient, CircuitSampler, StateFn, OpflowError
 from qiskit.providers import BaseBackend, Backend
 from qiskit.utils import QuantumInstance
+from qiskit.exceptions import MissingOptionalLibraryError
 
 from .sampling_neural_network import SamplingNeuralNetwork
 from ..exceptions import QiskitMachineLearningError, QiskitError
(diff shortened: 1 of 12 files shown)


qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -291,6 +313,12 @@ def _probability_gradients(
         # initialize empty gradients
         input_grad = None  # by default we don't have data gradients
         if self._sparse:
+            if not _HAS_SPARSE:
+                raise MissingOptionalLibraryError(
+                    libname="sparse",
+                    name="DOK",
+                    pip_install="pip install 'qiskit-machine-learning[sparse]'",
+                )
```
adekusar-drl, line 321: Since there are two identical `if` statements in the script, might be better to have a utility method for this check? Or a global utility method for such checks.

qiskit_machine_learning/neural_networks/sampling_neural_network.py
```diff
@@ -16,7 +16,18 @@
...
+try:
+    from sparse import SparseArray
+except ImportError:
+
+    class SparseArray:  # type: ignore
+        """Empty SparseArray class
+        Replacement if sparse.SparseArray is not present.
+        """
+
+        pass
```
adekusar-drl, line 29: This code is present in a few scripts. Any chance we can re-use it somehow? Or keep just one fake definition of `SparseArray`?

manoelmarques: The import and the "fake" SparseArray definitions need to be in the same file though. The "fake" SparseArray class is just because of the typehint definition not throwing errors if sparse is not installed and can only be declared if the real SparseArray is not present.

qiskit_machine_learning/neural_networks/circuit_qnn.py
manoelmarques: Since you need to import sparse at the file either at top level or at the moment it is used, I don't see you gaining much. You would have an import inside some utility to test sparse existence and throw the exception with a custom name, in this case 'DOK'. Then at that place in the code or at top level you still need to import SparseArray/DOK/COO and make sure to let the ImportError pass so that you can use it and then call the utility method. IN the current case, you only have the top import with the _HAS_SPARSE and the tests raising the exception.

qiskit_machine_learning/neural_networks/sampling_neural_network.py
adekusar-drl: What if 
```python
try:
    from sparse import SparseArray
except ImportError:
    from .utils import SparseArray    # there's a fake definition somewhere in qml
```
Or this does not save much as well?

manoelmarques: And then be careful that nobody ever imports utils by mistake if sparse exists. I believe it should not be called utils right ? As it might contain other things and if it contained the fake definition of SparseArray would prevent people from using it unless sparse is not installed. Maybe have a fake_sparse.py file with just this empty class definition and nothing else but then you just created an indirection and had to add a comment to explain what it meant for something that I could see immediately in the first case.

adekusar-drl (approved):

qiskit_machine_learning/neural_networks/sampling_neural_network.py
adekusar-drl: Importing by a mistake is not great at all.

PR #104, opened 2021-05-20
Add Getting Started Documentation

manoelmarques:
### Summary
Add a Getting Started documentation page that 'augments' the main one of Qiskit by providing additional installation information etc specific to this application.


woodsp-ibm (commented):
Looks good, thanks for doing this. Maybe stating that PyTorch enables the TorchConnector when installed is stating the obvious - maybe we can say something like ... and when PyTorch is installed then the TorchConnector facilitates its use of quantum computed networks. Or something like that. That way if you drop into the getting started to install things you at least see the TorchConnector mentioned.

woodsp-ibm (approved):

PR #111, opened 2021-05-31
Replace the `weights` property in the pytorch connector with `weight`

adekusar-drl:
### Summary
This PR deprecates the `weights` property in `TorchConnector` in favor of a new `weight` property which is pytorch compatible. By default, pytorch layers expose `weight` properties to get access to computed weights.


stefan-woerner (dismissed):
lgtm

qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -232,6 +232,18 @@ def neural_network(self):
...
+            "The 'weights' property is deprecated as of 0.2.0 "
+            "and will be removed no sooner than 3 months after the release. "
+            "You should use the 'weight' property instead.",
+            DeprecationWarning,
+        )
+
+        return self._weights
+
+    @property
+    def weight(self):
```
woodsp-ibm, line 246: I guess its a choice, we could rename the private var to _weight to make things perhaps more consistent, though of course is not needed.

Should we add a return typehint here (I see in Torch Linear weight is of type Tensor) Same about nice to have a return type for sparse() below and neural_network() above which I happen to notice don't have it either

adekusar-drl: We have `weights` everywhere in the code, so I prefer to keep the variable name as it is now. In the constructor I'd keep the parameter name as it is now as well. There's no a direct analogy of this constructor in pytorch. But `weight` instead of `weights` is a good change. In this case, if I have a large model with a QNN as one of the layers and I can inspect classical layers and a quantum one in the same fashion.
Type hints are added.

woodsp-ibm (approved):
Just an additional comment; I noticed in looking at the getters that there are no setters. Does say updating weight or sparse after creation make sense that this might be wanted. Can always be another issue/PR if it looks to be reasonable.

adekusar-drl:
> 
> 
> Just an additional comment; I noticed in looking at the getters that there are no setters. Does say updating weight or sparse after creation make sense that this might be wanted. Can always be another issue/PR if it looks to be reasonable.

I don't think we need these setters. For `weight` I'm pretty sure. In general, this is a whole new question how we want to initialize weights. If I recall correctly, in classical pytorch models weights are initialized via applying a certain function to a layer of the model. So here, I'd like to get back to pytorch documentation, especially to the part where they talk about extending pytorch. 
For the `sparse` property I'm not so sure, but I'd rather keep as is now.

PR #117, opened 2021-06-01
Allow input_gradients to be set in constructor of NeuralNetwork

### LINKED ISSUE #90: Introduce a parameter `input_gradients` to NeuralNetwork
As a feedback from #44,  we should add a parameter `input_gradients` to NeuralNetwork to expose it explicitly to users. By default it should be turned off. Also, we should not turn on input gradients explicitly in `TorchConnector` every time, turning on/off gradients is up a user via this new parameter.

jonvet:
### Summary
This PR introduces a parameter ``input_gradients`` to the constructor of ``NeuralNetwork`` and its subclasses. ``input_gradients`` defaults to False. The constructor of ``TorchConnector`` no longer sets ``input_gradients=True`` by default on the ``NeuralNetwork`` it was instantiated with. This PR closes #90.

### Details and comments
 If you are using ``TorchConnector`` and want to compute gradients w.r.t. to inputs, you will have to set ``input_gradients`` explicitly on the ``NeuralNetwork`` you pass to ``TorchConnector``.

Changelog: New Feature

### DIFF
--- a/qiskit_machine_learning/algorithms/classifiers/vqc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/vqc.py
@@ -50,11 +50,8 @@ class VQC(NeuralNetworkClassifier):
             loss: A target loss function to be used in training. Default is cross entropy.
             optimizer: An instance of an optimizer to be used in training.
             warm_start: Use weights from previous fit to start next fit.
-            input_gradients: Returns whether gradients with respect to input data are computed by
-                this neural network in the ``backward`` method or not. By default such gradients
-                are not computed.
+            input_gradients: Determines whether to compute gradients with respect to input data.
             initial_point: Initial point for the optimizer to start from.
-
         Raises:
             QiskitMachineLearningError: Needs at least one out of num_qubits, feature_map or
                 ansatz to be given.
--- a/qiskit_machine_learning/algorithms/classifiers/vqc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/vqc.py
@@ -39,6 +39,7 @@ class VQC(NeuralNetworkClassifier):
         warm_start: bool = False,
         quantum_instance: QuantumInstance = None,
         input_gradients: bool = False,
+        initial_point: np.ndarray = None,
     ) -> None:
         """
         Args:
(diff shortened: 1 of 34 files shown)


qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -49,8 +50,8 @@ def __init__(
             loss: A target loss function to be used in training. Default is cross entropy.
             optimizer: An instance of an optimizer to be used in training.
             warm_start: Use weights from previous fit to start next fit.
+            input_gradients: Determines whether to compute gradients with respect to input data.
```
stefan-woerner, line 53: This can be just set to `False` for VQC. Input gradients make only sense if the QNN is embedded in a larger workflow, VQC has no way to use it (otherwise one could use the NNClassifier and construct it manually).

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -110,6 +111,7 @@ def __init__(
             interpret=self._get_interpret(2),
             output_shape=2,
             quantum_instance=quantum_instance,
+            input_gradients=input_gradients,
```
stefan-woerner, line 114: see previous comment, can just be set to `False`.

stefan-woerner (changes requested):
Thanks! Could you have a look and update the VQC (and accordingly also the VQR which is not included in this PR yet).

jonvet:
all tests pass locally, but some fail on the CI. In particular, in the last build `test_classifier_with_opflow_qnn_4___cobyla____l2____qasm__` failed and yesterday `test_regressor_with_opflow_qnn_3___bfgs____statevector__` failed - but both pass on my laptop. any tips on how to debug this?

adekusar-drl:
> 
> 
> all tests pass locally, but some fail on the CI. In particular, in the last build `test_classifier_with_opflow_qnn_4___cobyla____l2____qasm__` failed and yesterday `test_regressor_with_opflow_qnn_3___bfgs____statevector__` failed - but both pass on my laptop. any tips on how to debug this?

This is a common problem, should be fixed separately, it is not a problem in your code. Thanks for the PR!

adekusar-drl:
@stefan-woerner please review again, the merge is blocked until your approval.

stefan-woerner (approved):
lgtm

adekusar-drl:
@jonvet Thanks a lot for your contribution! Really appreciated!

PR #121, opened 2021-06-07
Fixes batch gradient computation in TorchConnector

### LINKED ISSUE #100: Wrong Gradients when TorchConnector is used with a batch
### Information

- **Qiskit Machine Learning version**: 0.1.0
- **Python version**: 3.8.8
- **Operating system**: MacOS Big Sur

### What is the current behavior?
If a `CircuitQNN` is used with a `TorchConnector`, the result PyTorch model have some issues calculating the gradients of the parameters in the circuit, when the model is evaluated on a batch of data, and not a single sample.

Here is an example to reproduce the problem. I use the same circuitry defined in the tutorial (https://github.com/Qiskit/qiskit-machine-learning/blob/master/docs/tutorials/05_torch_connector.ipynb), using `CircuitQNN` and the `TorchConnector` to create a quantum neural network. I try to evaluate the gradients of the parameters on a regression task with a trivial dataset, consisting of 20 identical inputs and corresponding targets.  

As a loss function, I consider the `MSELoss` with `reduction=sum`, and I try to evaluate the loss and its gradients in different ways:

1. Use PyTorch `MSELoss` on the full dataset (consisting of 20 identical item)
2. Manually evaluate the loss as `(output-target).pow(2).sum()` again on the whole dataset
3. Manually evaluate the single losses for each item in the dataset, and sum them using a for  loop
4. Evaluate the loss only for a single data

Then, the gradients are evaluated using the `loss.backward()` and extracted with `model.weights.grad`.
Note that there is no optimizer step! I only evaluated the gradients without updating the weights.
All these methods should be fully equivalent, since the data are always the same, and there is no seed, ordering, or strange twists. Note that while Methods 1, 2 and 3 use the full dataset of 20 samples, Method 4 uses only a single item, so its gradient is expected to be 20 times smaller (since we are using `MSELoss(reduction="sum")`).

```
### Imports
import numpy as np
import matplotlib.pyplot as plt
from torch import Tensor
import torch
from torch.nn import MSELoss
from torch.optim import SGD
from qiskit  import Aer, QuantumCircuit
from qiskit.opflow import AerPauliExpectation
from qiskit.circuit import Parameter
from qiskit.circuit.library import RealAmplitudes, ZZFeatureMap
from qiskit_machine_learning.neural_networks import CircuitQNN, TwoLayerQNN
from qiskit_machine_learning.connectors import TorchConnector
qi = QuantumInstance(Aer.get_backend('statevector_simulator'))

### Create QNN
num_inputs = 2
feature_map = ZZFeatureMap(num_inputs)
ansatz = RealAmplitudes(num_inputs, entanglement='linear', reps=1)

qc = QuantumCircuit(num_inputs)
qc.append(feature_map, range(num_inputs))
qc.append(ansatz, range(num_inputs))

parity = lambda x: '{:b}'.format(x).count('1') % 2
output_shape = 2  # parity = 0, 1

qnn2 = CircuitQNN(qc, input_params=feature_map.parameters, weight_params=ansatz.parameters, 
                  interpret=parity, output_shape=output_shape, quantum_instance=qi)

# set up PyTorch module
initial_weights = np.array([0.1]*qnn2.num_weights)
model2 = TorchConnector(qnn2, initial_weights)

### Trivial dataset
X = Tensor(np.stack(([0.5, 0.5],)*20))
y = Tensor(np.stack(([-0.5, -0.5],)*20))

### Define optimizer and loss
optimizer = SGD(model2.parameters(), lr = 0.1)
f_loss = MSELoss(reduction = "sum")

### Method 1
output1 = model2(X)
loss1 = (output1-y).pow(2).sum()
optimizer.zero_grad()
loss1.backward()
print("Loss:", loss1) # -> 
print("Gradients:", model2.weights.grad) 
# Loss: tensor(40.0025, grad_fn=<SumBackward0>)
# Gradients: tensor([1.1921e-07, 1.9073e-06, 2.3842e-07, 1.1921e-07])

### Method 2
output2 = model2(X)
loss2 = f_loss(output2, y)
optimizer.zero_grad()
loss2.backward()
print("Loss:", loss2)
print("Gradients:", model2.weights.grad)
# Loss: tensor(40.0025, grad_fn=<MseLossBackward>)
# Gradients: tensor([1.1921e-07, 1.9073e-06, 2.3842e-07, 1.1921e-07])

### Method 3
loss3 = 0.0
for xt, yt in zip(X,y):
    output3 = model2(xt)
    loss3 += f_loss(output3, yt)
optimizer.zero_grad()
loss3.backward()
print("Loss:", loss3)
print("Gradients:", model2.weights.grad)
# Loss: tensor(40.0025, grad_fn=<AddBackward0>)
# Gradients: tensor([-0.0311,  0.3078,  0.0567, -0.0312])

### Method 4
output4 = model2(X[0])
loss4 = f_loss(output4, y[0])
optimizer.zero_grad()
loss4.backward()
print("Loss:", loss4)
print("Gradients:", model2.weights.grad)
# Loss: tensor(2.0001, grad_fn=<MseLossBackward>)
# Gradients: tensor([-0.0016,  0.0154,  0.0028, -0.0016])

```

The gradients should be all equals. In particular, evaluating the loss using a batch of data (Methods 1 and 2) yields vanishing gradients. 
Note that if one substitutes the quantum model created through Qiskit, with a simple `model2 = torch.nn.Linear(2,2)` then the gradients are correctly equals, so the problem is somewhere in the Qiskit's Machine Learning module. 
(to run the code above with the classical linea layer, also substitute `model2.weights.grad` with `model2.weight.grad`).

jonvet:
### Summary
Fixes wrong computation of gradients in TorchConnector when a batch of data is given as input. Closes #100 
Changelog: Bugfix

### DIFF
--- a/qiskit_machine_learning/connectors/torch_connector.py
+++ b/qiskit_machine_learning/connectors/torch_connector.py
@@ -169,6 +169,11 @@ class TorchConnector(Module):
                     input_grad = input_grad.to(grad_output.dtype)
                 else:
                     input_grad = Tensor(input_grad).to(grad_output.dtype)
+
+                # Takes gradients from previous layer in backward pass (i.e. later layer in forward
+                # pass) j for each observation i in the batch. Multiplies this with the gradient
+                # from this point on backwards with respect to each input k. Sums over all i and
+                # j to get total gradient of output w.r.t. each input k.
                 input_grad = einsum("ij,ijk->k", grad_output, input_grad)
 
             if weights_grad is not None:
@@ -184,6 +189,11 @@ class TorchConnector(Module):
                     weights_grad = weights_grad.to(grad_output.dtype)
                 else:
                     weights_grad = Tensor(weights_grad).to(grad_output.dtype)
+
+                # Takes gradients from previous layer in backward pass (i.e. later layer in forward
+                # pass) j for each observation i in the batch. Multiplies this with the gradient
+                # from this point on backwards with respect to each parameter k. Sums over all i and
+                # j to get total gradient of output w.r.t. each parameter k.
                 weights_grad = einsum("ij,ijk->k", grad_output, weights_grad)
 
             # return gradients for the first two arguments and None for the others (i.e. qnn/sparse)
--- a/qiskit_machine_learning/connectors/torch_connector.py
+++ b/qiskit_machine_learning/connectors/torch_connector.py
@@ -30,7 +30,7 @@ from ..exceptions import QiskitMachineLearningError
 logger = logging.getLogger(__name__)
 
 try:
-    from torch import Tensor, sparse_coo_tensor
+    from torch import Tensor, sparse_coo_tensor, einsum
     from torch.autograd import Function
     from torch.nn import Module, Parameter as TorchParam
 except ImportError:
@@ -150,6 +150,10 @@ class TorchConnector(Module):
                     + f" expected input compatible to {neural_network.num_inputs}"
                 )
 
+            # Ensure same shape for single observations and batch mode
+            if len(grad_output.shape) == 1:
+                grad_output = grad_output.view(1, -1)
+
             # evaluate QNN gradient
             input_grad, weights_grad = neural_network.backward(input_data.numpy(), weights.numpy())
             if input_grad is not None:
@@ -165,11 +169,7 @@ class TorchConnector(Module):
                     input_grad = input_grad.to(grad_output.dtype)
                 else:
                     input_grad = Tensor(input_grad).to(grad_output.dtype)
-
-                if len(grad_output.shape) == 2:
-                    input_grad = grad_output.transpose(0, 1) @ input_grad.transpose(0, 1)
-                else:
-                    input_grad = grad_output @ input_grad  # TODO: validate
+                input_grad = einsum("ij,ijk->k", grad_output, input_grad)
 
             if weights_grad is not None:
                 if np.prod(weights_grad.shape) == 0:
@@ -184,11 +184,7 @@ class TorchConnector(Module):
                     weights_grad = weights_grad.to(grad_output.dtype)
                 else:
                     weights_grad = Tensor(weights_grad).to(grad_output.dtype)
-
-                if len(grad_output.shape) == 2:
-                    weights_grad = grad_output.transpose(0, 1) @ weights_grad.transpose(0, 1)
-                else:
-                    weights_grad = grad_output @ weights_grad
+                weights_grad = einsum("ij,ijk->k", grad_output, weights_grad)
 
             # return gradients for the first two arguments and None for the others (i.e. qnn/sparse)
             return input_grad, weights_grad, None, None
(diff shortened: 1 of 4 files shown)


qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -184,11 +184,7 @@ def backward(ctx: Any, grad_output: Tensor) -> Tuple:  # type: ignore
                     weights_grad = weights_grad.to(grad_output.dtype)
                 else:
                     weights_grad = Tensor(weights_grad).to(grad_output.dtype)
-
-                if len(grad_output.shape) == 2:
-                    weights_grad = grad_output.transpose(0, 1) @ weights_grad.transpose(0, 1)
-                else:
-                    weights_grad = grad_output @ weights_grad
+                weights_grad = einsum("ij,ijk->k", grad_output, weights_grad)
```
adekusar-drl, line 187: Could you please a comment here for future reader what einsum does here.

test/connectors/test_torch_connector.py
```diff
@@ -568,6 +572,65 @@ def test_batch_gradients(self):
...
+            output = model(x_i)
+            sum_of_individual_losses += f_loss(output, y_i)
+        optimizer.zero_grad()
+        sum_of_individual_losses.backward()
+        sum_of_individual_gradients = deepcopy(model.weights.grad)
+
+        output = model(x)
+        batch_loss = f_loss(output, y)
+        optimizer.zero_grad()
+        batch_loss.backward()
```
adekusar-drl, line 627: An additional equality assert of `sum_of_individual_losses` and `batch_loss` would be great.

adekusar-drl (commented):
Looks solid! 
@stefan-woerner please take a look as well. I'm happy with the PR.

stefan-woerner:
lgtm, thanks a lot @jonvet!

test/connectors/test_torch_connector.py
jonvet: that's happening on lines 630-632. is that not sufficient?

adekusar-drl: on these line you compare gradients, not loss:
```python
        self.assertAlmostEqual(
            np.linalg.norm(sum_of_individual_gradients - batch_gradients), 0.0, places=4
        )
```
I suggest to add a comparison of loss as well since both values are already computed, something like:
```python
    self.assertAlmostEqual(sum_of_individual_losses, batch_loss)
```

adekusar-drl:
@jonvet please add `codec` and `stdlib` to `.pylintdict`. The spelling problem arises somewhere else, not in your code :) Probably a 3rd party library was updated. This is not the first time I see this spelling problem. Thanks.

test/connectors/test_torch_connector.py
jonvet: ah yes, you're right - my bad

adekusar-drl (approved):
All good, thanks a lot!

woodsp-ibm:
FYI: Just a comment on 'codec' and 'stdlib' - the reason we now need them is they are in QuantumCircuit, which we extend here and whose docstrings will come into play in say RawFeatureVector from the circuit library here. So if aspects of Terra like that we extend are changed - not only might we have code problems, but documentation issues like this spelling too.

PR #127, opened 2021-06-22
Fixed batch gradient warning

ElePT:
### Summary

Fixed warning in `TestTorchConnector.test_circuit_qnn_batch_gradients`

### Details and comments

I found that the dataset for `TestTorchConnector.test_circuit_qnn_batch_gradients` was not correct. I corrected the `output_size` parameter in this test and added warnings to `circuit_qnn.py` to help guide the user with `output_size`.


adekusar-drl (commented):
Thanks a lot, looks good!
Minor things as discussed:
- Update output_shape docstring in `circuit_qnn`.
- Add output_shape/interpret rules to the constructor docstring.
- Update wordings of the warning messages (mentioned by @stefan-woerner).

ElePT:
Thanks for the feedback. I have updated the docstring in `circuit_qnn` to add the constructor rules, and have explained each rule in what I have considered the most relevant argument (most of them are in `output_shape`). As a guidance, here is a compilation of all the rules that I have noticed and the corresponding argument where I have added them (**1.1** was already clarified in the docs, but I added it to the list to have a full overview of how the constructor works):

**Constructor Rules for CircuitQNN:**

1. If sampling==True:
        1.1. The output will always be DENSE (sparse==False), independently
            of the value provided for the 'sparse' argument (it will be overridden). **--> [already in docs] `sampling` argument**
        1.2. If sampling==True and interpret==None, the output shape will be automatically
            calculated as (num_samples,1). **--> [new addition] `output_shape` argument**
        1.3. If sampling==True and interpret!=None, the output shape will be automatically
            inferred from the interpret function as (num_samples, interpret_dim).  **--> [new addition] `output_shape` argument**
        1.4. 'statevector_simulator' is not a valid backend for 'quantum_instance'. **--> [new addition] `quantum_instance` argument**

    2. If sampling==False:
        2.1. If sampling==False and interpret==False, the output shape will be automatically
            calculated as 2^num_qubits, independently of whether the output is dense 
            (sparse==True) or sparse (sparse==False), or whether a custom shape has been
            provided through the 'output_shape' argument. **--> [new addition] `output_shape` argument**
        2.2. If sampling==False and interpret==True, the output shape MUST be provided
            through the 'output_shape' argument, independently of whether the output is dense
            (sparse==True) or sparse (sparse==False).** --> [updated with sampling condition] `output_shape` argument**

If anything isn't clear, please let me know. Thank you!

adekusar-drl (approved):

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -77,15 +77,22 @@ def __init__(
...
+            output_shape: The output shape of the custom interpretation, only used in the case
+                where an interpret function is provided and sampling==False. Note that in the
+                remaining cases, the output shape is automatically inferred by: `2^num_qubits` if
+                sampling==False and interpret==None, `(num_samples,1)` if sampling==True and
+                interpret==None, and `(num_samples,interpret_shape)` if sampling==True and an
+                interpret function is provided.
             gradient: The gradient converter to be used for the probability gradients.
-            quantum_instance: The quantum instance to evaluate the circuits.
+            quantum_instance: The quantum instance to evaluate the circuits. Note that
+                if sampling==True, 'statevector_simulator' is not a valid backend for the
```
woodsp-ibm: Can we avoid the use of a backend name here ie. `statevector_simulator` which is no longer supported for Aer anyway; rather simply say something like: that if sampling is True a statevector simulator is not a valid backend and will cause an error to be raised. Where statevector simulator is more just a description rather than a name - this is more how the error message itself looks.

ElePT: That makes sense, I did take it from the error message... Will do, thanks!

Issue #112, opened 2021-05-31
Add a callback interface to neural networks


adekusar-drl:
### What is the expected behavior?
In the current implementation of the neural networks it is impossible to track loss and other parameters while a model is being trained. Here, we suggest to introduce a callback that can give access some intermediate training data. This callback may make use of the callback interface of optimizers.

darshkaushik:
@adekusar-drl I am interested to fix this issue. 
I suppose the changes are to be made in `CircuitQNN` or any of the base classes? 

>This callback may make use of the callback interface of optimizers.

Also, the callback is available only for some specific optimizers and is not a feature of the class `Optimizers` I think.

adekusar-drl:
I guess the changes should go to `TrainableModel` or the subclasses. `VQE` class may be a good example to look at, it has a `callback` parameter. Anyways, this issue requires additional investigation to have a proper design to implement.

darshkaushik:
@adekusar-drl excuse my inactivity, was a little overwhelmed with `NeuralNetwork` and its subclasses.

> I guess the changes should go to `TrainableModel` or the subclasses. 

I thought `NeuralNetwork` or its subclasses should be changed?

>VQE class may be a good example to look at, it has a callback parameter.

Got some idea from there, but still unclear what specific parameters should `callback` contain.

darshkaushik:
@adekusar-drl excuse my inactivity, was a little overwhelmed with `NeuralNetwork` and its subclasses.
 
> I guess the changes should go to `TrainableModel` or the subclasses.
 
I thought `NeuralNetwork` or its subclasses should be changed?
 
> VQE class may be a good example to look at, it has a callback parameter.

Got some idea from there, but still unclear what specific parameters should `callback` contain.

adekusar-drl:
No worries at all, I did not say it would be easy. I don't know to what extent you understand implementation of `VQE`, so let me get to basics. Feel free to skip if it is clear.

[...]

PR #132, opened 2021-06-24
This fixes missing optimizer in NN classifier/regressor #124

### LINKED ISSUE #124: Improve handling of missing optimizer in NN classifier/regressor
When an optimizer is not passed to neural network classifier/regressor an exception occurs: `AttributeError: 'NoneType' object has no attribute 'optimize'`. In this issue I suggest to improve this behavior and either raise a more meaningful exception or introduce a default optimizer, e.g. Cobyla or SLSQP.

darshkaushik:
### Summary
This is to provide the default value of `optimizer` argument of class `TrainableMethod` constructor, which is the super-class of NN classifier/regressor.
This fixes #124.

### Details and comments
- Default value of optimizer set to SLSQP
- Changed the docstring of `__init__()` in class `TrainableMethod`
-


woodsp-ibm:
I think we should have a release note, since this is change in function/new feature (done with reno - see here https://github.com/Qiskit/qiskit-machine-learning/blob/main/CONTRIBUTING.md#adding-a-new-release-note) to say something like for the NN Classifier and regressor that the optimizer is now Optional and will default to SLSQP if not explicitly provided.

I imagine we do not have unit testing that covers the case where an optimizer is not provided. Perhaps we need a simple one (or two) added that checks the classifier and regressor function when no optimizer is provided to ensure everything works, and continues to do so, across any future changes. @adekusar-drl Does that seem reasonable?

adekusar-drl:
Agree with @woodsp-ibm. Please add:

- a reno file with a short description of the proposed change
- two simple tests for classifier and regressor when no optimizer is provided.

But @woodsp-ibm do we have to update docstrings in both of them, saying what the default value is? I'm also thinking of `VQC/VQR` they have optimizer arguments. Should we update them as well?

woodsp-ibm:
> do we have to update docstrings in both of them, saying what the default value is? I'm also thinking of VQC/VQR they have optimizer arguments. Should we update them as well?

Yes, I think the docstrings should be updated as you say, for VQC/VQR plus the typehints too to Optional now that I look. Also I guess NueralNetworkClassified too (the regressor equivalent does not override the init it seems so there is nothing to do there)

darshkaushik:
Thanks for the suggestions. I changed the unit tests, docstrings and type-hints of `VQR` and `VQC` in this PR itself. If you feel that the changes for `VQR` and `VQC` should be addressed in a separate PR or issue then let me know.

test/algorithms/classifiers/test_neural_network_classifier.py
```diff
@@ -72,17 +76,22 @@ def test_classifier_with_opflow_qnn(self, config):
...
 
-        classifier = NeuralNetworkClassifier(
-            qnn, optimizer=optimizer, loss=loss, initial_point=initial_point
-        )
+        if optimizer is None:
+            classifier = NeuralNetworkClassifier(qnn, loss=loss, initial_point=initial_point)
+        else:
+            classifier = NeuralNetworkClassifier(
+                qnn, optimizer=optimizer, loss=loss, initial_point=initial_point
+            )
```
adekusar-drl, line 94: I guess there's no need to have this if statement. Just passing all parameters including `optimizer` which is set to `None` is enough.

test/algorithms/classifiers/test_neural_network_classifier.py
```diff
@@ -150,9 +165,12 @@ def parity(x):
...
         # construct classifier
-        classifier = NeuralNetworkClassifier(
-            qnn, optimizer=optimizer, loss=loss, initial_point=initial_point
-        )
+        if optimizer is None:
+            classifier = NeuralNetworkClassifier(qnn, loss=loss, initial_point=initial_point)
+        else:
+            classifier = NeuralNetworkClassifier(
+                qnn, optimizer=optimizer, loss=loss, initial_point=initial_point
+            )
```
adekusar-drl, line 173: Same as above, I don't see a need for the `if` statement

test/algorithms/classifiers/test_neural_network_classifier.py
```diff
@@ -218,9 +240,14 @@ def parity(x):
...
-            qnn, optimizer=optimizer, loss=loss, one_hot=True, initial_point=initial_point
-        )
+        if optimizer is None:
+            classifier = NeuralNetworkClassifier(
+                qnn, loss=loss, one_hot=True, initial_point=initial_point
+            )
+        else:
+            classifier = NeuralNetworkClassifier(
+                qnn, optimizer=optimizer, loss=loss, one_hot=True, initial_point=initial_point
+            )
```
adekusar-drl, line 250: I don't comment all such places, I think you get the idea. Thanks.

test/algorithms/classifiers/test_neural_network_classifier.py
darshkaushik: Thought to make the need of these test explicit, but I definitely agree it does not make sense to keep them.

test/algorithms/classifiers/test_neural_network_classifier.py
darshkaushik: Agreed

darshkaushik:
@adekusar-drl let me know if there is a scope of any more redundancies or refactoring. 

Else if you think that the changes look fine and once this PR gets merged, please redirect me to others issues/features that I should be able to contribute to.

adekusar-drl (dismissed):
Looks good to me.
@woodsp-ibm do you have any questions?

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -66,7 +66,7 @@ def __init__(
                 as a set of individual predictions with occurrence probabilities (the index would be
                 the prediction and the value the corresponding frequency, e.g. for absolute/squared
                 loss). This option is ignored in case of a one-dimensional output.
-            optimizer: An instance of an optimizer to be used in training.
+            optimizer: An instance of an optimizer to be used in training. Default is SLSQP.
```
woodsp-ibm, line 69: Can we make the default behavior a little clearer for a reader/user by saying something like 
```
When `None` defaults to SLSQP.
```
instead of just `Default is SLSQP`.

Same for the other classes

darshkaushik: Should the reno be changed too?

woodsp-ibm: I think the reno is fine - its says we now have SLSQP as a default. Here, where its closer to the code, where the user needs to be a bit more aware of how the default comes about, i,e, via None, I think the docstrings suffice so they can find how how if/when they need to.

woodsp-ibm (approved):
Thanks for doing this! LGTM

adekusar-drl:
@darshkaushik you can take a look at #112. This feature request requires more efforts and more interesting for sure.

PR #133, opened 2021-06-25
Move away from using pylint header plugin

mtreinish:
### Summary

Currently we're relying on a pylint plugin to verify copyright headers
conform to the project guidelines. However this is problematic for a
number of reasons. The first is it has proven to be unstable there have
been several instances of both users and CI failing on header files that
do not have any issues. This is pretty common with pylint in general
since it's very sensitive to differences in environment. Additionally,
it was using a regex to try and match the entire header string. This
provided no practical debugability when it fails (especially when the
failure is unexpected) because we end up with only a match error.
This commit addresses these issues by removing the use of the regex
based pylint plugin and copying and adapting the small script qiskit-terra
uses to verify the header in every qiskit file and provide useful debugging
if there is an issue with a header in a python file.

### Details and comments

### DIFF
--- a/.pylintdict
+++ b/.pylintdict
@@ -108,7 +108,9 @@ precomputes
 precomputing
 priori
 ps
+py
 pytorch
+pxd
 qae
 qarg
 qargs
--- a/.pylintrc
+++ b/.pylintrc
@@ -23,9 +23,6 @@ persistent=yes
 # usually to register additional checkers.
 load-plugins=pylint.extensions.docparams, # enable checking of docstring args
     pylint.extensions.docstyle,  # basic docstring style checks
-    pylintfileheader  # Check license comments
-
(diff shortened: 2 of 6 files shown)


manoelmarques:
In qiskit-nature we don't check copyrights for anything under folder `gauopen` since it is Gaussian code with its own copyright. Pylint allows it with the argument `--ignore=gauopen`.
If this tool is used in qiskit-nature as it is, it will cause copyright failures that we can't fix.

mtreinish:
> In qiskit-nature we don't check copyrights for anything under folder `gauopen` since it is Gaussian code with its own copyright. Pylint allows it with the argument `--ignore=gauopen`.
> If this tool is used in qiskit-nature as it is, it will cause copyright failures that we can't fix.

It's trivial to modify the script to exclude a directory. We can either just add a condition to `discover_files()` for the version in nature. I don't think it's worth adding an option to everything since that's a pretty rare case. Especially in other places where we vendor things it's typically to fork it for Qiskit use and we just add the original license header below the qiskit apache header.

manoelmarques (approved):

manoelmarques:
@mtreinish It only validates against qiskit_machine_learning, missing the tests and tools folder. I will change to include thoses. Does it work in qiskit-terra ?

mtreinish:
> @mtreinish It only validates against qiskit_machine_learning, missing the tests and tools folder. I will change to include those. Does it work in qiskit-terra ?
>

Ah right I forgot about those. The script takes an arbitrary number of paths (it just defaults to only `qiskit_machine_learning`). So you can pass multiple paths to the script as positional arguments, like:

`./tools/verify_headers.py qiskit_machine_learning tests tools`

manoelmarques:
@mtreinish Thanks, I will put back the old default and change the CI then to include the folders as arguments.

manoelmarques (approved):

PR #135, opened 2021-06-25
Fixups to API ref documentation

woodsp-ibm:
Several fixups to docs, including:

* Fix NeuralNetwork so its docstring formats correctly
* Remove the duplicate of it from the index
* Include connectors to the index and hence TorchConnector into the docs which otherwise was missing
* Several other minor text changes/fixes

### DIFF
--- a/qiskit_machine_learning/neural_networks/__init__.py
+++ b/qiskit_machine_learning/neural_networks/__init__.py
@@ -21,6 +21,9 @@ to a discriminative or generative task.
 Neural Networks may be used, for example, with the
 :class:`~qiskit_machine_learning.algorithms.QGAN` algorithm.
 
+See also the :class:`~qiskit_machine_learning.connectors.TorchConnector` that allows the
+use of these neural networks in code written to `PyTorch <https://pytorch.org/>`_.
+
 .. currentmodule:: qiskit_machine_learning.neural_networks
 
 Neural Network Base Classes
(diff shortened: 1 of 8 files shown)


qiskit_machine_learning/neural_networks/__init__.py
```diff
@@ -39,10 +43,8 @@
    :toctree: ../stubs/
    :nosignatures:
 
-   NeuralNetwork
    OpflowQNN
    TwoLayerQNN
-   SamplingNeuralNetwork
```
adekusar-drl, line 45: Do you think we don't need base classes in the documentation? I don't a strong opinion, just curious.

woodsp-ibm: If you look at the `Neural Network Base Classes` section just above `NeuralNetwork` was there too. So I just removed the duplicate from `Neural Networks` section since having it twice seemed like a mistake. Since `SamplingNeuralNetwork` is and abstract class too I moved it up in the base classes section so the concrete instances were separate.

In answer to your specific question they should be there. We might expect people to extend them themselves, or want to refer to methods etc of the base class. Anything we consider public ought to be in the API ref,

adekusar-drl: Got it. Thanks!

adekusar-drl (approved):

PR #139, opened 2021-06-29
Change TorchConnector NB heading levels

woodsp-ibm:
Tutorial notebooks should have just one heading 1 as the main title - if not subsections appear in the index looking to be the same level as the otherwise different tutorials. Adjusted the headings down one level for Part 1 and Part 2, and their nested sections, such that this is the case.

The notebook seems to have had a bunch of random `id`s in the cells. At least they look random to me. In editing this in my `jupyter notebook` environment it seems to have removed them - I don't think they have importance but I noted it here in case.


woodsp-ibm:
I added OnHold label while looking at the outcome when we set input_gradients=True to the predictions

woodsp-ibm:
I am leaving this on hold. A separate PR is underway with the notebook and other code changes. Once that is readied here then I intend to close this off.

woodsp-ibm:
Closing in favor of #146 - also the CI build was patched in another PR to get it working again as was done here, so this is not needed at all.

PR #143, opened 2021-07-02
Added ability to automatically handle categorical labels (#62)

### LINKED ISSUE #62: Make `NeuralNetworkClassifier` to accept categorical input
Currently `NeuralNetworkClassifier` and `VQC` accepts only one hot encoding. Both classifiers should also accept categorical input.

mattwright99:
### Summary

`NeuralNetworkClassifier` now automatically checks for and encodes categorical target data. 

Closes #62

### Details and comments

Added method to `NeuralNetworkClassifier` that checks target data for the presence of categorical data in the `fit` and `score` methods. Uses sklearn's `OneHotEncoder` and `LabelEncoder` to transform data.


qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -119,4 +124,34 @@ def predict(self, X: np.ndarray) -> np.ndarray:  # pylint: disable=invalid-name
...
+
+    def _convert_categorical(self, y: np.ndarray) -> np.ndarray:
+        """Converts categorical target data using label or one-hot encoding."""
+        y = np.array(y)
+
+        # Test if data is numeric
+        try:
+            y.astype(float)
+        except ValueError:
+            # nonnumeric data is assumed to be cartegorical
```
woodsp-ibm, line 138:
```suggestion
            # non-numeric data is assumed to be categorical
```

releasenotes/notes/cateforical-output-049e682adc9d1e28.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Classificaiton models (i.e. models that extend the ``NeuralNetworkClassifier`` class like VQC)
```
woodsp-ibm, line 4:
```suggestion
    Classification models (i.e. models that extend the ``NeuralNetworkClassifier`` class like VQC)
```

releasenotes/notes/cateforical-output-049e682adc9d1e28.yaml
```diff
@@ -0,0 +1,8 @@
...
+    can now handle categorical target data in methods like ``fit()`` and ``score()``. Categorical
+    data is inferred from the presence of nonnumerical data and is automatically encoded using
```
woodsp-ibm, line 6:
```suggestion
    data is inferred from the presence of non-numerical data and is automatically encoded using
```

woodsp-ibm:
@mattwright99 I fixed a couple of spelling errors - we run spell check since the docstrings become the public API ref etc. So hopefully no lint/spell job failures this time around. Also there were some 500 Internal Errors connecting to servers which hopefully was transient. Anyway fixing the spelling kicked things off again so we can see.

releasenotes/notes/cateforical-output-049e682adc9d1e28.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Classification models (i.e. models that extend the ``NeuralNetworkClassifier`` class like VQC)
+    can now handle categorical target data in methods like ``fit()`` and ``score()``. Categorical
+    data is inferred from the presence of non-numerical data and is automatically encoded using
+    either one-hot or integer engodings. Encoder type is determined by the ``one_hot`` argument
```
woodsp-ibm, line 7:
```suggestion
    either one-hot or integer encodings. Encoder type is determined by the ``one_hot`` argument
```

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -119,4 +124,34 @@ def predict(self, X: np.ndarray) -> np.ndarray:  # pylint: disable=invalid-name
...
+        if is_categorical and self._one_hot:
+            if not self._target_encoder:
+                self._target_encoder = OneHotEncoder()
+                self._target_encoder.fit(y)
+            y = self._target_encoder.transform(y)
+        elif is_categorical and not self._one_hot:
+            if not self._target_encoder:
+                self._target_encoder = LabelEncoder()
+                self._target_encoder.fit(y)
+            y = self._target_encoder.transform(y)
```
woodsp-ibm, line 153: I think this should be more like `if self._target_encoder is None:`. However I wonder so you need to store this in an instance var? I only see it being used here. How about

```
        if is_categorical:
            target_encoder = OneHotEncoder() if self._one_hot else LabelEncoder()
            target_encoder.fit(y)
            y = target_encoder.transform(y)
```        
which keeps target encoder local, and condenses the code since its really the same in each branch except for the encoder.    (I was looking at this since mypy failed in this area)

woodsp-ibm:
FYI: CI here is failing due to an updated torchvision having an issue which fails our TorchConnector notebook see #139 where I was updating things and hit this. We will have to sort this CI failure but that failure has just started happening with the update and is unrelated to any changes you made here

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
woodsp-ibm: FYI. CI is now running again. There is some random failure that sometimes happen, that we should take care of due to precision, but is unrelated to your contribution. Other jobs are failing mypy as I mentioned related to the above code
```
qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py:147: error: "None" has no attribute "fit"
qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py:148: error: "None" has no attribute "transform"
qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py:152: error: "None" has no attribute "fit"
qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py:153: error: "None" has no attribute "transform"
```

mattwright99: @woodsp-ibm the only reason I store the encoder as an instance var is becasue of the potential edge case where a user calls `score()` on categorical test data that is missing data for a certain class. For example, if they train on data that has the classes "A", "B", and "C" but then call score and provide it a subset of data that happens to only have "A" and "C" - this would result in an inconsistent encoding. I don't think this would be very common becuase the test data should be balanced but it could still happen. 
Can you think of a better way to enforce the output of the encoding? Maybe I could try using the `output_shape` attribute of the `NeuralNetwork` or requiring that the user provide a list of classes to fit the encoder to? Should I even be concerned with this edgecase?

I agree with your other comments though and will make some changes. 
Thanks for taking care of the spelling mistakes btw!

woodsp-ibm: So the encoder is stateful I guess which is why you need to re-use it for score. Since the type really only depends on the one_hot setting would it be simpler then to create it in the constructor, rather than doing it lazily later. That would save the test of it being there or not and the code would be simply something like
```
        if is_categorical:
            self._target_encoder.fit(y)
            y = self._target_encoder.transform(y)
``` 
I am not so familiar with this area; if I then train later with different data say with more classes can I reuse the same encoder - in your logic once its created it will stay there and be re-used. On that basis is why I was suggesting creating in constructor - just checking (or do you need to create a new one when fitting each time, which I don;t see the code doing)

mattwright99: Ya the reason why I don't instantiate the proper encoder in the constructor and use `None` instead is because I wanted ensure that we call `encoder.fit()` only once to ensure the encoding stays consistent between all calls to the classifier's `score` and `fit` method. The other way I can think to ensure it is fit only once would be to call `encoder.transform()` in a try-catch block.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -75,8 +77,11 @@ def __init__(
         """
         super().__init__(neural_network, loss, optimizer, warm_start, initial_point)
         self._one_hot = one_hot
+        self._target_encoder = None  # encodes the target data if categorical
```
woodsp-ibm, line 80:
```suggestion
        self._target_encoder: Union[OneHotEncoder, LabelEncoder] = None  # encodes target data if categorical
```
Lets see if this typehint helps mypy; otherwise I think its not picking up on the type and giving errors about None not having fit or transform. What might be better than this though, if it works, would be to say its a `TransformerMixin` type and instead of doing `fit` then `transform` in the code below use the `fit_transform` defined by this, which seems to be a convenience method for calling the two in sequence. The encoders implement that Mixin so it seemed to be the better solution.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -75,8 +77,12 @@ def __init__(
         """
         super().__init__(neural_network, loss, optimizer, warm_start, initial_point)
         self._one_hot = one_hot
+        # encodes the target data if categorical
+        self._target_encoder: Optional[Union[OneHotEncoder, LabelEncoder]]: = None
```
woodsp-ibm, line 81:
```suggestion
        self._target_encoder: Optional[Union[OneHotEncoder, LabelEncoder]] = None 
```

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -75,8 +77,12 @@ def __init__(
         """
         super().__init__(neural_network, loss, optimizer, warm_start, initial_point)
         self._one_hot = one_hot
+        # encodes the target data if categorical
+        self._target_encoder: Optional[Union[OneHotEncoder, LabelEncoder]] = None
```
woodsp-ibm, line 81:
```suggestion
        # encodes the target data if categorical
        self._target_encoder: Optional[Union[OneHotEncoder, LabelEncoder]] = None
```
Removed trailing white space - third time lucky!

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
woodsp-ibm: I committed this, so it got marked resolved, but unresolved it so the comment is readily visible about TransformerMixin. The typehint change (after I fixed it up!) did sort things out though for mypy.

woodsp-ibm: My bad this can't be done. As fit is always called but transform is conditional!

mattwright99:
@woodsp-ibm It worked! Good idea defining the encoder type.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -119,4 +125,25 @@ def predict(self, X: np.ndarray) -> np.ndarray:  # pylint: disable=invalid-name
     def score(
         self, X: np.ndarray, y: np.ndarray, sample_weight: Optional[np.ndarray] = None
     ) -> float:
+        y = self._convert_categorical(y)
```
adekusar-drl, line 128: I guess it is not a good idea to call the same method in `score` as in `fit`:

- In `model.fit()` there should be a call `encoder.fit_transform()` (or `fit` and then `transform`)
- While in `score` only `encoder.transform()` must be applied.

Mixing up these two things in one method may lead to a less readable code.
Surely, `encoder.transform()` must be called in `predict` as well.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -119,4 +125,25 @@ def predict(self, X: np.ndarray) -> np.ndarray:  # pylint: disable=invalid-name
...
+        except ValueError:
+            # non-numeric data is assumed to be categorical
+            y = y.reshape(-1, 1)
+            if self._target_encoder is None:
+                self._target_encoder = OneHotEncoder() if self._one_hot else LabelEncoder()
+                self._target_encoder.fit(y)
+            y = self._target_encoder.transform(y)
+
+            if not isinstance(y, np.ndarray):
+                y = np.array(y.todense())
```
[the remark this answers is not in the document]
mattwright99: sometimes encoding returns a sparse matrix which throws an error down the line

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -119,4 +125,25 @@ def predict(self, X: np.ndarray) -> np.ndarray:  # pylint: disable=invalid-name
...
+        # Test if data is numeric
+        try:
+            y.astype(float)
+        except ValueError:
+            # non-numeric data is assumed to be categorical
+            y = y.reshape(-1, 1)
+            if self._target_encoder is None:
+                self._target_encoder = OneHotEncoder() if self._one_hot else LabelEncoder()
+                self._target_encoder.fit(y)
+            y = self._target_encoder.transform(y)
```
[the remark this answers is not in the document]
mattwright99: I kind of understand what you mean but when would someone retrain a model on a completely different structured output? Can you even change the `output_shape` attribute of an instance of `NeuralNetwork`?

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
mattwright99: That would definitely work too. That way we could create the `_target_encoder` attribute in the constructor like your first comment suggests and we would re-fit the encoder everytime we call `model.fit()` as you also noted. The only chance for error would be if the user tries to score their model before training it but I don't see that being very big deal.

There should be no calls to the encoder in `predict` becuase there is no instance of target data.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -119,4 +125,25 @@ def predict(self, X: np.ndarray) -> np.ndarray:  # pylint: disable=invalid-name
...
+        y = self._convert_categorical(y)
         return ClassifierMixin.score(self, X, y, sample_weight)
+
+    def _convert_categorical(self, y: np.ndarray) -> np.ndarray:
+        """Converts categorical target data using label or one-hot encoding."""
+        y = np.array(y)
+
+        # Test if data is numeric
+        try:
+            y.astype(float)
```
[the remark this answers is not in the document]
mattwright99: Good point, the reason why I did it this way was in case of any edge cases where, for example,  the user may have read data from a file and it stayed in string format. I really just wasn't sure what the best way was to handle all the different numpy data types so I figured seeing if it could be converted to a number was a catch-all way to check for categorical data but I am happy going your way, it is definitely more clear.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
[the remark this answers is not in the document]
adekusar-drl: Well, without details I may say it could be done intentionally. We do support sparse arrays.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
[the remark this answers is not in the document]
adekusar-drl: No, `output_shape` stays the same. It comes from an instance of NN and it is passed in the constructor and can't be updated. So, a classifer can be re-used for a dataset with the same dimensions. Rarely the case, but better to be on a safe side.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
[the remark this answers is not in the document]
mattwright99: Ok then we can totally get rid of it, I wasn't aware of that. I added it mostly becuase of the code [here](https://github.com/Qiskit/qiskit-machine-learning/blob/95a91fcffd2447645f78a004e0ea4e3b457d76c1/qiskit_machine_learning/datasets/dataset_helper.py#L180-L185).

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
adekusar-drl: Shame on me, you're right, there are no labels in `predict`. Unless we add support of categorical features.
If a user calls `score` or `predict` before `fit`, there's a check `if self._fit_result is None` that raises an exception.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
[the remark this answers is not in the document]
mattwright99: Ok so should I keep it the way it is and only fit the encoder once or call `encoder.fit` each time we train the model. Both would have the same effect most often so I am good with either.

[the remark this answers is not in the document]
adekusar-drl: Please call `encoder.fit` every time we train the model.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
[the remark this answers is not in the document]
mattwright99: Actually I got rid of this check and it throws an error when calculating the loss for one-hot encoded results that I forgot about. It gives the message:
```
File "C:\path\to\repo\qiskit-machine-learning\qiskit_machine_learning\algorithms\objective_functions.py", line 203, in gradient
    grad += loss_gradient[:, i] @ weight_prob_grad[:, i, :]
ValueError: matmul: Input operand 1 has a mismatch in its core dimension 0, with gufunc signature (n?,k),(k,m?)->(n?,m?) (size 5 is different from 1)
```
Should I keep these two lines then?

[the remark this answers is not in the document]
adekusar-drl: That's weird, is it somewhere in the tests? I'd check if all parameters are set correctly.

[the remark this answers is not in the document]
mattwright99: It is failing [this test](https://github.com/Qiskit/qiskit-machine-learning/pull/143/files#diff-b0eb994eebb823e4a87b21ccf3edd69499cf91a0adc3e9dcb2582afcdf3e112aR264) when one-hot is true and I remove those lines.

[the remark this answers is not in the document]
adekusar-drl: I see, set `OneHotEncoder(sparse=False)`, the encoder by default returns a sparse matrix and that causes the error. Well, even with `sparse=True` the code should work. So, if you have time you may investigate why it fails. This is sort of a bug.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -75,8 +77,12 @@ def __init__(
         """
         super().__init__(neural_network, loss, optimizer, warm_start, initial_point)
         self._one_hot = one_hot
+        # encodes the target data if categorical
+        self._target_encoder = OneHotEncoder(sparse=False) if one_hot else LabelEncoder()
 
     def fit(self, X: np.ndarray, y: np.ndarray):  # pylint: disable=invalid-name
+        y = self._fit_and_encode_categorical(y)
```
adekusar-drl, line 84: I'd rename to `_fit_and_encode_labels` or `_fit_transform_lables` with a stress on labels.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -119,4 +125,26 @@ def predict(self, X: np.ndarray) -> np.ndarray:  # pylint: disable=invalid-name
     def score(
         self, X: np.ndarray, y: np.ndarray, sample_weight: Optional[np.ndarray] = None
     ) -> float:
+        y = self._encode_categorical(y)
```
adekusar-drl, line 128: Same here, `_encode_labels` or `_transform_labels`.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -119,4 +125,26 @@ def predict(self, X: np.ndarray) -> np.ndarray:  # pylint: disable=invalid-name
...
+            y = y.reshape(-1, 1)
+            self._target_encoder.fit(y)
+            y = self._target_encoder.transform(y)
+
+        return y
+
+    def _encode_categorical(self, y: np.ndarray) -> np.ndarray:
+        """Converts categorical target data using label or one-hot encoding."""
+
+        if isinstance(y[0], str):
```
adekusar-drl, line 145: This `if` statement really annoys me, any ideas how we can avoid it?

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -119,4 +125,26 @@ def predict(self, X: np.ndarray) -> np.ndarray:  # pylint: disable=invalid-name
...
     ) -> float:
+        y = self._encode_categorical(y)
         return ClassifierMixin.score(self, X, y, sample_weight)
+
+    def _fit_and_encode_categorical(self, y: np.ndarray) -> np.ndarray:
+        """Fits label or one-hot encoder and converts categorical target data."""
+
+        if isinstance(y[0], str):
+            # string data is assumed to be categorical
+            y = y.reshape(-1, 1)
```
adekusar-drl, line 136: Are we sure that the second dimension in the reshape is `1`?

adekusar-drl (commented):
Sorry for the delay, looks good! Just a few minor questions left.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
mattwright99: I can't really see away around it - we need to check for categorical data somehow and looking for data of type `str` seems to be the cleanest way. I am always open to suggestions though.

What part bugs you?

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
mattwright99: I suppose the implicit assumption in these functions is that if the provided labels are multidimensional then any encoding has already been done. This is a result of checking if `y[0]` is a `string` which, if satisfied, suggests a 1D list of data is provided and can therefore be transformed to a column. I did actaully find a small bug thanks to this comment and will push a change to address it (reshaping to a column throws a warning when using the `LabelEncoder`).

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
adekusar-drl: Well, if we know how labels were fitted and transformed in training then likely we know what to do here. That's why I'm thinking of this.

adekusar-drl: I also don't have a neat solution, so let's keep it as it is now.
P.S. seems like the CI is stuck somewhere...

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -119,4 +125,26 @@ def predict(self, X: np.ndarray) -> np.ndarray:  # pylint: disable=invalid-name
...
     ) -> float:
+        y = self._encode_labels(y)
         return ClassifierMixin.score(self, X, y, sample_weight)
+
+    def _fit_and_encode_labels(self, y: np.ndarray) -> np.ndarray:
+        """Fits label or one-hot encoder and converts categorical target data."""
+
+        if isinstance(y[0], str):
+            # string data is assumed to be categorical
+            y = y.reshape(-1, 1) if self._one_hot else y  # reshape labels for encoder
```
adekusar-drl, line 136: So now I'm confused. Why you added `if self._one_hot else y`?

mattwright99: That was that bug I breifly mentioned earlier. `OneHotEncoder` expects input of shape `(n_samples, n_features)` (see [here](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html#sklearn.preprocessing.OneHotEncoder.fit)) while `LabelEncoder` takes in data with shape `(n_samples,)` (see [here](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.LabelEncoder.html#sklearn.preprocessing.LabelEncoder.fit)) so it was just printing a warning whenever someone uses the `LabelEncoder`.

adekusar-drl: Thanks a lot, would you mind to add a comment with this explanation to the code? Just one or two lines should be enough, let's make life easier for future readers. Other than that I don't have anything on my list.

mattwright99: Sounds good!  Added in [3f7c632](https://github.com/Qiskit/qiskit-machine-learning/pull/143/commits/3f7c632a6398b004ab3b42042a21a1011ddc9c83)

adekusar-drl (dismissed):
Thanks a lot for taking care, really appreciated!

woodsp-ibm (approved):

adekusar-drl (approved):
Just merged the main branch.

PR #150, opened 2021-07-08
Update default batch_size in quantum_kernel

attp:
The max circuits per job for most/all backend devices is 900. Setting the default `batch_size=900` instead of 1000 reduces the number of jobs being created.


woodsp-ibm:
Code must be formatted nowadays using black which looks like what is giving rise to the conflict. By running `make black` it formats the code to a given style and avoids the style checks/errors/manual correction cycle that use to happen in the past.

Just as an observation, going forwards, where grouping/batching is more for execution efficiency rather than some more fundamental algorithmic reason we should find a way to have this perhaps more automatically configured via the QuantumInstance. Lists of circuits already get automatically split if they exceed the backend limit but that does not lead to optimal job sizes/efficiency. I added a note to the issue in Terra around improving QI/circuit execution Qiskit/qiskit-terra#5717

adekusar-drl:
@woodsp-ibm So, are we fine with the proposed change or should introduce such things in Terra in `QuantumInstance` or somewhere else?

woodsp-ibm:
> So, are we fine with the proposed change

I think this is fine for now - I added a note to the Issue that covers execution improvements etc in Terra. We can potentially revisit this in the future when that issue gets looked at.

woodsp-ibm (approved):

PR #156, opened 2021-07-09
Add hybrid qnn unit tests for TorchConnector

ElePT:
### Summary
After noticing that there were no unit tests for hybrid qnns and `TorchConnector` (during PR #146), I have added 2 new test cases (one for CircuitQNN, one for OpflowQNN) to check for gradient propagation and to make sure that batch gradients are computed properly. 

### Details and comments
Accessing the qnn's weight using the torch `named_parameters()` method is not very comfortable (for example, in line 722), maybe we should reconsider changing the name from `self._weights` to `self.weight` for the sake of consistency.


adekusar-drl:
Unfortunately `weights` is consistent with the rest of the QML code.

woodsp-ibm:
Does the test represent how any end user would do things. Having to rely on a private variable, `_weights`, whatever the name, is not good,

ElePT:
@woodsp-ibm I would say that it does. It is common to iterate over the network parameters when you are checking gradients, and in PyTorch you normally use the `named_parameters()` method to see which layer they correspond to and whether they are weights or biases (which is what I check on line 722).

test/connectors/test_torch_connector.py
```diff
@@ -659,6 +659,205 @@ def test_circuit_qnn_batch_gradients(self, config):
...
+            output = model(x_i)
+            sum_of_individual_losses += f_loss(output, y_i)
+        sum_of_individual_losses.backward()
+        sum_of_individual_gradients = 0.0
+        for n, param in model.named_parameters():
+            name = n.split(".")
+            # make sure gradient is not None
+            self.assertFalse(param.grad is None)
+            if name[1] == "weight" or name[1] == "_weights":
+                sum_of_individual_gradients += np.sum(param.grad.numpy())
```
adekusar-drl, line 728: Can you please explain what's going on here and why we need this `if` statement? Please ping me to discuss this part when you have time.

adekusar-drl (commented):
I'm in favor of having these tests. Could it be better to move them to a separate script? Current script is already too large.

test/connectors/test_torch_connector.py
woodsp-ibm: My comment

>Having to rely on a private variable, _weights, whatever the name, is not good,

Was in direct relation to this code seeing the name being used here. Hence the question was this what we would normally expect an end user to do.

> Anyway, the parameter values are managed automatically by Torch's differentiation engine (it just detects the TorchParameter class and takes care of everything parameter-related), so can they be considered private?

All variables are accessible in Python whether they are considered private or not. The main difference is that we only consider the public API when doing changes and deprecate if/when compatibility breaks. As part of the private interface we freely change that, since its supposed to be part of the private implementation. Hence my concern if that access was considered how a user would normally do things I think we need to look at that!

adekusar-drl: We can avoid having both `_weights` and `weight`. I propose to introduce module parameters explicitly:
```python
        weight = TorchParam(Tensor(neural_network.num_weights))
        self.register_parameter("weight", weight)
        self._weights = weight    # this is for compatibility only, can be revised.
```
This creates a parameter on a module level that can be accessed in the same fashion as we access traditional layers. Thus, we can simplify the `if` statement and remove our property `weight`. In the same time we can keep `self._weights` for compatibility.

Or we can simply rename `_weights` in favor of `weight` to be compliant with pytorch but incompliant with the rest code. @stefan-woerner any opinion?

test/connectors/test_torch_connector.py
```diff
@@ -659,6 +659,205 @@ def test_circuit_qnn_batch_gradients(self, config):
...
+            batch_loss.detach().numpy(),
+            places=4,
+        )
+
+    @data(
+        # output_shape, interpret
+        (2, lambda x: "{:b}".format(x).count("1") % 2),
+    )
+    @requires_extra_library
+    def test_circuit_qnn_hybrid_batch_gradients(self, config):
```
adekusar-drl, line 759: Since there's just one tuple in `@data`, parameters can be safely moved in the test itself.

test/connectors/test_torch_connector.py
```diff
@@ -659,6 +659,205 @@ def test_circuit_qnn_batch_gradients(self, config):
...
+        for x_i, y_i in zip(x, y):
+            output = model(x_i)
+            sum_of_individual_losses += f_loss(output, y_i)
+        sum_of_individual_losses.backward()
+        sum_of_individual_gradients = 0.0
+        for n, param in model.named_parameters():
+            name = n.split(".")
+            # make sure gradient is not None
+            self.assertFalse(param.grad is None)
+            if name[1] == "weight" or name[1] == "_weights":
```
adekusar-drl, line 727: I think we can replace `name = n.split(".")`, `name[1] == "weight"` with `n.endwith(".weight")` once we have identical names for weights.

test/connectors/test_torch_connector.py
ElePT: @adekusar-drl True! Registering the parameter with a different name is a good idea. I don't know why I didn't think of that before... I will wait for consensus and make the corresponding changes.

test/connectors/test_torch_connector.py
ElePT: True, I will remove that line.

test/connectors/test_torch_connector.py
ElePT: Yes, once the names are fixed, this line can be simplified.

test/connectors/test_torch_connector.py
adekusar-drl: @ElePT Can you verify that registering parameter explicitly does not break anything? If this is the one things we should do to have compatible parameter naming, then, I think, we can proceed with this option.

ElePT: @adekusar-drl It works as expected, nothing broken as far as I can tell, so I have already pushed the changes. As this affects the way the user interacts with the TorchConnector, should I also add a release note?

adekusar-drl: I'm not sure it affect the way a user interacts with TorchConnector. There's the same property, I don't see how the changes may affect users, so in this case it is not required. Well, if I don't miss anything.

ElePT: I mean... it changes the output of  the torch function `model.named_parameters()`. Before , it would output `{'qnn._weights', 2.5}`, and now it will output `{'qnn.weight', 2.5}` (just like in the unit test). If somebody wants to check the gradients, they will see that difference in the name (that's why we changed it). It's true that it's not a direct interaction with the user, but maybe an interaction through PyTorch? I am asking because I it's not that clear to me either.

adekusar-drl: Now I see your point. Yes, better to have this change documented in a reno. Thanks for noting this.

qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -226,7 +226,12 @@ def __init__(
         self._neural_network = neural_network
 
         self._sparse = sparse
-        self._weights = TorchParam(Tensor(neural_network.num_weights))
+
+        weight = TorchParam(Tensor(neural_network.num_weights))
+        # register param. in graph following PyTorch naming convention
+        self.register_parameter("weight", weight)
```
woodsp-ibm, line 232: What does this actually register? The variable right above named weight is only in local scope of this method - does this end up registering the property below named weight? Just wondering since self._weights, which was how this was, is what is returned from weight property below - and this was explicitly changed to weight and then assigned to self._weights with this added register_parameter.

adekusar-drl: By registering a parameter under a custom name in a layer (let's call torch connector as a layer in NNs) you expose this parameter as `layer.name` (if I recall correctly), this is quite common in PyTorch models, and add this parameter to a tree of layers/parameters inside PyTorch.
So, when you query model parameters via `named_parameters()` you get them in a list of tuples `(layer_name.param_name, param)`. Usually you see `*.weight` for weights in PyTorch. So here, we'd like to be compatible with PyTorch. Just having an explicit property `weight` is not enough for PyTorch. Either you don't bother registering and in this case the name is your variable name, or you register under a more meaningful name, e.g. to be compliant with built-in layers.

woodsp-ibm: I think my concern was more that the `register_parameter` appears to be registering the local variable here `weight = TorchParam(Tensor(neural_network.num_weights))` - maybe that's fine and the register captures it somehow. Or did it really end up registering the property weight from self, since I guess it does use self.weight I guess its the former. I guess with the variable name `weight` the same as the property `weight` whatever was registered was not so clear. Maybe it has to be the local var since otherwise it would have had to have been self.weight. If its the local param or even an instance var I assume it could have been left as it was and done as `self.register_parameter("weight", self._weights)` - did the code need to be changed to make the name the same! Was less than clear to me

ElePT: PyTorch uses an automatic differentiation package called `autograd` which keeps track of all operations performed on torch `Tensor`s using a DAG (Directed Acyclical Graph).  There are different ways of adding nodes to the graph. In this case, we are dealing with `Module` parameters, so we don't only want to keep track of the operations performed on the tensors, we want to register them as parameters for the corresponding layer of the torch `Module`. 

For that, we normally assign an attribute (in this case, `self._weights`) to a`TorchParameter`, a special kind of torch `Tensor`, that gets automatically registered as a `Module` parameter once assigned to an attribute. This is the easy method, but it automatically registers the parameter with the attribute name (`_weights`). Because we don't want that, we propose the alternative of:

1. Declaring a `TorchParameter`, but not assigning it to any `Module` attribute yet (line 230)
2. Registering the parameter into the graph with the desired name (line 232)
3. Assigning the parameter to the attribute after registering it (it will not be re-registered). This operation is tracked by autograd, so any changes in `self._weights` will affect the node registered as `weight` (line 234)

Hopefully this makes sense and adds to Anton's reply.

ElePT: Oh, and in case you were wondering @woodsp-ibm, I thought about doing `self.register_parameter("weight", self._weights)` too, but this option doesn't work because, as explained before, the moment you do `self._weights=TorchParameter(...)`, the parameter gets registered in the graph, and an error is thrown telling you that you are trying to register it twice.

test/connectors/test_hybrid_connector.py
```diff
@@ -0,0 +1,347 @@
...
+            seed_transpiler=algorithm_globals.random_seed,
+        )
+        try:
+            import torch
+
+            torch.manual_seed(algorithm_globals.random_seed)
+        except ImportError:
+            pass
+
+    def validate_output_shape(self, model: TorchConnector, test_data: List[Tensor]) -> None:
```
adekusar-drl, line 72: I don't see this method is used in this test

test/connectors/test_hybrid_connector.py
```diff
@@ -0,0 +1,347 @@
...
+                q_worked = False
+
+            # compare results and assert that the behavior is equal
+            with self.subTest("c_worked == q_worked", tensor=x):
+                self.assertEqual(c_worked, q_worked)
+            if c_worked and q_worked:
+                with self.subTest("c_shape == q_shape", tensor=x):
+                    self.assertEqual(c_shape, q_shape)
+
+    def validate_backward_pass(self, model: TorchConnector) -> None:
```
adekusar-drl, line 125: same here, it is unused.

adekusar-drl (commented):
Thanks a lot, looks good, just minor comments and, please, add a reno as discussed.

qiskit_machine_learning/connectors/torch_connector.py
woodsp-ibm: Maybe this need to be added as comments then to me I see little logical difference between
```
self._weights = TorchParam(Tensor(neural_network.num_weights))
self.register_parameter("weight", self._weights)
```
and
```
weight = TorchParam(Tensor(neural_network.num_weights))
self.register_parameter("weight", weight)
self._weights = weight
```
except the latter, at least to me, brings in some confusion over name of `weight` when we also have a property called `weight`.  The mere setting of this to some instance varable registers it whereas to a local variable it does not - why does not assigning weight to self._weights in the latter register it again.

Would this would equally as well in regards of things. 
```
weight_tensor = TorchParam(Tensor(neural_network.num_weights))
self.register_parameter("weight", weight_tensor)
```

If its not very intuitive as you say why the code is as it is - for any one that may look at the code in the future to change it for one reason or another I think some explanation around why it is like it is, and that it needs to be that way etc is needed. As you can tell I am struggling with it somewhat as it seems less than logical.

adekusar-drl (approved):
Thanks!

woodsp-ibm (approved):
Thanks for adding the extra text describing the rationale behind the way the register parameter is done; hopefully it will be of help to anyone not familiar with the code in the future to understand why it is like it is there. So seems like its all done!

ElePT:
> Thanks for adding the extra text describing the rationale behind the way the register parameter is done; hopefully it will be of help to anyone not familiar with the code in the future to understand why it is like it is there. So seems like its all done!

Glad to help!

PR #178, opened 2021-08-01
Global random seed

### LINKED ISSUE #158: Random behavior improvements
Algorithms normally have any random behavior done via the algorithm globals random generator as a simple central place to set seed to accommodate reproducibility.

`QSVC` - subclasses sklearn SVC which has random_state param. It might be desirable to pass up the algorithm globals seed to the parent during construction to allow the random state to be set based on the global seed.

And for `VQC` since it use `TrainableModel` this would benefit from using algorithm globals for the random initial point computation https://github.com/Qiskit/qiskit-machine-learning/blob/c1df4d545f74fd166ee973868013a66478c07a20/qiskit_machine_learning/algorithms/trainable_model.py#L197

See also https://quantumcomputing.stackexchange.com/questions/18384/ramdon-accuracy-using-vqc-and-iris-dataset where the current behavior was initially brought up.

MartinBeseda:
### Summary
Global seed `algorithm_globals.random_seed` put in use both for `QSVC` and `VQC` objects. Fixes #158 .

### DIFF
--- a/qiskit_machine_learning/algorithms/classifiers/qsvc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/qsvc.py
@@ -16,7 +16,7 @@ from typing import Optional
 
 from sklearn.svm import SVC
 
-from qiskit.utils import algorithm_globals
+from qiskit.utils.algorithm_globals import algorithm_globals
 
 from qiskit_machine_learning.kernels.quantum_kernel import QuantumKernel
 
@@ -51,7 +51,9 @@ class QSVC(SVC):
 
         super().__init__(
             kernel=self._quantum_kernel.evaluate,
-            random_state=algorithm_globals.random_seed,
+            random_state=kwargs.get("random_state")
+            if "random_state" in kwargs
+            else algorithm_globals.random_seed,
             *args,
             **kwargs,
         )
--- a/qiskit_machine_learning/algorithms/classifiers/qsvc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/qsvc.py
@@ -16,9 +16,10 @@ from typing import Optional
 
 from sklearn.svm import SVC
 
+from qiskit.utils import algorithm_globals
+
 from qiskit_machine_learning.kernels.quantum_kernel import QuantumKernel
 
-from qiskit.utils import algorithm_globals
 
 class QSVC(SVC):
     r"""Quantum Support Vector Classifier.
@@ -48,9 +49,12 @@ class QSVC(SVC):
 
         self._quantum_kernel = quantum_kernel if quantum_kernel else QuantumKernel()
 
-        super().__init__(kernel=self._quantum_kernel.evaluate,
-                         random_state=algorithm_globals.random_seed,
-                         *args, **kwargs)
+        super().__init__(
+            kernel=self._quantum_kernel.evaluate,
+            random_state=algorithm_globals.random_seed,
+            *args,
+            **kwargs,
+        )
 
     @property
     def quantum_kernel(self) -> QuantumKernel:
--- a/qiskit_machine_learning/algorithms/classifiers/qsvc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/qsvc.py
@@ -18,6 +18,7 @@ from sklearn.svm import SVC
 
 from qiskit_machine_learning.kernels.quantum_kernel import QuantumKernel
 
+from qiskit.utils import algorithm_globals
 
 class QSVC(SVC):
     r"""Quantum Support Vector Classifier.
@@ -47,7 +48,9 @@ class QSVC(SVC):
 
         self._quantum_kernel = quantum_kernel if quantum_kernel else QuantumKernel()
 
-        super().__init__(kernel=self._quantum_kernel.evaluate, *args, **kwargs)
+        super().__init__(kernel=self._quantum_kernel.evaluate,
+                         random_state=algorithm_globals.random_seed,
+                         *args, **kwargs)
 
     @property
     def quantum_kernel(self) -> QuantumKernel:
--- a/qiskit_machine_learning/algorithms/classifiers/qsvc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/qsvc.py
@@ -49,11 +49,11 @@ class QSVC(SVC):
 
         self._quantum_kernel = quantum_kernel if quantum_kernel else QuantumKernel()
 
+        if "random_state" not in kwargs:
+            kwargs["random_state"] = algorithm_globals.random_seed
+
         super().__init__(
             kernel=self._quantum_kernel.evaluate,
-            random_state=kwargs.get("random_state")
-            if "random_state" in kwargs
-            else algorithm_globals.random_seed,
             *args,
             **kwargs,
         )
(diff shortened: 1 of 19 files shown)


qiskit_machine_learning/algorithms/trainable_model.py
```diff
@@ -202,7 +205,8 @@ def _choose_initial_point(self) -> np.ndarray:
         if self._warm_start and self._fit_result is not None:
             self._initial_point = self._fit_result[0]
         elif self._initial_point is None:
-            self._initial_point = np.random.rand(self._neural_network.num_weights)
+            rand_state = RandomState(MT19937(SeedSequence(algorithm_globals.random_seed)))
+            self._initial_point = rand_state.rand(self._neural_network.num_weights)
```
adekusar-drl, line 209: I don't fully understand what is going on here, but the reason of the problem is not here. The test `test_neural_network_classifier.test_classifier_with_circuit_qnn_and_cross_entropy` fails sometimes despite `initial_point` initialized as `np.array([0.5] * ansatz.num_parameters)`. So, we start optimization from the same point every time, but sometimes we end up having a score of `0.4`.

requirements.txt
```diff
@@ -5,4 +5,4 @@ psutil>=5
 scikit-learn>=0.20.0
 fastdtw
 setuptools>=40.1.0
-
+qiskit~=0.26.0
```
adekusar-drl, line 8: This is not required. There's `terra` dependency on the first line.

adekusar-drl (commented):
Please explain why this fixes the problem.

qiskit_machine_learning/algorithms/trainable_model.py
MartinBeseda: If I understand it correctly, @woodsp-ibm asked for utilization of `algorithm_globals.random_seed` in `VQC`in https://github.com/Qiskit/qiskit-machine-learning/issues/158#issue-942421425 .

Also, I've passed `algorithm_globals.random_seed` in `QSVC` constructor as `random_state` parameter, so that it "reacts" to the globally-set seed.

requirements.txt
MartinBeseda: Thank you! I'll remove that one.

qiskit_machine_learning/algorithms/trainable_model.py
MartinBeseda: @adekusar-drl Maybe it'd be better to make a separate issue for the problem with `test_neural_network_classifier.test_classifier_with_circuit_qnn_and_cross_entropy`? It'd prevent confusion, as this one seems to be specifically about `QSVC` and `VQC`.

adekusar-drl: The problem is not only in `VQC`. For instance, `test_neural_network_regressor.test_regressor_with_opflow_qnn()` fails for the same reason sometimes. `VQC`, in turns, extends `NeuralNetworkClassifier` and if the latter fails, the former fails as well.
If we are talking about algorithm globals as stated in the issue, then `np.random.rand(self._neural_network.num_weights)` is to be replaced with `algorithm_globals.random.random(...)`. Fine with me, we can just fix this, but such a fix won't fix the problem we try to cope.

woodsp-ibm: @adekusar-drl  The issue I had raised #158 was a) because I noticed direct use of np.random rather than the algorithm globals that is used elsewhere that we can seed centrally for reproducibility etc.  And b) for qsvc it was that sklearn svc has a random aspect that can be seeded, it had been noted to me that the results varied despite setting algorithm globals, the simulator seeds etc. And while in our subclass you can certainly still set the sklearn random state via kwargs, one needs to look at sklearn to figure this out, so simply setting this via algo globals random seed in the qsvc I figured would make it easier for users. So I raised the issue just with these points in mind, it was not motivated at all by the random unit test failure, that occurs.

I must admit I had imagine the change of

> self._initial_point = np.random.rand(self._neural_network.num_weights)

to algorithm_globals to be simply

> self._initial_point = algorithm_globals.random.rand(self._neural_network.num_weights)

where instead of using np global random object it used the algorithms global one.

MartinBeseda: @woodsp-ibm Thank you for the explanation! I've rewritten the random number generation according to your suggestion.

adekusar-drl: @woodsp-ibm OK, clear. I was expecting slightly more in this PR :)

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -47,7 +49,12 @@ def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwar
 
         self._quantum_kernel = quantum_kernel if quantum_kernel else QuantumKernel()
 
-        super().__init__(kernel=self._quantum_kernel.evaluate, *args, **kwargs)
+        super().__init__(
+            kernel=self._quantum_kernel.evaluate,
+            random_state=algorithm_globals.random_seed,
+            *args,
+            **kwargs,
+        )
```
adekusar-drl, line 57: Could it be better to set `random_state` to `algorithm_globals.random_seed` only if `random_state` is not passed in `kwargs`? Now, if I want to pass my own `random_state` it is overwritten in the constructor.

MartinBeseda: @adekusar-drl You're right, I've edited the code accordingly.

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -51,7 +51,9 @@ def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwar
 
         super().__init__(
             kernel=self._quantum_kernel.evaluate,
-            random_state=algorithm_globals.random_seed,
+            random_state=kwargs.get("random_state")
+            if "random_state" in kwargs
+            else algorithm_globals.random_seed,
```
woodsp-ibm, line 56: It maybe easier/simpler to say
```
```
ahead of the super init  call as it was and just pass it up in kwargs rather than do so explicitly by name parameter where we may end up with passing it in kwargs as well if it was there already,

MartinBeseda: @woodsp-ibm Thank you for the suggestion - I've edited the code.

qiskit_machine_learning/algorithms/trainable_model.py
```diff
@@ -202,7 +203,7 @@ def _choose_initial_point(self) -> np.ndarray:
         if self._warm_start and self._fit_result is not None:
             self._initial_point = self._fit_result[0]
         elif self._initial_point is None:
-            self._initial_point = np.random.rand(self._neural_network.num_weights)
+            self._initial_point = algorithm_globals.random.rand(self._neural_network.num_weights)
```
woodsp-ibm, line 206:
```suggestion
            self._initial_point = algorithm_globals.random.random(self._neural_network.num_weights)
```
The deprecated RandomState `rand()` method no longer exists in the Generator replacement. Replacing this by `random()`.

requirements.txt
```diff
@@ -5,4 +5,3 @@ psutil>=5
 scikit-learn>=0.20.0
 fastdtw
 setuptools>=40.1.0
-
```
adekusar-drl, line 8: Could you please rollback the changes here since and a empty line in the end as it was before?

MartinBeseda: @adekusar-drl OK, I did that ;)

adekusar-drl (approved):
Thanks. Looks good to me!

woodsp-ibm (approved):
Thanks for doing this. LGTM too.

Issue #95, opened 2021-05-11
Gradient compute fail when using RawFeatureVector


saalimon:
### Information

- **Qiskit Machine Learning version**: 0.1.0
- **Python version**: 3.7.10
- **Operating system**: All

### What is the current behavior?
This message occurs when using `RawFeatureVector` with `CircuitStateFn`
`Cannot compute gradient operator! Continuing without gradients!`

### Steps to reproduce the problem

```python
from qiskit import QuantumCircuit
from qiskit.opflow import StateFn, PauliSumOp, AerPauliExpectation, Gradient
from qiskit.utils import QuantumInstance
from qiskit_machine_learning.neural_networks import OpflowQNN
from qiskit_machine_learning.circuit.library import RawFeatureVector

expval = AerPauliExpectation()
gradient = Gradient()
qi_sv = QuantumInstance(Aer.get_backend('statevector_simulator'))

inputs = RawFeatureVector(16)
qc = QuantumCircuit(4)
qc.append(inputs,range(4))
qc_sfn = StateFn(qc)
H1 = StateFn(PauliSumOp.from_list([('Z', 1.0)]))
op = ~H1 @ qc_sfn
qnn4 = OpflowQNN(operator = op, input_params = inputs.parameter, weight_params = [], exp_val = expval, gradient = gradient, quantum_instance = qi_sv)

```

### What is the expected behavior?
Can pass the `Parameter` or `ParameterVector` to `OpflowQNN` or another Opflow from `qiskit_machine_learning.neural_networks`

### Suggested solutions
Check the Gradient condition on `qiskit.opflow`

ElePT:
Hello! I would like to work on this issue.

adekusar-drl:
Closed in #182.

PR #182, opened 2021-08-03
Fix `RawFeatureVector` copy

### LINKED ISSUE #180: Gradient warning with RawFeatureVector and Circuit/Opflow QNNs
### Information

- **Qiskit Machine Learning version**: 0.2.0
- **Python version**: 3.9.6
- **Operating system**: macOS Big Sur 11.4

### What is the current behavior?
In both `CircuitQNN` and `OpflowQNN`, if `RawFeatureVector` is used as feature map, the following warning arises during the network initialization step:

```
Cannot compute gradient operator! Continuing without gradients!
```

This behavior was originally pointed out on issue #95, but it was only noted for the case where `CircuitStateFn` is used, and I initially thought that it was `Opflow`-related. I have now noticed that this is not a `OpflowQNN` or `CircuitStateFn`  issue, but a more generalized one. I have tested out different examples using `RawFeatureVector` and QNNs, and all of them end up triggering this warning, including the code shown in the *Creating Your First Machine Learning Programming Experiment in Qiskit* section of this repo's ReadMe (not ideal). 

I have traced back this warning to the following error message:

- **"raw_feature_vector.py", line 170, in _define** --> `raise QiskitError("Cannot define a ParameterizedInitialize with unbound parameters")`

Binding the parameters looks like a solution to this message, but I am not 100% sure that it is what we want. After all, we want the feature map to be parametrized, and if we bind the parameters, it's not parametrized anymore.

During my investigation I have found some potential bugs in either `RawFeatureVector` or its base class `BlueprintCircuit`. I plan to make the corresponding issues in the Terra repo and link them below for reference, but I think that it is helpful to also keep track of the issue on the `qiskit-machine-learning` side. 

Run the following code:

```
        from qiskit import BasicAer
        from qiskit.utils import QuantumInstance, algorithm_globals
        from qiskit.algorithms.optimizers import COBYLA
        from qiskit.circuit.library import TwoLocal
        from qiskit_machine_learning.algorithms import VQC
        from qiskit_machine_learning.datasets import wine
        from qiskit_machine_learning.circuit.library import RawFeatureVector

        seed = 1376
        algorithm_globals.random_seed = seed

        # Use Wine data set for training and test data
        feature_dim = 4  # dimension of each data point
        training_size = 12
        test_size = 4

        # training features, training labels, test features, test labels as np.array,
        # one hot encoding for labels
        training_features, training_labels, test_features, test_labels = \
            wine(training_size=training_size, test_size=test_size, n=feature_dim)

        feature_map = RawFeatureVector(feature_dimension=feature_dim)
        ansatz = TwoLocal(feature_map.num_qubits, ['ry', 'rz'], 'cz', reps=3)
        vqc = VQC(feature_map=feature_map,
                  ansatz=ansatz,
                  optimizer=COBYLA(maxiter=100),
                  quantum_instance=QuantumInstance(BasicAer.get_backend('statevector_simulator'),
                                                   shots=1024,
                                                   seed_simulator=seed,
                                                   seed_transpiler=seed)
                  )
        vqc.fit(training_features, training_labels)

        score = vqc.score(test_features, test_labels)
        print('Testing accuracy: {:0.2f}'.format(score))
```

I would expect this warning not to appear. Furthermore, if you run the code above, you can see that the testing accuracy is still 100%, despite the warning. This is probably due to the simplicity of the dataset used (I have not found any other explanation at the moment, but haven't extensively tested this one either).

### Suggested solutions
The specific error that triggers this warning is:
- **"raw_feature_vector.py", line 170, in _define** --> `raise QiskitError("Cannot define a ParameterizedInitialize with unbound parameters")`

Doing something like `feature_map = feature_map.bind_parameters([bound1, bound2, bound3, bound4])` would prevent this error. However, this line of code does not work  as expected (neither does `feature_map = feature_map.assign_parameters()`), so these should be fixed first.

ElePT:
### Summary
This pull request fixes [this terra issue ](https://github.com/Qiskit/qiskit-terra/issues/6850), which was finally not terra-related. With this change, it is possible to retrieve the correct number of qubits when copying a `RawFeatureVector`, and this also fixes actions that involve copying, such as `bind_parameters()`, which currently does not support in-place execution (if I'm not mistaken).

### Details and comments

Code to check copying operation:

```
from qiskit_machine_learning.circuit.library import RawFeatureVector
feature_dim = 4  # dimension of each data point
feature_map = RawFeatureVector(feature_dimension=feature_dim)
print(feature_map.draw())
print(feature_map.num_qubits)
copy_map = feature_map.copy()
print(copy_map.draw())
print(copy_map.num_qubits)
```

Code to check parameter binding:

```
from qiskit_machine_learning.circuit.library import RawFeatureVector
feature_dim = 4  # dimension of each data point
feature_map = RawFeatureVector(feature_dimension=feature_dim)
print(feature_map.draw())
bound_map = feature_map.bind_parameters([0.25,0.25,0.25,0.25])
print(bound_map.draw())
print(bound_map.num_qubits)
```

This issue was found when looking into these other issues: #180, #95. 
Closes #180.

### DIFF
--- a/qiskit_machine_learning/circuit/library/raw_feature_vector.py
+++ b/qiskit_machine_learning/circuit/library/raw_feature_vector.py
@@ -103,6 +103,9 @@ class RawFeatureVector(BlueprintCircuit):
         Returns:
             The number of qubits.
         """
+        # ensure that num_qubits is up-to-date
+        if self._num_qubits is None and len(self.qubits) > 0:
+            self._num_qubits = len(self.qubits)
         return self._num_qubits if self._num_qubits is not None else 0
 
     @num_qubits.setter
--- a/qiskit_machine_learning/circuit/library/raw_feature_vector.py
+++ b/qiskit_machine_learning/circuit/library/raw_feature_vector.py
@@ -78,13 +78,6 @@ class RawFeatureVector(BlueprintCircuit):
 
     def _build(self):
         super()._build()
-        # we use len(_ordered_parameters) instead of feature_dimension
-        # to avoid checking if feature_dimension is None.
-        num_qubits = np.log2(len(self._ordered_parameters))
-
-        if num_qubits != self.num_qubits:
-            self._invalidate()  # rebuild circuit
-            self.num_qubits = num_qubits
 
         placeholder = ParameterizedInitialize(self._ordered_parameters[:])
         self.append(placeholder, self.qubits)
--- a/qiskit_machine_learning/circuit/library/raw_feature_vector.py
+++ b/qiskit_machine_learning/circuit/library/raw_feature_vector.py
@@ -76,7 +76,6 @@ class RawFeatureVector(BlueprintCircuit):
         super().__init__()
 
         self._ordered_parameters = ParameterVector("x")
-        self.num_qubits = None
         if feature_dimension is not None:
             self.feature_dimension = feature_dimension
 
@@ -146,7 +145,7 @@ class RawFeatureVector(BlueprintCircuit):
         if int(num_qubits) != num_qubits:
             raise ValueError("feature_dimension must be a power of 2!")
 
-        if self.num_qubits is None or num_qubits != self.num_qubits:
+        if num_qubits != self.num_qubits:
             self._invalidate()
             self.num_qubits = int(num_qubits)
 
--- a/qiskit_machine_learning/circuit/library/raw_feature_vector.py
+++ b/qiskit_machine_learning/circuit/library/raw_feature_vector.py
@@ -120,8 +120,8 @@ class RawFeatureVector(BlueprintCircuit):
         if self.num_qubits != num_qubits:
             # invalidate the circuit
             self._invalidate()
-            if num_qubits:
-                self.add_register(QuantumRegister(num_qubits, "q"))
+            if num_qubits > 0:
+                self.qregs = [QuantumRegister(num_qubits, name="q")]
             else:
                 self.qregs: List[QuantumRegister] = []
 
--- a/qiskit_machine_learning/circuit/library/raw_feature_vector.py
+++ b/qiskit_machine_learning/circuit/library/raw_feature_vector.py
@@ -152,9 +152,6 @@ class RawFeatureVector(BlueprintCircuit):
             self._invalidate()
             self.num_qubits = int(num_qubits)
 
-    def _invalidate(self):
-        super()._invalidate()
-
 
 class ParameterizedInitialize(Instruction):
     """A normalized parameterized initialize instruction."""
--- a/qiskit_machine_learning/circuit/library/raw_feature_vector.py
+++ b/qiskit_machine_learning/circuit/library/raw_feature_vector.py
@@ -80,7 +80,7 @@ class RawFeatureVector(BlueprintCircuit):
         super()._build()
         # we use len(_ordered_parameters) instead of feature_dimension
         # to avoid checking if feature_dimension is None.
-        num_qubits = np.log2(len(self._ordered_parameters[:]))
+        num_qubits = np.log2(len(self._ordered_parameters))
 
         if num_qubits != self.num_qubits:
             self._invalidate()  # rebuild circuit
--- a/qiskit_machine_learning/circuit/library/raw_feature_vector.py
+++ b/qiskit_machine_learning/circuit/library/raw_feature_vector.py
@@ -120,10 +120,10 @@ class RawFeatureVector(BlueprintCircuit):
         if self.num_qubits != num_qubits:
             # invalidate the circuit
             self._invalidate()
+            self.qregs: List[QuantumRegister] = []
             if num_qubits is not None and num_qubits > 0:
                 self.qregs = [QuantumRegister(num_qubits, name="q")]
-            else:
-                self.qregs: List[QuantumRegister] = []
+
 
     @property
     def feature_dimension(self) -> int:
--- a/qiskit_machine_learning/circuit/library/raw_feature_vector.py
+++ b/qiskit_machine_learning/circuit/library/raw_feature_vector.py
@@ -12,7 +12,7 @@
 
 """The raw feature vector circuit."""
 
-from typing import Optional
+from typing import Optional, List
 import numpy as np
 from qiskit.exceptions import QiskitError
 from qiskit.circuit import QuantumRegister, QuantumCircuit, ParameterVector, Instruction
@@ -123,7 +123,7 @@ class RawFeatureVector(BlueprintCircuit):
             if num_qubits:
                 self.add_register(QuantumRegister(num_qubits, "q"))
             else:
-                self.qregs = []
+                self.qregs: List[QuantumRegister] = []
 
     @property
     def feature_dimension(self) -> int:
--- a/qiskit_machine_learning/circuit/library/raw_feature_vector.py
+++ b/qiskit_machine_learning/circuit/library/raw_feature_vector.py
@@ -65,19 +65,24 @@ class RawFeatureVector(BlueprintCircuit):
     def __init__(self, feature_dimension: Optional[int]) -> None:
         """
         Args:
-            feature_dimension: The feature dimension and number of qubits.
+            feature_dimension: The feature dimension from which the number of
+                                qubits is inferred as ``n_qubits = log2(feature_dim)``
 
         """
         super().__init__()
 
-        self._num_qubits = None
         self._ordered_parameters = ParameterVector("x")
-
+        self.num_qubits = None
         if feature_dimension:
             self.feature_dimension = feature_dimension
 
     def _build(self):
         super()._build()
+        num_qubits = np.log2(len(self._ordered_parameters[:]))
+
+        if num_qubits != self.num_qubits:
+            self._invalidate()  # rebuild circuit
+            self.num_qubits = num_qubits
 
         placeholder = ParameterizedInitialize(self._ordered_parameters[:])
         self.append(placeholder, self.qubits)
@@ -103,10 +108,7 @@ class RawFeatureVector(BlueprintCircuit):
         Returns:
             The number of qubits.
         """
-        # ensure that num_qubits is up-to-date
-        if self._num_qubits is None and len(self.qubits) > 0:
-            self._num_qubits = len(self.qubits)
-        return self._num_qubits if self._num_qubits is not None else 0
+        return super().num_qubits
 
     @num_qubits.setter
     def num_qubits(self, num_qubits: int) -> None:
@@ -115,11 +117,13 @@ class RawFeatureVector(BlueprintCircuit):
         Args:
             The new number of qubits.
         """
-        if self._num_qubits != num_qubits:
+        if self.num_qubits != num_qubits:
             # invalidate the circuit
             self._invalidate()
-            self._num_qubits = num_qubits
-            self.add_register(QuantumRegister(self.num_qubits, "q"))
+            if num_qubits:
+                self.add_register(QuantumRegister(num_qubits, "q"))
+            else:
+                self.qregs = []
 
     @property
     def feature_dimension(self) -> int:
@@ -144,13 +148,12 @@ class RawFeatureVector(BlueprintCircuit):
         if int(num_qubits) != num_qubits:
             raise ValueError("feature_dimension must be a power of 2!")
 
-        if self._num_qubits is None or num_qubits != self._num_qubits:
+        if self.num_qubits is None or num_qubits != self.num_qubits:
             self._invalidate()
             self.num_qubits = int(num_qubits)
 
     def _invalidate(self):
         super()._invalidate()
-        self._num_qubits = None
 
 
 class ParameterizedInitialize(Instruction):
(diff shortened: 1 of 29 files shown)


woodsp-ibm:
Can you add unit test cases to ensure this works and continues to do so - maybe based off the sample code you show asserting that the copy looks/works like expected.

Can you also add a reno release not for this fix.

As long as the getter is used, normal for external users then this works out albeit it that it corrects the error later. Internally the code often directly accesses the _num_qubits variable - by fixing the mismatch later are we exposing ourselves to possible future bugs if code is added and uses the _num_qubits var directly. Can the mismatch be fixed earlier say by overriding copy and calling  super then fixing up the copy before returning it. This would then not have the circuit be in some inconsistent state until the getter is called.

ElePT:
@adekusar-drl @woodsp-ibm After checking that it's not used in any special way, I have ended up removing the `_num_qubits` variable for simplicity, and just added a few lines to the `_build` method. I have also included the checks for `num_qubits` and `feature_dimension` (this one should not be problematic, but just in case) in `test_raw_feature_vector.py`. I think that this should be all for this PR, unless you catch some other bug in my code 🐞 (or dislike this idea, of course).

ElePT:
By the way, the fail in `test_regressor_with_opflow_qnn` is not related to `RawFeatureVector`. This has happened before, and I have noticed that these mysterious fails usually happen under Python 3.7 (maybe it's just a coincidence, though).

qiskit_machine_learning/circuit/library/raw_feature_vector.py
```diff
@@ -112,11 +117,13 @@ def num_qubits(self, num_qubits: int) -> None:
         Args:
             The new number of qubits.
         """
-        if self._num_qubits != num_qubits:
+        if self.num_qubits != num_qubits:
             # invalidate the circuit
             self._invalidate()
-            self._num_qubits = num_qubits
-            self.add_register(QuantumRegister(self.num_qubits, "q"))
+            if num_qubits:
```
Cryoris, line 123: `if <variable>` is true for many different inputs and that led to some difficult to find bugs in the past. I think it's better to explicitly check for the value 🙂 
```suggestion
            if num_qubits > 0:
```

qiskit_machine_learning/circuit/library/raw_feature_vector.py
```diff
@@ -112,11 +117,13 @@ def num_qubits(self, num_qubits: int) -> None:
...
+                self.add_register(QuantumRegister(num_qubits, "q"))
```
Cryoris, line 124: You should probably reset the registers, otherwise calling the setter twice will add too many registers (and qubits!) and complain that "q" already exists:
```python
>>> from qiskit_machine_learning.circuit.library import RawFeatureVector
>>> r = RawFeatureVector(2)
>>> r.num_qubits = 1
>>> r.draw()
     ┌────────────────────────────────────┐
q_0: ┤ Parameterizedinitialize(x[0],x[1]) ├
     └────────────────────────────────────┘
>>> r.num_qubits = 3
  File "<stdin>", line 1, in <module>
  File "/Users/jul/Qiskit/qiskit-machine-learning/qiskit_machine_learning/circuit/library/raw_feature_vector.py", line 124, in num_qubits
    self.add_register(QuantumRegister(num_qubits, "q"))
  File "/Users/jul/Qiskit/qiskit-terra/qiskit/circuit/quantumcircuit.py", line 1179, in add_register
    raise CircuitError('register name "%s" already exists' % register.name)
qiskit.circuit.exceptions.CircuitError: 'register name "q" already exists'
```

qiskit_machine_learning/circuit/library/raw_feature_vector.py
ElePT: Good point 💫

qiskit_machine_learning/circuit/library/raw_feature_vector.py
ElePT: Yes!!! I actually ran into this situation with the original implementation but forgot about it. Thanks!

adekusar-drl (dismissed):
Looks good to me. @Cryoris do you see any issues in the PR?

qiskit_machine_learning/circuit/library/raw_feature_vector.py
Cryoris: Yes that would be good. Do you want to open a PR to fix it? 😇

qiskit_machine_learning/circuit/library/raw_feature_vector.py
```diff
@@ -65,19 +65,24 @@ class RawFeatureVector(BlueprintCircuit):
...
+                                qubits is inferred as ``n_qubits = log2(feature_dim)``
 
         """
         super().__init__()
 
-        self._num_qubits = None
         self._ordered_parameters = ParameterVector("x")
-
+        self.num_qubits = None
         if feature_dimension:
```
woodsp-ibm, line 76:
```suggestion
        if feature_dimension is not None:
```
I think this really should be an explicit check against None. (I realize this is how it was though but since things are being fixed here)

releasenotes/notes/fix-raw-feature-vector-copy-20534d595a0056de.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    This fix removes the `_num_qubits` variable of `RawFeatureVector`, which caused issues on the `copy()` and `bind_parameters()` methods.
```
woodsp-ibm, line 4: Since `_num_qubits` is a private variable, and the release note is really for user facing changes, where they should not be concerned with/accessing private aspects, can we just simply say something like RawFeatureVector was fixed so now copy and bind_parameter methods work correctly.

qiskit_machine_learning/circuit/library/raw_feature_vector.py
```diff
@@ -65,19 +65,24 @@ class RawFeatureVector(BlueprintCircuit):
...
             self.feature_dimension = feature_dimension
 
     def _build(self):
         super()._build()
+        num_qubits = np.log2(len(self._ordered_parameters[:]))
```
woodsp-ibm, line 81: The reason for using order parameters here rather than the instance feature dimension is because this is always set and no if None is needed? Since its set off feature dimension to start with then if the above is correct maybe add a comment for future readers of the code.

qiskit_machine_learning/circuit/library/raw_feature_vector.py
ElePT: Sure

releasenotes/notes/fix-raw-feature-vector-copy-20534d595a0056de.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    After this ``RawFeatureVector`` fix, ``copy()`` and ``bind_parameters()`` methods work correctly.
```
Cryoris, line 4: I think if users read the release notes they should understand what the issue was we fixed, so we could be a bit more detailed in the descriptions 🙂 
```suggestion
    Fix a bug, where :meth:`qiskit_machine_learning.circuit.library.RawFeatureVector.copy`
    didn't copy all internal settings which could lead to issues with the copied circuit.
    As a consequence :meth:`qiskit_machine_learning.circuit.library.RawFeatureVector.bind_parameters`
    is also fixed.
    didn't  ``RawFeatureVector`` fix, ``copy()`` and ``bind_parameters()`` methods work correctly.
```

test/circuit/library/test_raw_feature_vector.py
```diff
@@ -62,6 +62,10 @@ def test_fully_bound(self):
 
         self.assertEqual(bound.decompose(), ref)
 
+        # make sure that the bound circuit is a successful copy of the original circuit
+        self.assertEqual(circuit.num_qubits, bound.num_qubits)
+        self.assertEqual(circuit.feature_dimension, bound.feature_dimension)
```
Cryoris, line 67: Isn't this covered implicitly by the test in L63? (At least the number of qubits)

test/circuit/library/test_raw_feature_vector.py
```diff
@@ -82,6 +86,10 @@ def test_partially_bound(self):
             ref.initialize([0.2, 0.4, 0.4, 0.8], ref.qubits)
             self.assertEqual(bound.decompose(), ref)
 
+            # make sure that the bound circuit is a successful copy of the original circuit
+            self.assertEqual(circuit.num_qubits, bound.num_qubits)
+            self.assertEqual(circuit.feature_dimension, bound.feature_dimension)
```
Cryoris, line 91: See comment above

Cryoris (commented):
A few minor comments but otherwise LGTM 👍🏻

test/circuit/library/test_raw_feature_vector.py
adekusar-drl: Same as above

test/circuit/library/test_raw_feature_vector.py
adekusar-drl: I'd prefer to keep these line to have explicit tests. L63 compares versus a reference circuit, while here is a direct comparison.

releasenotes/notes/fix-raw-feature-vector-copy-20534d595a0056de.yaml
adekusar-drl: Updated as you suggested.

releasenotes/notes/fix-raw-feature-vector-copy-20534d595a0056de.yaml
adekusar-drl: Updated as @Cryoris suggested below.

adekusar-drl:
@woodsp-ibm please take a look, just minor updates. I hope we are good to go.

qiskit_machine_learning/circuit/library/raw_feature_vector.py
adekusar-drl: There's a comment added in the sources.

woodsp-ibm: I looked at the comment. I am still missing something I guess. If feature dimension is None what is len(ordered_parameters[:]). If that is 0, which one might think would be a valid value if num_qubits was not known, then that will be just as much as problem giving that to np.log2.

adekusar-drl: I spent sometime looking into `BlueprintCircuit` which I, honestly, hate, and realized that this additional check in `RawFeatureVector._build()` is not required anymore, hence `log2` is not required. The reason is that we call `super()._build()` first, which, in turns, calls `_check_configuration()` where parameters are tuned to the correct size and `RawFeatureVector._build()` works correctly afterwards.

adekusar-drl:
I also slightly updated the class docstring.

qiskit_machine_learning/circuit/library/raw_feature_vector.py
```diff
@@ -65,15 +69,15 @@ class RawFeatureVector(BlueprintCircuit):
...
+                                qubits is inferred as ``n_qubits = log2(feature_dim)``
 
         """
         super().__init__()
 
-        self._num_qubits = None
         self._ordered_parameters = ParameterVector("x")
-
-        if feature_dimension:
+        self.num_qubits = None
```
woodsp-ibm, line 79: Is this needed? If you read back, i.e. do a self.num_qubits does that not end up going to the property getter in QuantumCircuit? The _num_qubits and holding None I guess was done as a independent instance var on this class, same as in n_local. Hence the overridding of the num_qubits property - which now just calls super and as such could be deleted if thats what needed.

Now I note in the num_qubits setter that it does not update feature dimension (setting feature dim does update num qubits so it seems asymmetric in that regard). I am just noting that check_configuration does so against feature_dimension so hoepfully it never calls resize when that is still None.

qiskit_machine_learning/circuit/library/raw_feature_vector.py
```diff
@@ -141,14 +146,10 @@ def feature_dimension(self, feature_dimension: int) -> None:
         if int(num_qubits) != num_qubits:
             raise ValueError("feature_dimension must be a power of 2!")
 
-        if self._num_qubits is None or num_qubits != self._num_qubits:
+        if self.num_qubits is None or num_qubits != self.num_qubits:
```
woodsp-ibm, line 149: While self._num_qubits could have been None I do not think that is the case for this now which uses the QuantumCircuit getter property where 0 is what is returned for an empty cct.

qiskit_machine_learning/circuit/library/raw_feature_vector.py
adekusar-drl: Agree, this is not required anymore, this is a legacy from `_num_qubits`.

qiskit_machine_learning/circuit/library/raw_feature_vector.py
adekusar-drl: Corrected as well, no need for the additional check for `None`.

woodsp-ibm (approved):

PR #188, opened 2021-08-19
Issue 175 | Fix zero dim shaped gradients for NeuralNetworks

### LINKED ISSUE #175: Torchconnector fails in backward prop
In the backward propagation this code checks `input_grad` and if _not_ None under certain a condition sets it to None

https://github.com/Qiskit/qiskit-machine-learning/blob/00831061ca0206af42e131fcb9b605c4acd404fa/qiskit_machine_learning/connectors/torch_connector.py#L158-L160

later in the same method it makes this call

https://github.com/Qiskit/qiskit-machine-learning/blob/00831061ca0206af42e131fcb9b605c4acd404fa/qiskit_machine_learning/connectors/torch_connector.py#L178

which fails inside Torch einsum util with a complaint about NoneType (which seems to be about input_grad being None)

Similar logic is in the `weight_grad` part too.

I will note that the above logic path, where input_grad is set to None, is not covered by unit tests and so this failure does not arise there.

IceKhan13:
### Summary
Neural networks (`CircuitQNN`) currently failing on backward prop when input parameters or weight parameters are empty.

### Details and comments

This is a quick fix to use additional check in `_validate_backward_output`. 

Closes #175


IceKhan13:
Ideal option would be avoiding zero dim in shapes of gradients at all (a.k.a fallback earlier). This will require to refactor of `CircuitQNN` and similar classes inherited from `NeuralNetwork`.

adekusar-drl:
Is the PR ready for review or something is still missing?

IceKhan13:
> Is the PR ready for review or something is still missing?

Yes, quick fix is ready

IceKhan13:
Hm, latest errors from linter came from main branch

https://github.com/Qiskit/qiskit-machine-learning/pull/188/checks?check_run_id=3400590479

and I cannot reproduce them on my local machine

Screen Shot 2021-08-23 at 11 53 13 AM

adekusar-drl (commented):
Thanks a lot for taking care of the issue. As I understand from the tests, the problem arises when one tries to use `TorchConnector` without actual values bound to parameters? If so, could it be better to raise an exception with an explanation what happened?
Since this is a bug fix a reno file is required.

IceKhan13:
> could it be better to raise an exception with an explanation what happened?

Do you mean non parametric inputs or weights is not a valid case at all? So, developers can not define CQNN without parametric inputs?

> Since this is a bug fix a reno file is required.

Added release note

adekusar-drl (approved):
Thanks for taking care of this issue. I think, it is a viable solution.

PR #201, opened 2021-08-27
Use callbacks in Neural Network Classifier and Regressor tutorial (#195)

### LINKED ISSUE #195: Make use of callbacks in the Neural Network Classifier & Regressor tutorial
Training a classifier/regressor in the "Neural Network Classifier & Regressor" tutorial may take some time, so to make the tutorial more interactive and provide some insights into the training process I suggest to make use of the recently introduced callback interface and collect objective(loss) function values obtained from an optimizer. These values can be just printed, thus making the notebook interactive and/or plotted. Thus, we will achieve two goals: dynamic behaviour of the tutorial during long training processes and show how to use callbacks.

epelaaez:
### Summary
Integrates callbacks into the Neural Network Classifier and Regressor tutorial to make it more interactive and demonstrate the use of the newly added `callback` parameter.

### Details and comments
I added an example of the live plot I mentioned in the linked issue (closes #195). Let me know any feedback and/or things that need to be changed so I can go ahead and implement it for the four classes present in this tutorial.


docs/tutorials/02_neural_network_classifier_and_regressor.ipynb
```diff
@@ -152,51 +154,80 @@
...
+    "# callback function that draws a live plot when the .fit() method is called\n",
+    "def callback(x, y):\n",
+    "    clear_output(wait=True)\n",
+    "    objective_func_vals.append(y)\n",
+    "    plt.title(\"Objective function value against iteration for NeuralNetworkClassifier\")\n",
+    "    plt.xlabel(\"Iteration\")\n",
+    "    plt.ylabel(\"Objective function value\")\n",
+    "    plt.plot(range(len(objective_func_vals)), objective_func_vals)\n",
+    "    plt.show()\n",
+    "\n",
```
adekusar-drl, line 171: I'd rather add a text cell explaining what callback is and how it can be used. Other than that looks nice. I hope, the callback function can be reused across different models.

adekusar-drl (commented):
Thanks, looks great! This is what was expected, if you could add plots for other models that would be perfect.

docs/tutorials/02_neural_network_classifier_and_regressor.ipynb
epelaaez: I've added a text cell explaining the callback function just before a code cell only containing the callback function. Let me know if the content in the text cell is alright or if it should be changed.

epelaaez:
I added the live plot for all the models in the tutorial.

I've also added a text cell talking about the callback function just before the first model. Let me know if the content in this text cell is okay or if I should make any changes. Furthermore, the same callback function is passed to all the models to avoid declaring it before every model.

adekusar-drl (approved):
Thanks, looks good to me!

PR #209, opened 2021-09-09
Batching of circuits to overcome memory issues when using statevector simulator

rsln-s:
### Summary

Currently `batch_size` parameter is ignored if the statevector simulator is used. This leads to unreasonable memory use due to the size of the transpiled circuits, up to 1TB of RAM for 800 by 800 kernel matrix and 20 qubits (see [qiskit-terra, issue #6991](https://github.com/Qiskit/qiskit-terra/issues/6991)). This pull request fixes this by transpiling and simulating circuits in batches, never storing the entire 800 circuits. The modification uses `batch_size` parameter that is already used in non-statevector case.

### Details and comments

I had success by setting `batch_size=50` (memory footprint down to <20 GB).

### DIFF
--- a/qiskit_machine_learning/kernels/quantum_kernel.py
+++ b/qiskit_machine_learning/kernels/quantum_kernel.py
@@ -159,7 +159,7 @@ class QuantumKernel:
         """
         if is_statevector_sim:
             # |<0|Psi^dagger(y) x Psi(x)|0>|^2, take the amplitude
-            v_a, v_b = [results.get_statevector(int(i)) for i in idx]
+            v_a, v_b = [results[int(i)] for i in idx]
             tmp = np.vdot(v_a, v_b)
             kernel_value = np.vdot(tmp, tmp).real  # pylint: disable=no-member
         else:
@@ -287,11 +287,18 @@ class QuantumKernel:
                 for x in to_be_computed_data
             ]
 
-            results = self._quantum_instance.execute(circuits)
+            statevectors = []
+
+            for min_idx in range(0, len(circuits), self._batch_size):
+                max_idx = min(min_idx + self._batch_size, len(circuits))
+                num_circuits = max_idx - min_idx
+                results = self._quantum_instance.execute(circuits[min_idx : max_idx])
+                for j in range(num_circuits):
+                    statevectors.append(results.get_statevector(j))
 
             offset = 0 if is_symmetric else len(x_vec)
             matrix_elements = [
-                self._compute_overlap(idx, results, is_statevector_sim, measurement_basis)
+                self._compute_overlap(idx, statevectors, is_statevector_sim, measurement_basis)
                 for idx in list(zip(mus, nus + offset))
             ]
 
--- a/qiskit_machine_learning/kernels/quantum_kernel.py
+++ b/qiskit_machine_learning/kernels/quantum_kernel.py
@@ -282,18 +282,16 @@ class QuantumKernel:
                 is_statevector_sim=is_statevector_sim,
             )
             parameterized_circuit = self._quantum_instance.transpile(parameterized_circuit)[0]
-            circuits = [
-                parameterized_circuit.assign_parameters({feature_map_params: x})
-                for x in to_be_computed_data
-            ]
-
             statevectors = []
 
-            for min_idx in range(0, len(circuits), self._batch_size):
-                max_idx = min(min_idx + self._batch_size, len(circuits))
-                num_circuits = max_idx - min_idx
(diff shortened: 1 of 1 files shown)


adekusar-drl:
@rsln-s Thanks a lot for your contribution. Could you please add a reno file with a bug fix description? Also, if possible add a test to cover this fix.

@attp could you please take a look at the changes?

rsln-s:
Added reno file. The correctness of the output of the code is checked by tests already in place; testing the memory usage may be complicated within constraints of gh-actions.

releasenotes/notes/batch-circuits-statevector-522a842c6f68d954.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
+  - |
+    Fixed a bug in ``QuantumKernel`` where for statevector simulator
+    all circuits were constructed and transpiled at once, leading to high memory
+    usage. Now the circuits are batched similarly to how it was prevously
```
woodsp-ibm, line 6:
```suggestion
    usage. Now the circuits are batched similarly to how it was previously
```

woodsp-ibm:
The spell checker also fails on this `transpiled`  word from the release note. If you add it to `.pylintdict` file in the root, our custom dictionary, since its correctly spelt then it should pass CI. (You will see the words are in lowercase in alphabetic order so just add it appropriately)

rsln-s:
Updated the `.pylintdict` dictionary

adekusar-drl (approved):
Looks good to me, thanks a lot!

adekusar-drl:
@woodsp-ibm Do you have any comments?

woodsp-ibm:
Can I ask  you to make a minor change to the docstring for batch_size in the constructor. It says

>batch_size: Number of circuits to batch together for computation. Default 1000.

specifically default 1000 yet the code has

>batch_size: int = 900,

So it should state the default is 900. I think it may have been 1000 at one point but was changed, if I recall correctly, to 900 to fit more with the limits around the provider.

In terms of testing you say its covered by the current unit tests. From what I can see there is nothing explicitly testing that aspect. Of course batch_size defaults to 900 and is in main path. I am not sure what we do in test is ever affected by the batch size since from what I can see the tests are pretty small. Having said that if we had a test that dropped the batch size down I am not sure how to test it given its behavior is internal to evaluate - the only way comes to mind is hooking the quantum instance execute and checking the number of circuits is as expected along the way in addition the the final result being as expected.

attp:
Looks good to me. 

Regarding the `batch_size` docstring, you are correct @woodsp-ibm the default was originally 1000, and we changed it to 900 to match the backend limits and reduce the number of jobs sent. I must have missed updating the docstring in the PR.

rsln-s:
Updated the docstring as requested by @woodsp-ibm.

woodsp-ibm:
Just a note. Locally I changed the QuantumInstance `execute` method to print the number of circuits it was passed and ran the kernel unit tests. Its quite a small number so nowhere near the 900 limit - in fact its not even in double digits. Anyway I changed default back size in the kernel to a much small number and things worked but not all numbers were limited - presumably the ones via statevector usage as I was doing this from the main branch. Cloning your fork and doing the same from there then all the counts were limited - in fact I set it to 1 as a test and it printed all 1's and passed. So it seems its working ok. It would be nice if the test did somehow test out the batch size but maybe that could/should be raised as a separate issue. @adekusar-drl any thoughts here - you commented early on about a test around the fix.

adekusar-drl:
@woodsp-ibm When I mentioned unit tests I did not have anything special on my mind. In general, your idea of setting batch size to 1 and then running a test on the statevector simulator make sense.

@rsln-s What do you think?

woodsp-ibm:
> In general, your idea of setting batch size to 1 and then running a test on the statevector simulator make sense.
I had done it so it applied to the qasm mode as well - since it appeared not tested in general. While the final outcome can be checked as currently, what is more complicated to do is to check that indeed the batch size is limiting the number of internal computations (circuits). My only thought on perhaps how to do such a check. as I mentioned in an earlier comment was about hooking the QuantumInstance execute method on the instance used with the backend such that the number of circuits given to the method could be fairly easily intercepted and checked - i.e hook it and do whatever was needed to check then call over to the original method that was hooked so that the circuit results from execute can be returned.

adekusar-drl:
@woodsp-ibm I approve, merge this PR and open an issue to improve tests for QuantumKernel. Any thoughts?

woodsp-ibm (approved):
> merge this PR and open an issue to improve tests for QuantumKernel. Any thoughts?

@adekusar-drl that seems fine by me.

PR #211, opened 2021-09-13
Replace calls to the `optimize` method with the corresponding `minimize`

### LINKED ISSUE #212: Update variational algorithms to make use of a new interface of `Optimizer`
Calls to `Optimizer.optimize` must be replaced with `Optimizer.minimize()` as of https://github.com/Qiskit/qiskit-terra/issues/6382. The former method is deprecated now.

adekusar-drl:
### Summary
This is a follow up PR to make use of the changes introduced in https://github.com/Qiskit/qiskit-terra/issues/6382.
The `Optimizer.optimize()` method is deprecated in favor of `Optimizer.minimize()` and this is fixed in this PR.
Closes #212.

### DIFF
--- a/qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
+++ b/qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
@@ -104,11 +104,10 @@ class NeuralNetworkClassifier(TrainableModel, ClassifierMixin):
 
         objective = self._get_objective(function)
 
-        self._fit_result = self._optimizer.optimize(
-            self._neural_network.num_weights,
-            objective,
-            function.gradient,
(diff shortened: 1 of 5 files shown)


manoelmarques (approved):

woodsp-ibm:
The optimizer change is for Terra 0.19 - the requirements.txt file ought to have been updated to make the min level of Terra 0.19.0 since it will no longer work on prior versions.

Now when Terra 0.19 is release ML will start emitting deprecating warnings. And while the ML code will still work this often causes questions from people. Hence we may want to consider an updated ML release at some point.

adekusar-drl:
@woodsp-ibm Sorry about that, there's #214 to update Terra requirements.

PR #213, opened 2021-09-13
Fix forward pass computation cache

### LINKED ISSUE #204: AQGD does not return objective function values properly.
### Information

- **Qiskit Machine Learning version**: Tested on 0.3.0 and 0.2.1
- **Python version**: 3.7.9
- **Operating system**: OSX

### What is the current behaviour?
Returns the same value for all iterations. Graph produced by `callback_graph` in Tutorial 2 looks like this:
Screenshot 2021-09-05 at 18 00 50

Change optimiser used in Tutorial 2 in docs for machine learning.

### What is the expected behaviour?
Changing values to be able to monitor convergence as intended. 

### Suggested solutions
No clue.

adekusar-drl:
### Summary
Objective functions cache result of forward computation. Initially this cache was based on the identifier (`a call of id()`) of the weights arraay. AQGD re-uses the same array for weights: it updates the values keeping an instance of the array the same. This caused to re-use the same forward pass value across all iteration. The forward pass cache now is based on actual values of weights.
Closes #204.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -43,6 +43,7 @@ jobs:
           QISKIT_DOCS_BUILD_TUTORIALS: 'always'
         run: |
           echo "earliest_version: 0.1.0" >> releasenotes/config.yaml
+          tools/ignore_untagged_notes.sh
           make html SPHINXOPTS=-W
           tools/deploy_documentation.sh
         shell: bash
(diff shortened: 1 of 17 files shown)


adekusar-drl:
@woodsp-ibm Do we need a reno here? What do you think?

woodsp-ibm:
It should have a reno when the bug affects end users.

woodsp-ibm:
Should this be backported to stable - I don't know its severe enough to warrant a bug fix release just for this, but it can at least be backported ready I think.

adekusar-drl:
Added reno, added backport label.

adekusar-drl:
Seems like I messed up with the `stable backport potential` label and now the PR can't pass the CI.

adekusar-drl:
@woodsp-ibm Could you please take a look? Seems all good now.

PR #219, opened 2021-09-15
Add support for tunable feature maps in QuantumKernel class.

### LINKED ISSUE #215: Implement Quantum Kernel Alignment Algorithm
Implement the quantum kernel alignment (**QKA**) algorithm as a Qiskit application. Generally, the inputs/outputs will be:

**Inputs**:
    * feature_map: A `QuantumCircuit` object which maps input data to a quantum feature space
    * data: A labeled dataset
    * optimizer: An `Optimizer` object to be used during kernel tuning
    * backend: A quantum backend instance

**Outputs**:
    * optimal_params: Iterable of real values associated with feature map free parameters
    * kernel_matrix: 2D np.ndarray of size NxN representing the kernel function, N=num_training_samples
    
    
### Implementation Plan
Before the **QKA** algorithm can be implemented, some modifications to Qiskit are necessary. These 3 pieces will be submitted consecutively as sub-branches of one another:

- [x] Modify `QuantumKernel` class to facilitate binding and re-binding tunable feature map parameters. `qiskit-machine-learning.kernels.QuantumKernel (QK)` only supports feature map circuits which contain a number of parameters equal to the dimensionality of the input data, and it provides no interface for clients to update/bind parameters manually. 

- [x] Implement a `QuantumKernelTrainer` application which provides a general interface for kernel tuning algorithms, such as **QKA**, to interact with `QSVC` and `QSVR` classes to find optimized kernel functions. This includes implementing kernel-based loss functions.

calebj15:
### Summary
Closes #215 
Currently, qiskit-machine-learning.kernels.QuantumKernel (QK) only supports feature map circuits which contain a number of parameters equal to the dimensionality of the input data, and it provides no interface for clients to update/bind parameters manually. This makes it cumbersome for clients to use QK within algorithms which require many updates to feature map parameters.

This pull request provides updates to the QuantumKernel class which facilitate binding and re-binding tunable feature map parameters.

### Details and comments
All source code changes are contained within `qiskit_machine_learning.kernels.quantum_kernel.py` and the associated test module.

### DIFF
--- a/qiskit_machine_learning/kernels/quantum_kernel.py
+++ b/qiskit_machine_learning/kernels/quantum_kernel.py
@@ -12,12 +12,14 @@
 
 """Quantum Kernel Algorithm"""
 
-from typing import Optional, Union
+from typing import Optional, Union, Sequence, Mapping
+import copy
+import numbers
 
 import numpy as np
 
 from qiskit import QuantumCircuit, QuantumRegister, ClassicalRegister
-from qiskit.circuit import ParameterVector
+from qiskit.circuit import Parameter, ParameterVector
 from qiskit.circuit.library import ZZFeatureMap
 from qiskit.providers import Backend, BaseBackend
 from qiskit.utils import QuantumInstance
@@ -50,6 +52,7 @@ class QuantumKernel:
         enforce_psd: bool = True,
         batch_size: int = 900,
         quantum_instance: Optional[Union[QuantumInstance, BaseBackend, Backend]] = None,
+        free_parameters: Optional[Union[ParameterVector, Sequence[Parameter]]] = None,
     ) -> None:
         """
         Args:
@@ -59,11 +62,21 @@ class QuantumKernel:
                 Only enforced when not using the state vector simulator. Default True.
             batch_size: Number of circuits to batch together for computation. Default 1000.
             quantum_instance: Quantum Instance or Backend
+            free_parameters: Iterable containing Parameter objects which correspond to
+                free parameters (non-data-bound parameters) in the feature map circuit
         """
         self._feature_map = feature_map if feature_map else ZZFeatureMap(2)
         self._enforce_psd = enforce_psd
         self._batch_size = batch_size
         self._quantum_instance = quantum_instance
+        self._free_parameters = free_parameters
+        self._base_feature_map = copy.deepcopy(self._feature_map)
+        if free_parameters:
+            self._free_param_binds = {
+                free_parameters[i]: free_parameters[i] for i in range(len(free_parameters))
+            }
+        else:
+            self._free_param_binds = None
 
     @property
     def feature_map(self) -> QuantumCircuit:
@@ -74,6 +87,14 @@ class QuantumKernel:
     def feature_map(self, feature_map: QuantumCircuit):
         """Sets feature map"""
         self._feature_map = feature_map
+        self._base_feature_map = copy.deepcopy(self._feature_map)
+        self._free_parameters = None
+        self._free_param_binds = None
+
+    @property
+    def base_feature_map(self) -> QuantumCircuit:
+        """Returns unbound feature map"""
+        return copy.deepcopy(self._base_feature_map)
 
     @property
     def quantum_instance(self) -> QuantumInstance:
@@ -89,6 +110,87 @@ class QuantumKernel:
             self._quantum_instance = QuantumInstance(quantum_instance)
         else:
             self._quantum_instance = quantum_instance
+
+    @property
+    def free_parameters(self) -> Union[ParameterVector, Sequence[Parameter]]:
+        """Return the vector of free parameters."""
+        return self._free_parameters
+
+    @free_parameters.setter
+    def free_parameters(self, free_params: Union[ParameterVector, Sequence[Parameter]]):
+        """Sets the free parameters"""
+        self._free_param_binds = {free_params[i]: free_params[i] for i, _ in enumerate(free_params)}
+        self._free_parameters = free_params
+
+    def assign_free_parameters(
+        self, values: Union[Mapping[Parameter, float], Sequence[float]]
+    ) -> None:
+        """
+        Bind free parameters in the QuantumKernel feature map.
+
+        Args:
+            values (dict or iterable): {parameter: value, ...} or [value1, value2, ...]
+
+        Return:
+            None
+
+        Raises:
+            ValueError: Incompatible number of free parameters and values
+        """
+        if isinstance(values, dict):
+            if self._free_param_binds is None:
+                self._free_param_binds = values
+            else:
+                self._free_param_binds.update(values)
+            self._feature_map = self._base_feature_map.assign_parameters(self._free_param_binds)
+
+        else:
+            if self._free_parameters is None:
+                raise ValueError(
+                    """
+                    An iterable of free parameter values was given as input,
+                    but the number of parameter values ({}) does not match the
+                    number of free parameters tracked by the QuantumKernel (None).
+                    """.format(
+                        len(values)
+                    )
+                )
+            if len(values) != len(self._free_parameters):
+                raise ValueError(
+                    """
+                    An iterable of free parameter values was given as input,
+                    but the number of parameter values ({}) does not match the
+                    number of free parameters tracked by the QuantumKernel ({}).
+                    """.format(
+                        len(values), len(self._free_parameters)
+                    )
+                )
+
+            param_binds = {param: values[i] for i, param in enumerate(self._free_parameters)}
+            if self._free_param_binds is None:
+                self._free_param_binds = param_binds
+            else:
+                self._free_param_binds.update(param_binds)
+            self._feature_map = self._base_feature_map.assign_parameters(self._free_param_binds)
+
+    @property
+    def free_param_binds(self) -> Mapping[Parameter, float]:
+        """Return a copy of the current free parameter mappings for the feature map circuit."""
+        return copy.deepcopy(self._free_param_binds)
+
+    def bind_free_parameters(
+        self, values: Union[Mapping[Parameter, float], Sequence[float]]
+    ) -> None:
+        """
+        Alternate function signature for assign_free_parameters
+
+        Args:
+            values (dict or iterable): {parameter: value, ...} or [value1, value2, ...]
+
+        Return:
+            None
+        """
+        self.assign_free_parameters(values)
 
     def construct_circuit(
         self,
@@ -116,7 +218,27 @@ class QuantumKernel:
         Raises:
             ValueError:
                 - x and/or y have incompatible dimension with feature map
+                - unbound free parameters in the feature map circuit
         """
+        if self._free_param_binds is None:
+            unbound_free_params = None
+        else:
+            # Get all free parameters not associated with numerical values
+            unbound_free_params = [
+                val
+                for val in self._free_param_binds.values()
+                if not isinstance(val, numbers.Number)
+            ]
+        if unbound_free_params:
+            raise ValueError(
+                """
+                The feature map circuit contains unbound free parameters ({}).
+                All free parameters must be bound to numerical values before constructing
+                inner product circuit.
+                """.format(
+                    unbound_free_params
+                )
+            )
 
         if len(x) != self._feature_map.num_parameters:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


calebj15:
Reopening this PR due to some commits not being attached to a git user.

adekusar-drl (commented):
Thanks a lot for the contribution! Please pay attention at the duplicated code and try to remove it. The rest are minor things, I hope.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -90,6 +111,87 @@ def quantum_instance(
...
+        return self._free_parameters
+
+    @free_parameters.setter
+    def free_parameters(self, free_params: Union[ParameterVector, Sequence[Parameter]]):
+        """Sets the free parameters"""
+        self._free_param_binds = {free_params[i]: free_params[i] for i, _ in enumerate(free_params)}
+        self._free_parameters = free_params
+
+    def assign_free_parameters(
+        self, values: Union[Mapping[Parameter, float], Sequence[float]]
```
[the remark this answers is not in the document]
calebj15: I have updated the float hint to ParameterValueType. I'm not sure we want to provide an inplace option here, since we are essentially always spawning a copy of the unbound_feature_map when we assign parameters.

I don't believe QuantumCircuit provides a mechanism to bind and re-bind parameters in place?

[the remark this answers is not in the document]
adekusar-drl: That's right, you can't re-bind parameters, once they are bound they are excluded from the list of parameters.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -59,11 +63,14 @@ def __init__(
                 Only enforced when not using the state vector simulator. Default True.
             batch_size: Number of circuits to batch together for computation. Default 1000.
             quantum_instance: Quantum Instance or Backend
+            free_parameters: Iterable containing Parameter objects which correspond to
+                free parameters (non-data-bound parameters) in the feature map circuit
```
woodsp-ibm, line 67: Can we say something more here in this docstring to better explain them. I imagine whatever these free parameters are that the user of this class needs to set their values ahead of the kernel being evaluated so in this case the user of any quantum kernel instance ends up being responsible for them - to that end they are more like 'user parameters' in that regard.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -90,6 +105,115 @@ def quantum_instance(
         else:
             self._quantum_instance = quantum_instance
 
+    @property
+    def free_parameters(self) -> Union[ParameterVector, Sequence[Parameter]]:
+        """Return the vector of free parameters."""
+        return self._free_parameters
```
woodsp-ibm, line 111: We normally define instance vars in the constructor so the in an overview of the instance etc there - maybe initialized to None at that point if a value cannot be determined. I just commented against this one but there are others including the change in init() to set feature map via the setter rather than define/set the _feature_map var.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -90,6 +105,115 @@ def quantum_instance(
...
+        Raises:
+            ValueError: Incompatible number of free parameters and values
+        """
+        if not isinstance(values, dict):
+            if self._free_parameters is None:
+                raise ValueError(
+                    f"""
+                    An iterable of free parameter values was given as input,
+                    but the number of parameter values ({len(values)}) does not match the
+                    number of free parameters tracked by the QuantumKernel (None).
```
woodsp-ibm, line 155: Wouldn't it be simpler for the error message to be more like 'An iterable of free parameters was given but no free parameters were defined for the QuantumKernel` - it matched before more the length mismatch one below but I think its kinda wordy/complicated for this case

woodsp-ibm:
Given that this PR introduces a new feature it should include reno release note. (See https://github.com/Qiskit/qiskit-machine-learning/blob/main/CONTRIBUTING.md#release-notes for more info if you need it)

adekusar-drl:
@calebj15 I'm not quite sure I understand your intentions, but right now in this PR we have updates in `QuantumKernel` and a new `QuantumKernelTrainer`. I thought there would be 2 PRs for these two features.

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -52,11 +52,7 @@ def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwar
         if "random_state" not in kwargs:
             kwargs["random_state"] = algorithm_globals.random_seed
 
-        super().__init__(
-            kernel=self._quantum_kernel.evaluate,
-            *args,
-            **kwargs,
-        )
+        super().__init__(kernel=self._quantum_kernel.evaluate, *args, **kwargs)
```
calebj15, line 55: `make black` is automatically reformatting this for me. I suppose I should leave it and allow it to be merged into main?

test/kernels/test_qkernel.py
```diff
@@ -358,5 +359,73 @@ def test_ydim(self):
...
+        qkclass.bind_user_parameters(param_binds)
+
+        # Ensure the old values are properly overwritten in the feature map
+        bound_user_param_vals = [
+            val for val in qkclass.user_param_binds.values() if isinstance(val, numbers.Number)
+        ]
+        self.assertEqual(user_param_values, bound_user_param_vals)
+
+        # Ensure unbound feature map still holds all parameters of input feature map
+        self.assertEqual(len(qkclass.unbound_feature_map.parameters), 4)
```
woodsp-ibm, line 427: This test seems to comprise a series of tests. If an assert fails then the rest of the test will not be run. If the test aspects are independent, which at first glance they seem to be, then it is better to use the subTest capability (assuming that its not perhaps better to have individual tests for each part). Each subTest will be run even if one fails - idea being you get a fuller picture of what is/is not working then if the test fails at the first bit. See this for an example https://github.com/Qiskit/qiskit-machine-learning/blob/223c4127cf8874afeb349eba86d6bca1f498d14f/test/circuit/library/test_raw_feature_vector.py#L35-L49

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -59,21 +63,39 @@ def __init__(
...
+        self._feature_map = None
+        self._unbound_feature_map = None
+        self._user_parameters = None
+        self._user_param_binds = None
         self._enforce_psd = enforce_psd
         self._batch_size = batch_size
         self._quantum_instance = quantum_instance
 
+        self.feature_map = feature_map if feature_map else ZZFeatureMap(2)
+        self.user_parameters = user_parameters
```
adekusar-drl, line 80: I think these declarations are not required since we have kind of private class members (`_feature_map`, `_user_parameters` and property setters/getters.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -90,6 +112,108 @@ def quantum_instance(
...
 
+    @property
+    def user_parameters(self) -> Union[ParameterVector, Sequence[Parameter]]:
+        """Return the vector of user parameters."""
+        return self._user_parameters
+
+    @user_parameters.setter
+    def user_parameters(self, user_params: Union[ParameterVector, Sequence[Parameter]]) -> None:
+        """Sets the user parameters"""
+        if user_params:
```
adekusar-drl, line 123: This `if` statement implies that `user_params` may be `None`. 
So, two things:

- If `None` is possible, then perhaps the type hint should be `Optional[Union[ParameterVector, Sequence[Parameter]]]`?
- Better to have an explicit test for `None`: `if user_params is not None:`

Am I missing anything here? I suspect that `user_params` should not be `None`.

adekusar-drl (commented):
All good aside two minor things: redundant declarations of public properties and an ambiguous `if` statement.

qiskit_machine_learning/kernels/quantum_kernel.py
calebj15: Each of these  setters contains a good amount of logic, so I kept these lines in here to keep the constructor clean.

qiskit_machine_learning/kernels/quantum_kernel.py
calebj15: The logic is a little funny here. If users want to use `QuantumKernel` in the traditional way (no user parameters), this field will stay `None`; however the setter is always called in the constructor, so I need to handle the 'None' case.

I will take your advice of explicitly handling the `None` case.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: > 
> Each of these setters contains a good amount of logic, so I kept these lines in here to keep the constructor clean.

Ah, I see now, I forgot about setters. You're right.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: > I will take your advice of explicitly handling the `None` case.

OK, thanks. Looks good.

adekusar-drl:
@woodsp-ibm This PR looks good to me. Is there anything left open?

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -90,6 +112,108 @@ def quantum_instance(
...
+                copy of the param key as a value causes that parameter to be ignored.
+                This allows the circuit's user parameters to be bound consecutively.
+
+                Iterable [value_0, value_1, ..., value_N]:
+                If a list of real values is passed, the elements are assigned to the parameters
+                stored in the user_parameters field.
+
+        Return:
+            None
+
```
woodsp-ibm, line 153: I don't think you need to add anything if nothing is returned

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -90,6 +112,108 @@ def quantum_instance(
         else:
             self._quantum_instance = quantum_instance
 
+    @property
+    def user_parameters(self) -> Union[ParameterVector, Sequence[Parameter]]:
+        """Return the vector of user parameters."""
+        return self._user_parameters
+
+    @user_parameters.setter
+    def user_parameters(self, user_params: Union[ParameterVector, Sequence[Parameter]]) -> None:
```
woodsp-ibm, line 121: The typehints here should inlcude Optional I think since the getter will return None if user parameters are never set, and the setter allows you to pass in None presumably to remove any user parameters and let them all be controlled by the kernel.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -90,6 +112,108 @@ def quantum_instance(
...
+            ValueError: Incompatible number of user parameters and values
+        """
+        if not isinstance(values, dict):
+            if self._user_parameters is None:
+                raise ValueError(
+                    """
+                    An iterable of user parameter values was given, but no user
+                    parameters were defined for the QuantumKernel.
+                    """
+                )
```
woodsp-ibm, line 164: Should this check be outside of the above check of whatever the type of values is? No matter the type, if there were no user defined parameters set then this should raise an error right independent of the form of values being passed in?

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -90,6 +112,108 @@ def quantum_instance(
...
+        self, values: Union[Mapping[Parameter, float], Sequence[float]]
+    ) -> None:
+        """
+        Alternate function signature for assign_user_parameters
+
+        Args:
+            values (dict or iterable): {parameter: value, ...} or [value1, value2, ...]
+
+        Return:
+            None
```
woodsp-ibm, line 200: Similarly here - I don't think anything needs to be said if it doesn't return anything

qiskit_machine_learning/kernels/quantum_kernel.py
calebj15: If an iterable is passed, but no user parameters have been specified, we throw an error because we have no information for how to map the values.

If a dict is passed, we are able to find the values to the Parameters "blindly" by simply passing them straight down to the QuantumCircuit API for assignment/binding.

woodsp-ibm: If there are no user parameters on the kernel though why would we let it bind that way - seems odd to let the user bind parameters when the kernel has none even though you might not need an additional code as such for the binding. Arguably this way you can bind parameters without ever defining them as user parameters - or am I missing something

adekusar-drl:
@woodsp-ibm Is there anything left open after the recent updates?

qiskit_machine_learning/kernels/quantum_kernel.py
woodsp-ibm: Ok, but I think the getter above should be Optional for the return type since it will return None by default when there are no user parameters right. And not allowing the setter to take None I guess means that you cannot go back to having no user parameters once you chose to have some - aside from setting a new feature map that has the 'side-effect' of removing them.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -90,6 +114,94 @@ def quantum_instance(
...
+    def bind_user_parameters(self, values: Sequence[float]) -> None:
+        """
+        Alternate function signature for assign_user_parameters
+
+        Args:
+            values (iterable): [value1, value2, ...]
+        """
+        self.assign_user_parameters(values)
+
+    def unbound_user_parameters(self) -> List[Parameter]:
```
woodsp-ibm, line 192: I guess this could be a property if it was desired. Should the docstring say 'Returns a list of the unbound....' - yield is something particular in Python

Issue #218, opened 2021-09-15
Printing a QSVC instance raises an exception


adekusar-drl:
### Information

- **Qiskit Machine Learning version**: latest sources
- **Python version**: all
- **Operating system**: all

### What is the current behavior?
If you print an instance of `QSVC` an exception is risen:
```
RuntimeError: scikit-learn estimators should always specify their parameters in the signature of their __init__ (no varargs). <class 'qiskit_machine_learning.algorithms.classifiers.qsvc.QSVC'> with constructor (self, *args, quantum_kernel: Union[qiskit_machine_learning.kernels.quantum_kernel.QuantumKernel, NoneType] = None, **kwargs) doesn't  follow this convention.
```

### Steps to reproduce the problem
Snippet to reproduce the exception:
```python
from qiskit_machine_learning.algorithms import QSVC
from qiskit_machine_learning.kernels import QuantumKernel

qkernel = QuantumKernel()

qsvc = QSVC(quantum_kernel=qkernel)
print(qsvc)
```

### What is the expected behavior?
No exception to be risen.

### Suggested solutions
Remove support for position arguments, e.g. remove `*args` from the `QSVC/QSVR` classes. Positional arguments are deprecated in scikit-learn.

PR #220, opened 2021-09-16
Fix to issue #218

desireevl:
### Summary
Fix to issue #218 

### Details and comments
Removed *args from QSVC and QSVR classes

### DIFF
--- a/qiskit_machine_learning/algorithms/classifiers/qsvc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/qsvc.py
@@ -39,11 +39,10 @@ class QSVC(SVC):
         qsvc.predict(sample_test)
     """
 
-    def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwargs):
+    def __init__(self, quantum_kernel: Optional[QuantumKernel] = None, **kwargs):
         """
         Args:
             quantum_kernel: QuantumKernel to be used for classification.
-            *args: Variable length argument list to pass to SVC constructor.
             **kwargs: Arbitrary keyword arguments to pass to SVC constructor.
         """
 
@@ -54,7 +53,6 @@ class QSVC(SVC):
 
         super().__init__(
             kernel=self._quantum_kernel.evaluate,
-            *args,
             **kwargs,
         )
 
--- a/qiskit_machine_learning/algorithms/regressors/qsvr.py
+++ b/qiskit_machine_learning/algorithms/regressors/qsvr.py
@@ -37,17 +37,16 @@ class QSVR(SVR):
         qsvr.predict(sample_test)
     """
 
-    def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwargs):
+    def __init__(self, quantum_kernel: Optional[QuantumKernel] = None, **kwargs):
         """
         Args:
             quantum_kernel: QuantumKernel to be used for regression.
-            *args: Variable length argument list to pass to SVR constructor.
             **kwargs: Arbitrary keyword arguments to pass to SVR constructor.
         """
 
         self._quantum_kernel = quantum_kernel if quantum_kernel else QuantumKernel()
 
-        super().__init__(kernel=self._quantum_kernel.evaluate, *args, **kwargs)
+        super().__init__(kernel=self._quantum_kernel.evaluate, **kwargs)
 
     @property
     def quantum_kernel(self) -> QuantumKernel:


adekusar-drl (commented):
Looks good. 

Please add a unit test that covers the issue. To avoid actual printing, I think, the test should have something like `_ = str(qsvc)` and no exception should be risen.

Also a reno file with a short bug fix description is required. Take a look at https://github.com/Qiskit/qiskit-machine-learning/blob/main/CONTRIBUTING.md#release-notes for more details how to create a reno file.

woodsp-ibm:
Now while positional arguments is deprecated in sklearn and args should be removed, by removing it completely here, rather than deprecating it and removing it later, its a breaking change for anyone that happens to do that with today's QSVC/R. Maybe this is low risk - it would certainly need more code to deal with that. Maybe a different way along the lines to removing this might be to override repr (or str) in the interim as well to avoid using sklearns which has some problem it seems around args - I could only see repr in the sklearn source but I did not look too hard.

I see in the test it only tested against extra named arguments (kwargs) so no tests are impacted.

adekusar-drl:
@woodsp-ibm I don't fully understand your suggestion. We can deprecate them manually (I don't think `@deprecate_arguments` can handle this), e.g. we still have `*args` in `QSVC`, but if they are not empty we print a deprecation warning, ignore them and don't pass them to the super class. Is it a preferable solution here? 

I'd like to keep as less code as possible and not override `str` or `repr` from sklearn.

woodsp-ibm:
It maybe that the deprecation of args might need to be more manual and not use the decorator we have. Use of args can of course be detected - so we can say its deprecated and remove later. We could also call super init and include args if it was not empty otherwise call it without any args. That would allow any code that did have args to carry on working as it did - but I imagine in that case if one does a str/repr on it, it would still fail. One could in the interim override repr to return super repr but catch the error there and give back some other repr as a workaround which would be removed once the args are removed. How much is worth doing I guess depends - we only had the one error report so maybe its low risk as no-one really used args - indeed the tests we have use kwargs.

adekusar-drl:
I don't think we have to pass positional arguments to `SVC`, they don't use them and this leads to the issue we have right now. We can just raise a warning and discard them.

woodsp-ibm:
> I don't think we have to pass positional arguments to SVC, they don't use them

As far as I could tell from briefly looking at sklearn they have a deprecation decorator and end up converting any positional args to named args. Since it seems supported I do not know why the repr/str ends up failing when passed args but either way we should deprecate our exposure of args etc.

adekusar-drl:
@woodsp-ibm as agreed today: we deprecate `*args`, raise a deprecation warning, if arguments are not empty we pass them to the super-class and if a user prints this instance and gets an error, then they must cope with the problem on their own. I certainly don't want to override additional methods.

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -47,16 +48,23 @@ def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwar
             **kwargs: Arbitrary keyword arguments to pass to SVC constructor.
         """
 
+        if (len(args)) != 0:
+            msg = (
+                f"Positional arguments ({args}) are deprecated as of version 0.3.0 and "
+                f"will be removed no sooner than 3 months after the release. Instead use "
+                f"keyword arguments. Positional arguments are ignored and not passed to the superclass."
```
woodsp-ibm, line 55: `Positional arguments are ignored and not passed to the superclass.` This is not the case correct so should be deleted from the message

qiskit_machine_learning/algorithms/regressors/qsvr.py
```diff
@@ -45,9 +46,20 @@ def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwar
             **kwargs: Arbitrary keyword arguments to pass to SVR constructor.
         """
 
+        if (len(args)) != 0:
+            msg = (
+                f"Positional arguments ({args}) are deprecated as of version 0.3.0 and "
+                f"will be removed no sooner than 3 months after the release. Instead use "
+                f"keyword arguments. Positional arguments are ignored and not passed to the superclass."
```
woodsp-ibm, line 53: Similar comment for ` Positional arguments are ignored and not passed to the superclass.`

releasenotes/notes/deprecated-positional-arguments-cb78dd22eb0894e0.yaml
```diff
@@ -0,0 +1,8 @@
+---
+issues:
+  - |
+    If positional arguments were passed into QSVR or QSVC and these classes were printed, an 	 exception was raised.
```
woodsp-ibm, line 4: Presumably this is to note that if you do pass positional arguments then and use of repr/str conversion on the instance will error. This is still the case right? I.e the issues: section is to note issues that are in the release - so for me the sentence could be changed from past to present tense. If the tense was more related to this is what was fixed then it should go in the fixes: section.

releasenotes/notes/deprecated-positional-arguments-cb78dd22eb0894e0.yaml
```diff
@@ -0,0 +1,8 @@
...
+fixes:
+  - |
+    Positional arguments in QSVR and QSVC are deprecated.
```
woodsp-ibm, line 7: This should be a `deprecations` section since it notes what is deprecated really not any fixes.

test/algorithms/classifiers/test_qsvc.py
```diff
@@ -101,6 +101,19 @@ def test_qsvc_parameters(self):
 
         self.assertEqual(score, 0.5)
 
+    def test_args_warning(self):
+        """Test QSVC deprecation warning appears when *args passed in"""
+        qsvc = QSVC(1)
+
+        with self.assertWarns(DeprecationWarning):
+            print(qsvc)
```
woodsp-ibm, line 109: Can we do say `_ = str(qsvc)` in al cases so we call the string conversion that print would be doing - but unlike print it will not be printing anything to the console/log.

test/algorithms/classifiers/test_qsvc.py
```diff
@@ -101,6 +101,19 @@ def test_qsvc_parameters(self):
 
         self.assertEqual(score, 0.5)
 
+    def test_args_warning(self):
+        """Test QSVC deprecation warning appears when *args passed in"""
+        qsvc = QSVC(1)
+
+        with self.assertWarns(DeprecationWarning):
```
woodsp-ibm, line 108: Isn't the deprecation warning issued when the class in constructed? Now it maybe that since we are passing in args that it is failing in this line like the issue reported - not sure whether we want to test it failing. For sure passing no args it should convert to str, so would print, so checking that the issue no longer exists in that case seems fine.

desireevl:
@woodsp-ibm, I believe I have addressed all your comments above. I am seeing though that even when no args are passed into QSVR() and QSVC(), it is unable to print the class still (so the test function fails).

woodsp-ibm:
I was looking at this and see a couple of hours ago scikit-learn 1.0 was released and the deprecated args that SVC had allowed seems to have been removed. 

As far as I could tell so far (with the 0.24.2) just having the args in the QSVC init signature is what ends up making it fail. I was starting to look at perhaps using a similar (or the same) deprecation decorator for the args, to see if that would work. But I do not know now given 1.0 is out - needs some further investigation I think.

adekusar-drl:
@woodsp-ibm please take look.

PR #229, opened 2021-09-23
Re-enable lint f-string check

manoelmarques:

### DIFF
--- a/test/test_readme_sample.py
+++ b/test/test_readme_sample.py
@@ -72,7 +72,7 @@ class TestReadmeSample(QiskitMachineLearningTestCase):
         vqc.fit(training_features, training_labels)
 
         score = vqc.score(test_features, test_labels)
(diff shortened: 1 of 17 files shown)


test/test_readme_sample.py
```diff
@@ -72,7 +72,7 @@ def print(*args):
         vqc.fit(training_features, training_labels)
 
         score = vqc.score(test_features, test_labels)
-        print("Testing accuracy: {:0.2f}".format(score))
+        print(f"Testing accuracy: {score:0.2f}")
```
woodsp-ibm, line 75: As this code is part of the README file can we modify that the same way so this code that is tested is an exact copy still of the code we show there.

woodsp-ibm: My bad - I scrolled through the files and missed that it was there , duh! I see now its the second one in this list here. Thx!

woodsp-ibm (approved):

PR #230, opened 2021-09-23
Implement QuantumKernelTrainer

calebj15:
### Summary
Variational quantum algorithms generally aim to train some quantum circuit parameters using a classical optimizer. To implement this, parameters of a `QuantumCircuit` must be bound and re-bound many times during optimization; however, the `QuantumCircuit` API "forgets" the parameter after it is bound, making re-binding a cumbersome process.

The introduction of a `QuantumKernelTrainer` class simplifies this process by providing a `fit_kernel` method, which will return some optimized `QuantumKernel.feature_map` parameter values which can be used to generate an optimized kernel function. Additionally, we have overridden `sklearn.svm.SVC.fit` in `QSVC.fit`. `QSVC.fit` will now check the status of the input `QuantumKernel`, run a kernel optimization routine if appropriate, and finally call `sklearn.svm.SVC.fit()`.


qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -39,24 +42,29 @@ class QSVC(SVC):
...
+    def __init__(
+        self,
+        *args,
+        quantum_kernel: Optional[QuantumKernel] = None,
+        train_kernel: bool = False,
+        **kwargs,
+    ):
         """
         Args:
             quantum_kernel: QuantumKernel to be used for classification.
```
calebj15, line 54: Add train_kernel to args

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -66,5 +74,54 @@ def quantum_kernel(self) -> QuantumKernel:
...
+            if unbound_params:
+                # If user has not specified they wish to train the kernel,
+                # provide a warning
+                if not self.train_kernel:
+                    warnings.warn(
+                        f"""
+                        Unbound user parameters were detected in the quantum
+                        kernel feature map ({unbound_params}). If these are
+                        not bound before calling QSVC.fit, a computationally
+                        expensive kernel training procedure will precede
```
calebj15, line 99: kernel-training

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -66,5 +74,54 @@ def quantum_kernel(self) -> QuantumKernel:
...
+        return self._kernel_trainer
+
+    def fit(self, X: np.ndarray, y: np.ndarray, sample_weight=None):
+        """
+        Wrapper method for SVC.fit which optimizes the quantum kernel's
+        user parameters before fitting the SVC.
+        """
+        if self._kernel_trainer:
+            results = self._kernel_trainer.fit_kernel(X, y)
+            self.quantum_kernel.assign_user_parameters(results.optimal_parameters)
```
calebj15, line 125: This line of code should be redundant. fit_kernel updates the inner feature_map at the end of training. Consider removing.

qiskit_machine_learning/algorithms/kernel_trainers/quantum_kernel_trainer.py
```diff
@@ -0,0 +1,169 @@
...
+        qsvc.fit(X_train, y_train)
+        score = qsvc.score(X_test, y_test)
+    """
+
+    def __init__(
+        self,
+        quantum_kernel: QuantumKernel,
+        loss: Optional[Union[str, KernelLoss]] = "weighted_alignment",
+        optimizer: Optional[Optimizer] = None,
+        initial_point: Optional[np.ndarray] = None,
```
calebj15, line 53: trailing comma

qiskit_machine_learning/algorithms/kernel_trainers/quantum_kernel_trainer.py
```diff
@@ -0,0 +1,169 @@
...
+    ):
+        """
+        Args:
+            quantum_kernel: A QuantumKernel object to be tuned
+            loss: A target loss function to be used in training. Default is `weighted_alignment`,
```
calebj15, line 58: change weighted_alignment to svc_alignment

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -88,6 +91,56 @@ def gradient(self, predict: np.ndarray, target: np.ndarray) -> np.ndarray:
...
 
 
+class KernelLoss(ABC):
+    """
+    Abstract base class for computing Loss of a kernel function.
+    """
+
+    def __init__(self, *args, **kwargs) -> None:
+        # Save classical optimizer args for use in loss evaluation
+        self.loss_args = args
```
calebj15, line 101: double check this logic

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -88,6 +91,56 @@ def gradient(self, predict: np.ndarray, target: np.ndarray) -> np.ndarray:
...
+        Args:
+            user_parameters: an array of values to assign to the user params
+            kernel: An NxN matrix representing the kernel function
+                    N = # samples
+            data: An NxM matrix containing the data
+                    N = # samples, M = data dimensionality
+            labels: A length-N array containing the truth labels
+
+        Returns:
+            A loss value
```
calebj15, line 136: Say a little more

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -88,6 +91,56 @@ def gradient(self, predict: np.ndarray, target: np.ndarray) -> np.ndarray:
...
+
+        Raises:
+            QiskitMachineLearningError: shapes of predict and target do not match
```
calebj15, line 139: Ensure this is updated

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -196,3 +249,38 @@ def gradient(self, predict: np.ndarray, target: np.ndarray) -> np.ndarray:
         return target * (1.0 / (1.0 + np.exp(-predict)) - 1) + (1 - target) * (
             1.0 / (1.0 + np.exp(-predict))
         )
+
+
+class SVCAlignment(KernelLoss):
+    """
+    This class computes the weighted kernel alignment loss using SKLearn SVC class.
+    """
+
```
calebj15, line 258: whitespace

test/algorithms/classifiers/test_qsvc.py
```diff
@@ -13,6 +13,7 @@
 """ Test QSVC """
```
calebj15, line 13: Copyright

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -34,29 +37,34 @@ class QSVC(SVC):
...
+        qsvc = QSVC(quantum_kernel=quant_kernel)
         qsvc.fit(sample_train,label_train)
         qsvc.predict(sample_test)
     """
 
-    def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwargs):
+    def __init__(
+        self,
+        *args,
+        quantum_kernel: Optional[QuantumKernel] = None,
```
woodsp-ibm, line 48: From what I can see this needs to be Optional[Union[QuantumKernel, QuantumKernelTrainer]] given what I see below

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -66,5 +74,54 @@ def quantum_kernel(self) -> QuantumKernel:
     @quantum_kernel.setter
     def quantum_kernel(self, quantum_kernel: QuantumKernel):
         """Sets quantum kernel"""
-        self._quantum_kernel = quantum_kernel
+        self._kernel_trainer = None
+
+        # If no quantum kernel passed, generate a default QuantumKernel
+        if not quantum_kernel:
+            backend = Aer.get_backend("qasm_simulator")
+            self._quantum_kernel = QuantumKernel(quantum_instance=backend)
```
woodsp-ibm, line 82: Why is this needed to be added? The setter says it takes a QuantumKernel not an Optional[QuantumKernel] where None would be a valid value. The user can easily enough pass in a valid object. Also like the constructor the typehint should include the trainer now. Though since there is a separate getter for the kernel and the trainer would it make sense to have the setters separate too?

qiskit_machine_learning/algorithms/classifiers/qsvc.py
woodsp-ibm: Does it make sense to spin up a default trainer rather than say error out and require the user to provide a trainer which they can customize etc. I guess if this default is generally useful then maybe this is fine to do it implicitly and warn rather than force the user to do it?

qiskit_machine_learning/algorithms/kernel_trainers/quantum_kernel_trainer.py
```diff
@@ -0,0 +1,171 @@
...
+        # Number of parameters to tune
+        num_params = len(self.quantum_kernel.user_parameters)
+
+        # Randomly initialize our user parameters if no initial point was passed
+        if not self.initial_point:
+            self.initial_point = algorithm_globals.random.random(num_params)
+
+        # Perform kernel done
+        ignore = self.optimizer.is_initial_point_ignored
+        print(f"Is Ignored: {ignore}")
```
woodsp-ibm, line 158: Maybe printing is useful for debugging but it should not exist in final code. If it might be helpful to have this information, such that a user can enable it to be "printed" then consider using the logger facilities.

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -196,3 +250,44 @@ def gradient(self, predict: np.ndarray, target: np.ndarray) -> np.ndarray:
...
+    ) -> float:
+        print(f"user parameters: {user_parameters}")
+        # Bind training parameters
+        kernel.assign_user_parameters(user_parameters)
+
+        # Train a quantum support vector classifier
+        kernel_matrix = kernel.evaluate(data, labels)
+        print(f"kernel:\n{kernel_matrix}")
+        svc = SVC(
+            kernel=kernel.evaluate, *self.loss_args, **self.loss_kwargs
```
woodsp-ibm, line 275: In the latest sklearn release, args went from being deprecated to being completly removed such that SVC only supports named parameters.

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -66,5 +74,54 @@ def quantum_kernel(self) -> QuantumKernel:
...
+
+    @property
+    def kernel_trainer(self) -> QuantumKernelTrainer:
+        """Returns quantum kernel trainer"""
+        return self._kernel_trainer
+
+    def fit(self, X: np.ndarray, y: np.ndarray, sample_weight=None):
+        """
+        Wrapper method for SVC.fit which optimizes the quantum kernel's
+        user parameters before fitting the SVC.
```
woodsp-ibm, line 121: This should document args since this will now come out in docs. It subclasses the method, and it maybe enough to say that if a kernel trainer instance was given instead of a kernel, then the kernel will first be trained.

woodsp-ibm:
Is this something that applies to the QSVR too?

qiskit_machine_learning/algorithms/kernel_trainers/quantum_kernel_trainer.py
```diff
@@ -0,0 +1,171 @@
...
+    def initial_point(self) -> np.ndarray:
+        """Returns current initial point"""
+        return self._initial_point
+
+    @initial_point.setter
+    def initial_point(self, initial_point: np.ndarray) -> None:
+        """Sets the initial point"""
+        self._initial_point = initial_point
+
+    def fit_kernel(
```
woodsp-ibm, line 130: Would it make sense to take out the kernel from the trainer such that you pass it in on this method to fit it. I am wondering whether different trainers could be done such that QSVC was passed a kernel and a trainer as different instances so that is training was needed it could call the trainer with the kernel that was provided. This means you always pass a kernel and potentially an optional trainer.

qiskit_machine_learning/algorithms/kernel_trainers/quantum_kernel_trainer.py
calebj15: Oops this is a remnant of me debugging :)

calebj15:
> Is this something that applies to the QSVR too?

Yes, but the trainer will need to be extended to support a regression KernelLoss function, and we still need to implement at least one regression KernelLoss class

.pylintdict
```diff
@@ -10,6 +10,7 @@ autosummary
 backend
 backpropagation
 bergholm
+bincount
```
calebj15, line 13: sklearn no longer takes kwargs and args, so we need to surface their parameters in some of our modules. All of these additions to the pylint dictionary are sklearn variable names.

adekusar-drl:
@calebj15 Could you please fix the conflict? I guess you should pull the changes from the main branch to this one and replace the files with conflicts.

PR #237, opened 2021-10-04
Implement Pegasos algorithm to train an alternative QSVC

Arne-Thomsen:
### Summary
implement Pegasos algorithm to train an alternative QSVC

### Details and comments
Implementation of the Pegasos algorithm from https://home.ttic.edu/~nati/Publications/PegasosMPB.pdf for quantum support vector classifiers. The algorithm is analyzed for the quantum case in the not publicly available Master's Thesis "Comparing Quantum Neural Networks and Quantum Support Vector Machines" by Arne Thomsen.


qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,223 @@
...
+                        * (precomputed_kernel[i, j] + self._kernel_offset)
+                    )
+
+            if (self._label_dict[y[i]] * self.C / step) * value < 1:
+                self._alphas[i] = self._alphas.get(i, 0) + 1
+
+        self._fit_status = True
+
+        if self._verbose:
+            print(f"fit completed after {str(datetime.now() - t_0)[:-7]}")
```
adekusar-drl, line 213: Replace with debug.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,223 @@
...
+        super().__init__(C=C)
+        self._num_steps = num_steps
+        self._verbose = verbose
+
+        # these are the parameters being fit and are needed for prediction
+        self._fit_status = False
+        self._alphas: Optional[Dict[int, int]] = None
+        self._x_train: Optional[np.ndarray] = None
+        self._y_train: Optional[np.ndarray] = None
+        self._label_dict: Optional[Dict[int, int]] = {}
```
adekusar-drl, line 68: Could you rename `_label_dict` to something without the `_dict` suffix?

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,223 @@
...
+                    )
+                else:
+                    value += (
+                        self._alphas[j]
+                        * self._label_dict[y[j]]
+                        * (precomputed_kernel[i, j] + self._kernel_offset)
+                    )
+
+            if (self._label_dict[y[i]] * self.C / step) * value < 1:
+                self._alphas[i] = self._alphas.get(i, 0) + 1
```
adekusar-drl, line 208: If you can add a few comments on the implementation here would be nice.

adekusar-drl (commented):
Looks good, a lot of comments, but mostly minor. Please take a look at  the pre-computed kernel comment. I hope we can mimic SVC with a pre-computed kernel.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,223 @@
...
+            `Pegasos for SVM <https://home.ttic.edu/~nati/Publications/PegasosMPB.pdf>`_
+
+    """
+
+    def __init__(
+        self,
+        quantum_kernel: Optional[QuantumKernel] = None,
+        C: float = 1,
+        num_steps: int = 1000,
+        verbose: bool = False,
```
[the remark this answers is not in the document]
Arne-Thomsen: See 5c5f0947765a52b6f7295f8c7ed19142932d3a19

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
Arne-Thomsen: See 5b8382e35450af16c8b6a766a9be2becd849e947

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
Arne-Thomsen: See 0a3726d0bc75a6b4512e542066bc433cb0eefa64

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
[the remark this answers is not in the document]
adekusar-drl: Agree, there's `verbose`, but here we diverge from `scikit-learn`. Regarding `num_steps`, at least it's worth mentioning in the docstring that the algorithm iterates over all steps, no early stopping as you wrote in the comment. I don't think we should rename to `max_iter`, it has slightly different meaning, but I don't have a strong opinion here. Maybe `reps`(repetitions) could be a choice as we have such parameters in a few algorithms.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,240 @@
...
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Quantum Pegasos Support Vector Classifier"""
+
+from __future__ import annotations
```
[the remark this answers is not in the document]
adekusar-drl: Take a look here: https://docs.python.org/3/library/__future__.html

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,223 @@
...
+            NotImplementedError:
+                - when a sample_weight which is not None is passed
+        """
+        if sample_weight is not None:
+            raise NotImplementedError("all samples have to be weighed equally")
+
+        self._fit_internal(X, y)
+
+    # pylint: disable=invalid-name
+    def fit_precomputed(self, X: np.ndarray, y: np.ndarray, precomputed_kernel: np.ndarray) -> None:
```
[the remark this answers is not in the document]
Arne-Thomsen: I hope 5d0faa7b397e952100e00e078bbdaad05b0834eb settles this

adekusar-drl:
@woodsp-ibm, @stefan-woerner Could you please take a look?

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,276 @@
...
+            )
+
+        return value
+
+    @property
+    def quantum_kernel(self) -> QuantumKernel:
+        """Returns quantum kernel"""
+        return self._quantum_kernel
+
+    @quantum_kernel.setter
```
stefan-woerner, line 273: should training status (and other values) be reset when the kernel is replaced?

stefan-woerner (changes requested):

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,276 @@
...
+
+                # evaluate kernel function only for the fixed datum and the data with non zero alpha
+                kernel[(index, j)] = kernel.get(
+                    (index, j), self._quantum_kernel.evaluate(X[index], x_j)
+                )
+                kernel_value = kernel[(index, j)]
+
+            # consider a precomputed kernel
+            else:
+                kernel_value = X[index, j]
```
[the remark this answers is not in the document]
Arne-Thomsen: `X` should be correct the way this method is called. Since for training and prediction, this precomputed `X` will be include the kernel evaluations between different features. (the way the precomputed kernel is implemented is intended to resemble sklearn as closely as possible)

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,276 @@
...
+            if value > 0:
+                y[i] = self._label_pos
+            else:
+                y[i] = self._label_neg
+
+        logger.debug("prediction completed after %s", str(datetime.now() - t_0)[:-7])
+
+        return y
+
+    def _compute_weighted_kernel_sum(self, index, X, training) -> float:
```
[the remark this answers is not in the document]
Arne-Thomsen: Like in the previous case, X consists either of the features, or is a matrix of kernel values. That should probably be made more clear in the documentation

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
Arne-Thomsen: See 258cabfa6f609159cc33d5ee94f2b899c45e7a91

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,276 @@
...
+            quantum_kernel: QuantumKernel to be used for classification.
+            C: positive regularization parameter.
+            num_steps: number of steps in the Pegasos algorithm. There is no early stopping
+                criterion. The algorithm iterates over all steps.
+            precomputed: flag indicating whether a precomputed kernel is used.
+
+        Raises:
+            ValueError:
+                - if ``quantum_kernel`` is passed and ``precomputed`` is set to ``True``. To use
+                a precomputed kernel, ``quantum_kernel`` has to be of the ``None`` type.
```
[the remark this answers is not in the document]
Arne-Thomsen: See 5621aa6c6ede9592824dd4b25671441bd9827c33

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,276 @@
...
+            sample_weight: this parameter is not supported, passing a value raises an error.
+
+        Returns:
+            ``self``, a trained model.
+
+        Raises:
+            ValueError:
+                - X and/or y have the wrong shape.
+                - X and y have incompatible dimensions.
+                - y includes more than two unique labels.
```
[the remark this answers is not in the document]
Arne-Thomsen: See 87de8b0efeb6ff85d70a6dbc25921965bba8a33d

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,276 @@
...
+    def __init__(
+        self,
+        quantum_kernel: Optional[QuantumKernel] = None,
+        C: float = 1,
+        num_steps: int = 1000,
+        precomputed: bool = False,
+    ) -> None:
+        """
+        Args:
+            quantum_kernel: QuantumKernel to be used for classification.
```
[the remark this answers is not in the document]
Arne-Thomsen: I didn't include it on purpose since there's also nothing like this in the documentation of the regular QSVC and the behavior of None depends on the default behavior of the `QuantumKernel` class

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,276 @@
...
+        Raises:
+            ValueError:
+                - if ``quantum_kernel`` is passed and ``precomputed`` is set to ``True``. To use
+                a precomputed kernel, ``quantum_kernel`` has to be of the ``None`` type.
+            TypeError:
+                - if ``quantum_instance`` neither instance of ``QuantumKernel`` nor ``None``.
+        """
+        if quantum_kernel is None:
+            if precomputed:
+                self._precomputed = True
```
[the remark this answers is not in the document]
Arne-Thomsen: See c97e1c60b713fdc7275659afbc5cd1a5f9479b0a

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,276 @@
...
+        """
+        kernel: Dict[Tuple, np.ndarray] = {}
+        # this value corresponds to a sum of kernel values weighted by their labels and alphas
+        value = 0.0
+        # only loop over the non zero alphas (preliminary support vectors)
+        for j in self._alphas:
+            # perform the kernel evaluations on the fly, as intended
+            if not self._precomputed:
+                # for training
+                if training:
```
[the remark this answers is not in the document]
Arne-Thomsen: I think so, because in one case we compute the kernel on the fly and in the other we just look it up from `X`

qiskit_machine_learning/algorithms/__init__.py
```diff
@@ -104,6 +104,7 @@
     "MultiClassObjectiveFunction",
     "OneHotObjectiveFunction",
     "QSVC",
+    "PegasosQSVC",
```
woodsp-ibm, line 107: This should be added to the docstring classifier section above where QSVC is - i.e. around line 50 as that is what causes it to be included in the API ref documentation

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,285 @@
...
+    **References**
+        [1]: Shalev-Shwartz et al., Pegasos: Primal Estimated sub-GrAdient SOlver for SVM.
+            `Pegasos for SVM <https://home.ttic.edu/~nati/Publications/PegasosMPB.pdf>`_
+
+    """
+
+    def __init__(
+        self,
+        quantum_kernel: Optional[QuantumKernel] = None,
+        C: float = 1,
```
woodsp-ibm, line 51: Is this the only parameter from the parent SVC we want to expose. In the SVC we have kwargs here that allows one to pass any of the parameters that the SVC supports without listing them explicitly here. And minor comment since its a float it might be nice to put it as `1.0` rather than `1`

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,285 @@
...
+        num_steps: int = 1000,
+        precomputed: bool = False,
+    ) -> None:
+        """
+        Args:
+            quantum_kernel: QuantumKernel to be used for classification. Has to be ``None`` when
+                a precomputed kernel is used.
```
woodsp-ibm, line 58: I believe the SVC has various predefined strings and a callable for its kernel. Would it make sense to somehow combine the precomputed parameter and this to avoid having None here. E.g Optional[Union[QuantumKernel, str]]  - If None is the default one that is created, if "precomputed" then we do that, otherwise its a quantum kernel instance. Any other things are invalid. Maybe thats cleaner that having the interaction between two parameters?

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,285 @@
...
+        """
+        if quantum_kernel is None and not precomputed:
+            self._quantum_kernel = QuantumKernel()
+            
+        elif isinstance(quantum_kernel, QuantumKernel):
+            # pylint is too restrictive
+            # pylint: disable=no-else-raise
+            if precomputed:
+                raise ValueError("'quantum_kernel' has to be None to use a precomputed kernel")
+            else:
```
woodsp-ibm, line 79: Why do you want the else here. The following is what I think its after and will behave just the same
```
if precomputed:
  raise ValueError...

self._quantum_kernel = quantum_kernel
```

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,285 @@
...
+        self._label_neg: Optional[int] = None
+
+        # added to all kernel values to include an implicit bias to the hyperplane
+        self._kernel_offset = 1
+
+    # pylint: disable=invalid-name
+    def fit(
+        self, X: np.ndarray, y: np.ndarray, sample_weight: Optional[np.ndarray] = None
+    ) -> "PegasosQSVC":
+        """Implementation of the kernelized Pegasos algorithm to fit the QSVC.
```
woodsp-ibm, line 105: When I saw 'kernelized' above in the custom dictionary I wondered about it - I see that is how the algorithm is referred to in the paper. Since there is no dictionary definition, as far as I can see, presumably this means its the Pegasos algorithm adapted to work with kernels i.e kernelized- meaning converted to kernel form caramel/caramelized (sorry I did not read the paper to see).

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,285 @@
...
+
+        Args:
+            X: Train features. For a callable kernel shape (n_samples, n_features), for a precomputed
```
woodsp-ibm, line 108: A 'callable' kernel meaning when you supply a QuantumKernel versus precomputed?

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,285 @@
...
+               kernel shape (n_samples, n_samples).
+            y: shape (n_samples), train labels . Must not contain more than two unique labels.
+            sample_weight: this parameter is not supported, passing a value raises an error.
+
+        Returns:
+            ``self``, a trained model.
```
woodsp-ibm, line 114: The sklearn SVC says it returns 'Fitted estimator' - since this overrides that method does it make sense to stay with the same wording/terminology

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,285 @@
...
+
+        return self
+
+    # pylint: disable=invalid-name
+    def predict(self, X: np.ndarray) -> np.ndarray:
+        """
+        Perform classification on samples in X.
+
+        Args:
+            X: Features. For a callable kernel shape (m_samples, n_features), for a precomputed
```
woodsp-ibm, line 183: Same comment as in 'fit' about callable kernel

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,285 @@
...
+        Returns:
+            y_pred: Shape (n_samples), the predicted class labels for samples in X.
+
+        Raises:
+            QiskitMachineLearningError:
+                - predict is called before the model has been fit.
+            ValueError:
+                - Pre-computed kernel matrix has the wrong shape and/or dimension.
+        """
+        if not self._fit_status:
```
woodsp-ibm, line 197: I see sklearn in its SVCs has a `fit_status` attribute - does this integrate in someway with that - or should it. To me here it looks standalone but I am not an expert on sklearn

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,285 @@
...
+        """Helper function to compute the weighted sum over support vectors used for both training
+        and prediction with the Pegasos algorithm.
+
+        Args:
+            index: fixed index distinguishing some datum
+            X: Features
+            training: flag indicating whether the loop is used within training or prediction
+
+        Returns:
+            float: weighted sum of kernel evaluations employed in the Pegasos algorithm
```
woodsp-ibm, line 230: The return type, in the docs, gets picked out from the method signature when there are typehints. So no need to replicate here otherwise it makes maintenance harder when its duplicated. On the subject of typehints the parameters in the above signature are  missing types.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,285 @@
...
+        self._fit_status = False
+        self._alphas = None
+        self._x_train = None
+        self._n_samples = None
+        self._y_train = None
+        self._label_map = None
+        self._label_pos = None
+        self._label_neg = None
+
+        self._quantum_kernel = quantum_kernel
```
woodsp-ibm, line 285: Should we have other getter/setters for the other contructors params so for instance I can get/change `num_steps`. What about switching to a precomputed kernel - might be easier with my suggestion above rather than the separate params. Since C is a parameter of the parent  SVC class I think I can set/get that already by virtue of the parent

adekusar-drl: Added getters/setters and more tests.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: Type hints are added, return type from the docstring is removed.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: They have `fit_status_`. Unless we want to make it as a public API, I don't think we should re-use their property.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: Updated in the same manner.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: Replaced with `Fitted estimator`.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: That's right. Slightly updated the docstring.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: Sorry, I don't get it.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: This part is rewritten.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: > 
> I believe the SVC has various predefined strings and a callable for its kernel. Would it make sense to somehow combine the precomputed parameter and this to avoid having None here. E.g Optional[Union[QuantumKernel, str]] - If None is the default one that is created, if "precomputed" then we do that, otherwise its a quantum kernel instance. Any other things are invalid. Maybe thats cleaner that having the interaction between two parameters?

Personally, I don't like type mixtures like `Union[QuantumKernel, str]`. We can mimic here the classical `SVC` and if you think it is better, I can do this.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: Replaced with `1.0`.
As I see from the implementation, this is what can be re-used right now.

qiskit_machine_learning/algorithms/__init__.py
adekusar-drl: Added a reference in the API ref.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,276 @@
...
+    def fit(
+        self, X: np.ndarray, y: np.ndarray, sample_weight: Optional[np.ndarray] = None
+    ) -> "PegasosQSVC":
+        """Implementation of the kernelized Pegasos algorithm to fit the QSVC.
+
+        Args:
+            X: Train features. For a callable kernel shape (n_samples, n_features), for a precomputed
+               kernel shape (n_samples, n_samples).
+            y: shape (n_samples), train labels.
+            sample_weight: this parameter is not supported, passing a value raises an error.
```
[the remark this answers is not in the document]
adekusar-drl: Let's put this aside at least in this version. Maybe @Arne-Thomsen has the correct answer.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,276 @@
...
+        num_steps: int = 1000,
+        precomputed: bool = False,
+    ) -> None:
+        """
+        Args:
+            quantum_kernel: QuantumKernel to be used for classification.
+            C: positive regularization parameter.
+            num_steps: number of steps in the Pegasos algorithm. There is no early stopping
+                criterion. The algorithm iterates over all steps.
+            precomputed: flag indicating whether a precomputed kernel is used.
```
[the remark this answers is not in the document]
adekusar-drl: This is a boolean flag, if it is set to `True` we expect a precomputed kernel instead of `X`. Updated the docstring.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -0,0 +1,276 @@
...
+        self,
+        quantum_kernel: Optional[QuantumKernel] = None,
+        C: float = 1,
+        num_steps: int = 1000,
+        precomputed: bool = False,
+    ) -> None:
+        """
+        Args:
+            quantum_kernel: QuantumKernel to be used for classification.
+            C: positive regularization parameter.
```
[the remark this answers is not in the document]
adekusar-drl: @Arne-Thomsen could you please update the docstring with more details on the `C` parameter?

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
stefan-woerner: excatly, the standard version of the algorithm does not use kernels, and the "kernelized version" is extended to use kernels.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
woodsp-ibm: I guess my comment was more that this sub-class seems to have its own status and SVC as a class had its. Since this is a derived class from SVC I would expect all its methods etc to be applicable - they may be overridden of course to be more specialized. So if I call fit should I not be able to look at fit_status_ and get the status as I do if I use the SVC directly - or is that aspect of SVC broken here. Should this then be extending SVC - I see there is a classifermixin but maybe thats not enough.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
woodsp-ibm: I guess it was partly a reaction to the wording on quantum_kernel specifically "Has to be None when...." but since that's the default its not as bad I guess as it sounds. I assume we want the quantum kernel param here to be just like the QSVC where None means we spin one up internally hence needing the extra flag (precomputed) to negate that. The one thing that is not so great about the suggestion I made is that it needs this magic string "precomputed" to be given - versus the boolean , but it did resemble the kernel parameter of SVC which is some union of string, callable. I don't have a strong opinion one way or the other - guess I was mainly trying to have something that was easier where it was either/or rather than don't do this if you do that.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
woodsp-ibm: Hopefully people will know what this means - if you are not a native English speaker and look it up maybe its not so easy. `Pegasos algorithm adapted to work with kernels` might be easier then for them. Would it help to say this up front for people so when they see it they know what it means from the alternative wording?

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: The problem is that there's no a clean interface in `sklearn` we could derive from. `ClassifierMixin` does not define all methods like `fit/predict/score`, it has only `score`. Other methods are defined in `BaseLibSVM`. We only use method definitions here, that's why we extend `SVC`. This may not be the best way and I'm happy to change this. 

Perhaps, we should design our own base interface for classifiers/regressors that resembles `sklearn` and derive all classes from it. In this case we would have full compatibility in the interface, but things `isinstance` would not work.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: Well, in the paper they also use "The Kernelized Pegasos Algorithm". I have not studied the paper in details, looks good to me as a non-native speaker :)

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: I'd keep as it it now if there's no a string objection. May be as an argument to keep it is that in `QSVC/QSVR` we don't accept a string as `quantum_kernel`, but only a kernel instance or ``None``.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
woodsp-ibm: Yes, that was the only place I saw it i.e. in the paper. Also keep in mind the docs are translated and not being able to find kernelized in a dictionary may be problematic. I guess we'll see.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
woodsp-ibm: It seems that in extending SVC you want to be able to use this, as a specialized version of an SVC, in contexts where SVC can be used, which makes a great deal of sense. I.e. has the same public API/interface and in fact some of the methods etc only exist in the parent. Maybe there are other aspects that will mean that a simple swap does not work in practice depending on the code in question though. For this example I happened to notice, around fit_status, if you checked that in the code you had been using with the SVC then things are not going to work as expected with this algorithm. In the case of fit_status it might just be simple to use the same instance var - the only place I see it being set is after the call to libsvm.fit So maybe linking to this is not so bad in practice though of course things might change with it in the future. Now this is the aspect I noticed and commented on - I saw other attributes etc but I did not look further. Perhaps mostly in practice these are aspects that are infrequently used and most people would be able to use this algorithm, how they were using an SVC in sklearn, without noticing such. Its not like things can't be changed - we can certainly see how this goes and whether we get any feedback in these regards.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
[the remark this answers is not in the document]
Arne-Thomsen:
```suggestion
            C: Positive regularization parameter. The strength of the regularization is inversely proportional to C. Must be strictly positive. The penalty is a l1 penalty.

```

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: Added a line to the class docstring to reflect "adapted to work with kernels". Rephased the `fit` method docstring.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: OK, replaced `_fit_status` with `fit_status_`, I doubt this is a good choice.

stefan-woerner (dismissed):

docs/tutorials/07_pegasos_qsvc.ipynb
```diff
@@ -0,0 +1,338 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "opened-florist",
+   "metadata": {},
+   "source": [
+    "# Pegasos Quantum Support Vector Classifier\n",
+    "\n",
+    "There's another SVM based algorithm that benefits from the quantum kernel method. Here, we introduce an implementation of a another classification algorithm, which is an alternative version to the QSVC shown above. This classification algorithm implements the Pegasos algorithm from the paper \"Pegasos: Primal Estimated sub-GrAdient SOlver for SVM\" by Shalev-Shwartz et al., see: https://home.ttic.edu/~nati/Publications/PegasosMPB.pdf.\n",
+    "\n",
+    "This algorithm:\n",
```
stefan-woerner, line 12: I would remove this paragraph and integrate it into the next one (see suggestions below)

docs/tutorials/07_pegasos_qsvc.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "\n",
+    "- Allows application of the kernel trick that we have seen in the previous sections\n",
+    "- Alternative to the dual optimization from the `scikit-learn` package.\n",
+    "- Finds a $\\delta$ accurate solution.\n",
+    "\n",
+    "This algorithm yields a training complexity that is independent of the size of the training set, such that the `PegasosQSVC` is expected to train faster than QSVC for sufficiently large training sets:\n",
```
stefan-woerner, line 18:
```suggestion
    "This algorithm is an alternative to the dual optimization from the `scikit-learn` package, and yields a training complexity that is independent of the size of the training set. Thus, the `PegasosQSVC` is expected to train faster than QSVC for sufficiently large training sets:\n",
```

docs/tutorials/07_pegasos_qsvc.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "\n",
+    "|     | Dual optimization | Pegasos algorithm|\n",
+    "| --- | --- | --- |\n",
+    "| Ideal kernel evaluations (classical) | $\\mathcal{O}(M^2)$ | $\\mathcal{O}(\\frac{M}{\\delta})$ |\n",
+    "| Ideal quantum circuit evaluations | $\\mathcal{O}(\\frac{M^{4.67}}{\\epsilon^2})$ | $\\mathcal{O}(\\text{min}[\\frac{M^2}{\\delta^3}, \\frac{1}{\\delta^5}])$ |\n",
+    "\n",
```
stefan-woerner, line 24:
```suggestion
    "\n where:",
```

docs/tutorials/07_pegasos_qsvc.ipynb
```diff
@@ -0,0 +1,338 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "central-poverty",
+   "metadata": {},
+   "source": [
+    "We have two features in the dataset, so we set a number of qubits to the number of features in the dataset.\n",
+    "\n",
+    "Then we set $\\tau$ to the number of steps performed during the training procedure. Please note that, there is no early stopping criterion in the algorithm. The algorithm iterates over all $\\tau$ steps.\n",
+    "\n",
+    "And the last one is the hyperparameter $C$, the larger $C$ is, the faster the algorithm runs."
```
stefan-woerner, line 88: This needs more explanation, otherwise, why wouldn't you always set `C` to infinity?

docs/tutorials/07_pegasos_qsvc.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "\n",
+    "qkernel = QuantumKernel(feature_map=feature_map, quantum_instance=pegasos_backend)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "attractive-stationery",
+   "metadata": {},
+   "source": [
+    "The implementation `PegasosQSVC` is compatible with thee `scikit-learn` interfaces and has a pretty standard way of training a model. In the constructor we pass parameters of the algorithm, in this case there are a regularization hyper-parameter $C$ and a number of steps.\n",
```
stefan-woerner, line 150:
```suggestion
    "The implementation `PegasosQSVC` is compatible with the `scikit-learn` interfaces and has a pretty standard way of training a model. In the constructor we pass parameters of the algorithm, in this case there are a regularization hyper-parameter $C$ and a number of steps.\n",
```

docs/tutorials/07_pegasos_qsvc.ipynb
```diff
@@ -0,0 +1,338 @@
...
+    "pegasos_score = pegasos_qsvc.score(test_features, test_labels)\n",
+    "print(f'PegasosQSVC classification test score: {pegasos_score}')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "sustainable-empire",
+   "metadata": {},
+   "source": [
+    "For visualization purposes we create a mesh grid of a predefined step than spans our minimum and maximum values we applied in MinMaxScaler. We also add some margin to the grid for better representation of the training and test samples."
```
stefan-woerner, line 189:
```suggestion
    "For visualization purposes we create a mesh grid of a predefined step that spans our minimum and maximum values we applied in MinMaxScaler. We also add some margin to the grid for better representation of the training and test samples."
```

stefan-woerner (changes requested):

docs/tutorials/07_pegasos_qsvc.ipynb
adekusar-drl: Slightly re-arrange the text.

docs/tutorials/07_pegasos_qsvc.ipynb
adekusar-drl: This is a good question, I noticed `C` is large compared to classical SVM, but did not have chance to review the algorithm in details. This is taken as is from the @Arne-Thomsen 's notebook.
@Arne-Thomsen Can you clarify a bit here?

adekusar-drl: OK, I replaced this sentence with the docstring. Now it is: `And the last one is the hyperparameter 𝐶. Positive regularization parameter. The strength of the regularization is inversely proportional to 𝐶. Must be strictly positive. The penalty is a L1 penalty.`

stefan-woerner (approved):

PR #243, opened 2021-10-12
Transpile circuit in CircuitQNN for better performance

adekusar-drl:
### Summary
Currently every time when `CircuitQNN` executes a circuit it is transpiled internally by `QuantumInstance`. Thus, overall fir time of a model based on `CircuitQNN` is larger than it could be. In this PR we introduce an additional transpilation step in `CircuitQNN` that is done when a quantum instance is set. The circuit is transpiled and saved for further re-use. So, every time when the circuit is executed it is already transpiled and overall time of the forward pass is reduced.

### Details and comments
A simple model based on the iris dataset shows a decrease in fit time from 39.5 to 25 seconds on 5 iterations.

This enhancement does not work with `RawFeatureVector`. This feature map can't be transpiled until all parameters are bound, hence it can not be transpiled in advance. This is a known problem.

### DIFF
--- a/.pylintdict
+++ b/.pylintdict
@@ -165,6 +165,7 @@ toctree
 todo
 traceback
 transpilation
+transpile
 transpiled
 uncompiled
 unitaries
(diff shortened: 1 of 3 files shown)


woodsp-ibm (approved):
LGTM

PR #244, opened 2021-10-18
Add QuantumKernelTrainer and KernelLoss classes

### LINKED ISSUE #227: Implement a quantum kernel trainer class

calebj15:
### Summary
Closes #227 
Variational quantum algorithms generally aim to train some quantum circuit parameters using a classical optimizer. To implement this, parameters of a `QuantumCircuit` must be bound and re-bound many times during optimization; however, the `QuantumCircuit` API "forgets" the parameter after it is bound, making consecutive binds of a `QuantumCircuit` parameter a cumbersome process.

The introduction of a `QuantumKernelTrainer (QKT)` class simplifies this process by providing a `fit_kernel` method, which will return some optimized `QuantumKernel.feature_map` parameter values which can be used to generate an optimized kernel operator.

Changes from this PR:

* Addition of a  `qiskit_machine_learning.kernels.algorithms.QuantumKernelTrainer (QKT)` class which facilitates optimizing a set of a parameters within a `QuantumKernel` feature map with regard to a given dataset.

* Addition of a new `qiskit_machine_learning.utils.loss_functions` base class, `KernelLoss`. `KernelLoss` functions take a set of circuit parameters to optimize, a `QuantumKernel` object, and a labeled dataset as input and provide a loss value as output.

* Addition of a new `KernelLoss` child class, `SVCLoss`.


test/algorithms/classifiers/test_qsvc.py
```diff
@@ -114,6 +116,39 @@ def test_with_kernel_parameter(self):
         with self.assertWarns(QiskitMachineLearningWarning):
             QSVC(kernel=1)
 
+    def test_unbound_user_params(self):
```
adekusar-drl, line 119: The test name is slightly misleading. Is it a test that tests the whole process? If not, then there should be such a test.

adekusar-drl (commented):
Thank! A couple of questions:
- Are there any papers to reference in the documentation?
- Does QKA works for QSVR?

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -64,24 +67,65 @@ def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwar
...
             del kwargs["kernel"]
 
-        self._quantum_kernel = quantum_kernel if quantum_kernel else QuantumKernel()
-
         if "random_state" not in kwargs:
             kwargs["random_state"] = algorithm_globals.random_seed
 
-        super().__init__(kernel=self._quantum_kernel.evaluate, *args, **kwargs)
+        # Class Fields
+        self._quantum_kernel = quantum_kernel or QuantumKernel()
```
woodsp-ibm, line 74: Ideally when checking against None it should be explicitly done since objects can override __bool__ and return True/False for non None entities.

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -33,27 +38,25 @@ class QSVC(SVC):
...
+        kernel_trainer: Optional[QuantumKernelTrainer] = None,
+        **kwargs,
+    ):
+        r"""
         Args:
-            quantum_kernel: QuantumKernel to be used for classification.
-            *args: Variable length argument list to pass to SVC constructor.
-            **kwargs: Arbitrary keyword arguments to pass to SVC constructor.
+            quantum_kernel:``QuantumKernel`` to be used for classification
+            kernel_trainer: ``QuantumKernelTrainer`` to be used for kernel optimization
```
woodsp-ibm, line 57: Can we elaborate here on the default behavior in the docstrings as I think its kinda different. For quantum_kernel, if one is not provided the algorithm spins up a default internally, and hence the getter will always get a valid instance despite one not being passed here. For the trainer, if nothing is passed, then there is simply no training done right. We don't have any trainer instance that does no training as some sort of default.

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -64,24 +67,65 @@ def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwar
...
+    ) -> SVC:
+        # Ensure a backend was specified
+        if self._quantum_kernel.quantum_instance is None:
+            raise QiskitMachineLearningError(
+                """
+            Error fitting QSVC. No quantum instance was specified.
+            """
+            )
+
+        # Ensure there are no unbound user parameters
```
woodsp-ibm, line 113: Reading the message in the exception answers the question why and in that regard this comment could be improved to be more useful to someone new to the code reading it.

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -64,24 +67,65 @@ def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwar
...
+        unbound_user_params = self._quantum_kernel.get_unbound_user_parameters()
+        if (len(unbound_user_params) > 0) and (self._kernel_trainer is None):
+            raise QiskitMachineLearningError(
+                f"""
+            Cannot fit QSVC while feature map has unbound user parameters ({unbound_user_params}).
+            """
+            )
+
+        # Conduct kernel optimization, if required
+        if self._kernel_trainer:
```
woodsp-ibm, line 123: if self._kernel_trainer is not None. See earlier comments about testing explicitly to None when that is the intent rather than this boolean test.

qiskit_machine_learning/kernels/algorithms/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+Quantum Kernel Algorithms (:mod:`qiskit_machine_learning.kernels.algorithms`)
```
woodsp-ibm, line 14: Quantum Kernel Algorithms - QKA? or is that Quantum Kernel Alignment. If the latter maybe this is Quantum Kernel algorithms (lowercase) to avoid confusion. In relation to the runtime do we have a local QKA , whatever that is - or is that part of the QSVC by virtue of adding the trainer?

adekusar-drl:
@calebj15 Are there any updates on this PR?

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -196,3 +261,56 @@ def gradient(self, predict: np.ndarray, target: np.ndarray) -> np.ndarray:
...
         )
+
+
+class SVCLoss(KernelLoss):
+    r"""
+    .. math::
+
+        \text{This class provides a kernel loss function for classification tasks by
+        fitting an ``SVC`` model from sklearn. Given training samples, x_{i}, with binary
+        labels, y_{i}, and a kernel, K_{θ}, parameterized by values theta, the loss is defined as}
```
jenglick, line 272: can we use `θ` instead of `theta`? Here and below

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -196,3 +261,56 @@ def gradient(self, predict: np.ndarray, target: np.ndarray) -> np.ndarray:
...
+
+        SVCLoss = \sum_{i} a_i - 0.5 \sum_{i,j} a_i a_j y_{i} y_{j} K_θ(x_i, x_j)
+
+        \text{where a_i are the optimal Lagrange multipliers found by solving the standard
+        SVM quadratic program. Note that the hyper-parameter C for the soft-margin penalty can
+        be specified through the keyword args.}
+
+    Minimizing this loss over the parameters theta of the kernel is equivalent to maximizing a
+    weighted kernel alignment, which in turn yields the smallest upper bound to the SVM
+    generalization error for a given set of parameters.
```
jenglick, line 282: "for a given set of parameters" --> "for a given parametrization"

jenglick (commented):
@calebj15 added a couple minor comments to the SVCLoss doc string

qiskit_machine_learning/kernels/quantum_kernel_trainer.py
```diff
@@ -0,0 +1,228 @@
...
+        optimizer = ...
+        initial_point = ...
+
+        qk_trainer = QuantumKernelTrainer(
+                                        loss=loss_func,
+                                        optimizer=optimizer,
+                                        initial_point=initial_point,
+                                        )
+
+        qsvc = QSVC(quantum_kernel=quant_kernel, kernel_trainer=qk_trainer)
```
stefan-woerner, line 71: this seems not up to date.

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -88,6 +95,64 @@ def gradient(self, predict: np.ndarray, target: np.ndarray) -> np.ndarray:
         raise NotImplementedError
 
 
+class KernelLoss(ABC):
```
stefan-woerner, line 98: I would put kernel loss in a separate file `kernel_loss_functions.py`

qiskit_machine_learning/utils/loss_functions/loss_functions.py
```diff
@@ -196,3 +261,56 @@ def gradient(self, predict: np.ndarray, target: np.ndarray) -> np.ndarray:
...
+        svc.fit(data, labels)
+
+        # Get dual coefficients
+        dual_coefs = svc.dual_coef_[0]
+
+        # Get support vectors
+        support_vecs = svc.support_
+
+        # Get estimated kernel matrix
+        kmatrix = quantum_kernel.evaluate(np.array(data))[support_vecs, :][:, support_vecs]
```
stefan-woerner, line 311: more efficient if you restrict the data and then evaluate the kernel than if you first evaluate the full kernel matrix and then drop what's not needed.

stefan-woerner (changes requested):

qiskit_machine_learning/utils/loss_functions/loss_functions.py
stefan-woerner: we should add an option to tell the kernel to cache results. Right now the kernel matrix is evaluated twice per iteration.

calebj15: We have addressed this by passing the kernel matrix into the SVC as a 'precomputed' kernel, which should prevent it from being evaluated twice.

We have some ideas about caching, but they are more involved and should probably go in another PR. Resolving this; will follow up about caching.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -172,29 +187,91 @@ def assign_user_parameters(
...
-            self._user_param_binds = param_binds
+            values = {p: values[i] for i, p in enumerate(self._user_parameters)}
         else:
-            self._user_param_binds.update(param_binds)
+            if not isinstance(values, dict):
+                raise ValueError(
+                    f"""
+                'values' must be of type Dict or Sequence.
+                Type {type(values)} is not supported.
+                """
```
garrison, line 197: I suggest indenting these three lines one additional level, to match the preceding line.

releasenotes/notes/add-quantum-kernel-trainer-a31bfa1c91a05e32.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Addition of a ``QuantumKernelTrainer`` object which may be used by kernel-based machine learning algorithms to perform
+    optimization of some ``QuantumKernel`` parameters before training the model. Addition of a new base class ,
```
garrison, line 5: The space after the word "class" is unnecessary.

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
```diff
@@ -0,0 +1,225 @@
...
+        self._optimizer = optimizer
+        self._initial_point = initial_point
+        self._loss = None
+
+        # Setters
+        self.loss = loss
+
+    @property
+    def quantum_kernel(self) -> "QuantumKernel":
+        """Returns the quantum kernel object."""
```
garrison, line 108: I would suggest replacing the word "Returns" with "Return" here.  This is based on the following guideline from [PEP 257](https://www.python.org/dev/peps/pep-0257/):

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

This suggestion could also be applied to many other docstrings throughout this file.

qiskit_machine_learning/utils/loss_functions/kernel_loss_functions.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def get_variational_callable(
+        self, quantum_kernel: QuantumKernel, data: Iterable, labels: Iterable
+    ) -> Callable[[Sequence[float]], float]:
+        """
+        Return a callable variational loss function given some inputs. The sole input to the
+        callable should be an array of numerical feature map parameter values, and the output should
+        be a numerical loss value.
+        """
+        raise NotImplementedError
```
adekusar-drl, line 54: Don't really get why this one is required.

qiskit_machine_learning/utils/loss_functions/kernel_loss_functions.py
```diff
@@ -0,0 +1,143 @@
...
+
+    Minimizing this loss over the parameters, θ, of the kernel is equivalent to maximizing a
+    weighted kernel alignment, which in turn yields the smallest upper bound to the SVM
+    generalization error for a given parametrization.
+
+    See https://arxiv.org/abs/2105.03406 for further details.
+    """
+
+    def __init__(self, **kwargs):
+        self.kwargs = kwargs
```
adekusar-drl, line 102: Please document the constructor, especially why `**kwargs` are here.

qiskit_machine_learning/utils/loss_functions/kernel_loss_functions.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def get_variational_callable(
+        self, quantum_kernel: QuantumKernel, data: Iterable, labels: Iterable
+    ) -> Callable[[Sequence[float]], float]:
+        return partial(
+            self.evaluate,
+            quantum_kernel=quantum_kernel,
+            data=data,
+            labels=labels,
+        )
```
adekusar-drl, line 112: Can we get rid of that?

qiskit_machine_learning/utils/loss_functions/kernel_loss_functions.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def evaluate(
+        self,
+        parameter_values: Sequence[float],
+        quantum_kernel: "QuantumKernel",
```
adekusar-drl, line 117: Type hint in quotes

test/kernels/algorithms/test_qkernel_trainer.py
```diff
@@ -0,0 +1,111 @@
...
+    data_block = ZZFeatureMap(2)
+    trainable_block = ZZFeatureMap(2)
+    user_parameters = trainable_block.parameters
+
+    for i, _ in enumerate(user_parameters):
+        user_parameters[i]._name = f"θ[{i}]"
+
+    feature_map = data_block.compose(trainable_block).compose(data_block)
+
+    return feature_map, user_parameters
```
adekusar-drl, line 44: Do we need as a separate public function?

test/kernels/algorithms/test_qkernel_trainer.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def test_qkt(self):
+        """Test QuantumKernelTrainer"""
+        self.setUp()
+        with self.subTest("check default fit"):
+            qkt = QuantumKernelTrainer(quantum_kernel=self.quantum_kernel, optimizer=self.optimizer)
+            qkt_result = qkt.fit_kernel(self.sample_train, self.label_train)
+
+            # Ensure user parameters are bound to real values
+            self.assertFalse(qkt_result.quantum_kernel.get_unbound_user_parameters())
```
adekusar-drl, line 84: I'd rather explicitly compare with an empty list or check length is zero.

test/kernels/algorithms/test_qkernel_trainer.py
```diff
@@ -0,0 +1,111 @@
...
+        with self.subTest("check fit with params"):
+            self.setUp()
+            loss = SVCLoss(C=0.8, gamma="auto").get_variational_callable(
+                self.quantum_kernel, self.sample_train, self.label_train
+            )
+            qkt = QuantumKernelTrainer(quantum_kernel=self.quantum_kernel, loss=loss)
+            qkt_result = qkt.fit_kernel(self.sample_train, self.label_train)
+
+            # Ensure user parameters are bound to real values
+            self.assertFalse(qkt_result.quantum_kernel.get_unbound_user_parameters())
```
adekusar-drl, line 101: Same as above

test/kernels/test_qkernel.py
```diff
@@ -359,15 +359,15 @@ def test_ydim(self):
...
     """Test QuantumKernel user parameter support"""
 
     def setUp(self):
         super().setUp()
 
         # Create an arbitrary 2-qubit feature map circuit
-        circ1 = ZZFeatureMap(2)
-        circ2 = ZZFeatureMap(2)
+        circ1 = ZZFeatureMap(3)
+        circ2 = ZZFeatureMap(3)
```
adekusar-drl, line 370: I guess the comment should be updated to 3-qubits

test/kernels/algorithms/test_qkernel_trainer.py
```diff
@@ -0,0 +1,111 @@
...
+
+            # Ensure kernel training functions and is deterministic
+            qsvc = QSVC(quantum_kernel=qkt_result.quantum_kernel)
+            qsvc.fit(self.sample_train, self.label_train)
+            score = qsvc.score(self.sample_test, self.label_test)
+            self.assertAlmostEqual(score, 1.0)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
adekusar-drl, line 111: Can add a test with all possible parameters set to default values?

adekusar-drl (commented):
Sorry for the delay, quite a large amount of code. There is some room for improvement.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -122,14 +122,31 @@ def quantum_instance(
     @property
     def user_parameters(self) -> Optional[Union[ParameterVector, Sequence[Parameter]]]:
         """Return the vector of user parameters."""
-        return self._user_parameters
+        return copy.copy(self._user_parameters)
```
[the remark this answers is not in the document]
calebj15: We want to force users to use the setter, as other fields need to be updated when our user parameters change

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -172,30 +187,92 @@ def assign_user_parameters(
...
+                )
+
+        # Because QuantumKernel supports parameter rebinding, entries of the `values` dictionary must
+        # be handled differently depending on whether they represent numerical assignments or parameter
+        # reassignments. However, re-ordering the values dictionary inherently changes the expected
+        # behavior of parameter binding, as entries in the values dict do not commute with one another
+        # in general. To resolve this issue, we handle each entry of the values dict one at a time.
+        for param, bind in values.items():
+            if isinstance(bind, ParameterExpression):
+                self._unbound_feature_map.assign_parameters({param: bind}, inplace=True)
```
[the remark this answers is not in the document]
calebj15: Yes, we do it here because it is a parameter reassignment, so the unbound_feature_map needs to take in those new parameters

qiskit_machine_learning/utils/loss_functions/kernel_loss_functions.py
calebj15: Because other kernel loss functions may not need to take the same inputs in order to evaluate loss. Requiring this function be implemented, ensures that we can retrieve the loss by getting a callable that takes in only the set of parameter values and returns the loss. The inputs to this will always be the kernel, data, and labels, and any additional fields the loss needs would need to be held in the class

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
```diff
@@ -0,0 +1,225 @@
...
+
+    @loss.setter
+    def loss(self, loss: Union[str, Callable[[Sequence[float]], float]]) -> None:
+        """Set the loss."""
+        if isinstance(loss, str):
+            self._loss = loss.lower()
+            if self._loss == "svc_loss":
+                pass
+            else:
+                raise ValueError(f"Unknown loss {loss}!")
```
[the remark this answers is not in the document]
calebj15: I'm not sure I follow your if/else comment. This logic looks fine to me, what am I missing?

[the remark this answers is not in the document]
adekusar-drl: I mean `if` clause has just `pass`, while `else` throws an error. Maybe better to swap them and make `if` to throw an error and then there's no need for `else`.

test/kernels/algorithms/test_qkernel_trainer.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class TestQuantumKernelTrainer(QiskitMachineLearningTestCase):
+    """Test QuantumKernelTrainer Algorithm"""
+
+    def setUp(self):
+        super().setUp()
+        algorithm_globals.random_seed = 10598
+        self.optimizer = COBYLA(maxiter=25)
+        self.backend = Aer.get_backend("statevector_simulator")
```
adekusar-drl, line 38: Just noticed, please replace `statevector_simulator` with `aer_simulator_statevector`, otherwise there's a deprecation warning.
```suggestion
        self.backend = Aer.get_backend("aer_simulator_statevector")
```

adekusar-drl (commented):
Looks good! Any plans to add a tutorial on QKT?

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
```diff
@@ -0,0 +1,231 @@
...
+                optimizers are not recommended for training kernels.
+            initial_point: Initial point from which the optimizer will begin.
+
+        Raises:
+            ValueError: unknown loss function
+        """
+        # Class fields
+        self._quantum_kernel = quantum_kernel
+        self._initial_point = initial_point
+        self._optimizer = optimizer if optimizer else SPSA()
```
adekusar-drl, line 116: Can be simplified:
```suggestion
        self._optimizer = optimizer or SPSA()
```

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
```diff
@@ -0,0 +1,231 @@
...
+        if isinstance(loss, str):
+            loss = loss.lower()
+            if loss == "svc_loss":
+                self._loss = self._str_to_loss(loss)
+            else:
+                raise ValueError(f"Unknown loss {loss}!")
+        elif isinstance(loss, KernelLoss):
+            self._loss = loss
+        else:
+            raise ValueError(f"Unknown loss {loss}!")
```
adekusar-drl, line 128: So if I pass a callable `Callable[[Sequence[float]], float]`, I get an exception.
Could you also add a test with a custom loss to check this scenario?

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
```diff
@@ -0,0 +1,231 @@
...
+        """Set the quantum kernel."""
+        self._quantum_kernel = quantum_kernel
+
+    @property
+    def loss(self) -> KernelLoss:
+        """Return the loss object."""
+        return self._loss
+
+    @loss.setter
+    def loss(self, loss: KernelLoss) -> None:
```
adekusar-drl, line 146: Setter type should be the same as in the constructor.

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
```diff
@@ -0,0 +1,231 @@
...
+        output_kernel.assign_user_parameters(result.optimal_parameters)
+        result.quantum_kernel = output_kernel
+
+        return result
+
+    def _str_to_loss(self, loss_str: str) -> KernelLoss:
+        if loss_str == "svc_loss":
+            loss_obj = SVCLoss()
+        else:
+            raise ValueError(f"Unknown loss {loss_str}!")
```
adekusar-drl, line 229: This check is in the constructor now. Is this method now required?

adekusar-drl (commented):
Thanks! I left a few minor comments.

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
calebj15: Yes I believe we decided against this usage, so I just need to update type hint. Thanks

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
calebj15: Good catch. I think I'd rather leave it here and remove from constructor

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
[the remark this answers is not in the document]
calebj15: Oh yes, I set it up in this funky way to lay the way for future losses. Since there is only one right now, the logic is a bit strange.

test/kernels/algorithms/test_qkernel_trainer.py
```diff
@@ -0,0 +1,102 @@
...
+
+        self.quantum_kernel = QuantumKernel(
+            feature_map=self.feature_map,
+            user_parameters=self.user_parameters,
+            quantum_instance=self.backend,
+        )
+
+    def test_qkt(self):
+        """Test QuantumKernelTrainer"""
+        self.setUp()
```
adekusar-drl, line 71: `self.setUp()` is not required. Please remove.

test/kernels/algorithms/test_qkernel_trainer.py
```diff
@@ -0,0 +1,102 @@
...
+            self.assertEqual(len(qkt_result.quantum_kernel.get_unbound_user_parameters()), 0)
+
+            # Ensure kernel training functions and is deterministic
+            qsvc = QSVC(quantum_kernel=qkt_result.quantum_kernel)
+            qsvc.fit(self.sample_train, self.label_train)
+            score = qsvc.score(self.sample_test, self.label_test)
+            self.assertAlmostEqual(score, 0.5)
+
+        with self.subTest("check fit with params"):
+            self.setUp()
```
adekusar-drl, line 86: `self.setUp()` is not required. Please remove.

adekusar-drl:
@calebj15 you have a conflict in `.pylintdict`

adekusar-drl (commented):
Thanks a lot for the tutorial!I think we are converging and a few minor things are left in the code and tutorial.

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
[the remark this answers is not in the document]
woodsp-ibm: In the future, with more than one value you can do something like an if not self._loss in ["a", "b", "c"] etc so as to throw an exception if its not a recognised one, assuming its all done the same way. So that reversal and throwing an error if its not in the recognized set is still a simple one line thing and does not require a complicated if/else structure. Assuming this is all this if test is expected to do based on type.

docs/tutorials/08_quantum_kernel_trainer.ipynb
```diff
@@ -0,0 +1,457 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b6f4fd48",
+   "metadata": {},
+   "source": [
+    "### Train the Quantum Kernel\n",
+    "\n",
+    "To train the quantum kernel on the dataset (samples and labels), we call the `fit_kernel` method of `QuantumKernelTrainer`.\n",
```
adekusar-drl, line 268:
```suggestion
    "To train the quantum kernel on the dataset (samples and labels), we call the `fit` method of `QuantumKernelTrainer`.\n",
```

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
```diff
@@ -0,0 +1,233 @@
...
+        self._initial_point = initial_point
+        self._optimizer = optimizer or SPSA()
+
+        # Set up loss function
+        if isinstance(loss, str):
+            loss = loss.lower()
+            if loss == "svc_loss":
+                self._loss = self._str_to_loss(loss)
+            else:
+                raise ValueError(f"Unknown loss {loss}!")
```
adekusar-drl, line 122: I guess a string to loss conversion is done in the `_str_to_loss`, so these checks are unnecessary as you mentioned below.

adekusar-drl (commented):
Thanks, I hope I left last two comments :)

adekusar-drl:
@stefan-woerner Please review as well as you left a change request.

stefan-woerner (dismissed):
I would consider renaming `user_parameters` to `kernel_parameters` but otherwise I am good.

adekusar-drl (approved):
Thanks a lot!

PR #247, opened 2021-10-20
Added missing had_transpiled=True to quantum_kernel

cnktysz:
### Summary

Circuits are being transpiled in `quantum_kernel.py` but the necessary `had_transpiled=True` flag is not being passed when the circuits are being executed. This makes the transpilation operation re-run many times and this slows down the QSVM algorithm. 

### Details and comments

A test with the IRIS dataset is performed for 10 data samples from 2 classes is performed. Run times are given below:

Before the change: 55.0 seconds
After the change: 5.56 seconds

### DIFF
--- a/.pylintdict
+++ b/.pylintdict
@@ -167,6 +167,8 @@ traceback
 transpilation
 transpile
 transpiled
+transpiler
+transpiles
 uncompiled
 transpiling
 unitaries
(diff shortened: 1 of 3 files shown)


adekusar-drl (approved):
Looks good to me. Thanks!

PR #248, opened 2021-10-25
Fix vqc initialisation with default parameters

### LINKED ISSUE #217: VQC fails to initialize from minimal settings
This problem was brought up here https://qiskit.slack.com/archives/CB6C24TPB/p1631635618251500

The following should create a VQC instance with default feature map and ansatz.
```
vqc = VQC(num_qubits=4)
```
however it fails with
```
  File qiskit_machine_learning\algorithms\classifiers\vqc.py", line 107, in __init__
    self._circuit.compose(feature_map, inplace=True)
  File qiskit\circuit\quantumcircuit.py", line 727, in compose
    qubits = list(range(other.num_qubits))
AttributeError: 'NoneType' object has no attribute 'num_qubits'
```
It seems to be from this logic https://github.com/Qiskit/qiskit-machine-learning/blob/673140c0624ad928ec1b6da307e58221847c48da/qiskit_machine_learning/algorithms/classifiers/vqc.py#L106-L108

where the latter tries to compose `feature_map` and `ansatz`, which are directly the params from init. I think more likely it should be using `self.feature_map` and `self.ansatz` to pull in whatever these ended up as, whether supplied or internally created.

Needs some unit tests to cover the various paths whether defaulted or not.

jonvet:
### Summary
Fixes #217 where ```VQC``` couldn't be instantiated unless either ```feature_map``` or ```ansatz``` were provided. ```VQC``` is now instantiated with the default ```feature_map``` and/or ```ansatz```.

### Details and comments
Changelog: Bugfix


jonvet:
> Thanks a lot, looks good! Nice to see you are back. I left just minor comments. Could you please take a quick look if `VQR` does not have the same problem.

thanks @adekusar-drl 

I checked and `VQR` doesn't have the problem, you can instantiate it with `vqr = VQR(num_qubits=4)` without problems. 

The behaviour of the init method is very different though. In the init method of `VQC` we create a quantum circuit and then a `CircuitQNN`. But the init method of `VQR` just creates a `TwoLayerQNN` (which handles the case that was a problem here, where no ansatz or feature map were provided). Is it intentional that `VQC` and `VQR` are initialised differently or should they be made more similar?

adekusar-drl:
> I checked and `VQR` doesn't have the problem, you can instantiate it with `vqr = VQR(num_qubits=4)` without problems.
> 
> The behaviour of the init method is very different though. In the init method of `VQC` we create a quantum circuit and then a `CircuitQNN`. But the init method of `VQR` just creates a `TwoLayerQNN` (which handles the case that was a problem here, where no ansatz or feature map were provided). Is it intentional that `VQC` and `VQR` are initialised differently or should they be made more similar?

Oh, thanks for looking into this, all correct!

adekusar-drl (approved):

PR #253, opened 2021-10-30
Stable/0.2

transdigiware:


woodsp-ibm:
@transdigiware Was this PR somehow unintentional? If so it can be closed. If you are looking to contribute something it should be based off a branch done from main branch on your fork. It should also have a title, description etc etc so it can be understood - since the above has none of these things, it seems like perhaps you did not intend to do this PR but perhaps something else?

adekusar-drl:
@woodsp-ibm I suggest to close as a mistake.

adekusar-drl:
According to the github activity, it is not the first PR like this, closing.

PR #273, opened 2021-12-14
Improved efficiency of PegasosQSVM

gentinettagian:
### Summary

Suggestion to improve the efficiency of the PegasosQSVM algorithm by calling kernel.evaluate() once every iteration for all support vectors simultaneously instead of looping over all non-zero alphas.

### Details and comments

I noticed that at least for the state vector simulator this provides a significant speedup, as the state vectors can be cached in the QuantumKernel class.


qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -249,38 +249,25 @@ def _compute_weighted_kernel_sum(self, index: int, X: np.ndarray, training: bool
...
+        else:
+            x = self._x_train[support_vectors]
+        if not self._precomputed:
+            # evaluate kernel function only for the fixed datum and the support vectors
+            kernel = self._quantum_kernel.evaluate(X[index], x) + self._kernel_offset
+        else:
+            kernel = X[index, support_vectors]
+
+        y = np.array(list(map(self._label_map.get, self._y_train[support_vectors])))
+        a = np.array(list(self._alphas.values()))
```
adekusar-drl, line 268: Could you please either rename `x`, `y` and `a` to something more meaningful or add comments to the code what they mean?

adekusar-drl (commented):
Thanks for the PR. Am I right that the algorithm is still the same, you replaced the loop with a few vectorized operations?

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
gentinettagian: Thanks for the comment, I (hopefully) clarified the names. And yes, the change seems trivial. However, evaluating the kernel entries as a batch is much more efficient compared to doing it over a loop due to how QuantumKernel is optimized, so I think the change is justified.

adekusar-drl (approved):
Thanks, looks good.

PR #274, opened 2021-12-20
Enable some mypy options

manoelmarques:
### Summary
Following Qiskit/qiskit-nature#468

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -2,3 +2,6 @@
 warn_unused_configs = True
 ignore_missing_imports = True
 strict_optional = False
+no_implicit_optional = True
+warn_redundant_casts = True
+warn_unused_ignores = True
--- a/qiskit_machine_learning/algorithms/distribution_learners/qgan/qgan.py
+++ b/qiskit_machine_learning/algorithms/distribution_learners/qgan/qgan.py
@@ -109,9 +109,9 @@ class QGAN:
         if bounds is None:
             bounds_min = np.percentile(self._data, 5, axis=0)
             bounds_max = np.percentile(self._data, 95, axis=0)
-            bounds = []  # type: ignore
+            bounds = []
             for i, _ in enumerate(bounds_min):
-                bounds.append([bounds_min[i], bounds_max[i]])  # type: ignore
+                bounds.append([bounds_min[i], bounds_max[i]])
         if np.ndim(data) > 1:
             if len(bounds) != (len(num_qubits) or len(data[0])):
                 raise QiskitMachineLearningError(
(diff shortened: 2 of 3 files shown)


qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -273,13 +273,14 @@ def _compute_weighted_kernel_sum(self, index: int, X: np.ndarray, training: bool
...
+            value += cast(
+                float,
                 # alpha weights the contribution of the associated datum
                 self._alphas[j]
                 # the class membership labels have to be in {-1, +1}
                 * self._label_map[self._y_train[j]]
                 # the offset to the kernel function leads to an implicit bias term
-                * (kernel_value + self._kernel_offset)
+                * (kernel_value + self._kernel_offset),
             )
```
adekusar-drl, line 284: Not quite sure why `cast` is required. Tried locally with the latest `mypy` installed, don't see problems without this cast.

manoelmarques: it requires with those new mypy settings

adekusar-drl: I've checked out the PR, updated `mypy`, looks good without `cast`.

manoelmarques: I will try in the CI without it then because it fails locally for me.

manoelmarques: You are right. I didn't have `sparse` installed. Once I installed it, it didn't require the cast. I removed it in this PR. Thanks for double checking it.

adekusar-drl: Oh, that's very strange, there's nothing that really depends on `sparse`.

manoelmarques: Looking further when I installed `sparse`, it downgraded my numpy to 1.20.3 from 1.21.5. Reinstalling 1.21.5 caused mypy to require the cast  again. Going back to 1.20.3, it works.

adekusar-drl (approved):

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: OK, I see. So, perhaps this problem will arise again when we upgrade `numpy`.

PR #284, opened 2022-01-15
Fix VQC training on multiclass datasets

### LINKED ISSUE #251: Multi-class classification with qiskit's VQC
I am following the tutorial given in qiskit's website [Neural Network Classifier and Regressor][1]. In the first part, classification, the third section refers to qiskit's VQC library. Everything works fine with the given X and y where there are only two classes. I modified the X and y slightly to include four classes instead of two using the following lines:

    num_inputs = 2
    num_samples = 100
    
    X = 2*np.random.rand(num_samples, num_inputs) - 1
    y = np.random.choice([0,1,2,3], 100)
    
    y_one_hot = np.zeros(( num_samples, 4 ))
    for i in range(len(y)):
        y_one_hot[i, y[i]]=1

The rest of the code is untouched. VQC with ZZFeatureMap, RealAmplitudes ansatz, cross_entropy loss function and COBYLA() optimizer. 
However, when I try to fit with this new data, the classifier is only running for 5 iterations and the weights are not being changed at all. The loss or objection function's value is always calculated as "nan". 

There is a [similar question][2] I had posted about weights not being optimized with VQC, but then I thought it was because of my data or VQC's configuration. After trying this example, I realised it is clearly to do something with multiple classes and not just the classifier's configuration.
 
Please shine light on how to do multi-class classification using the qiskit's VQC library.

### TLDR: VQC not optimizing weights if multi-class classification. Code: [Colab Notebook][3]  

  [1]: https://qiskit.org/documentation/machine-learning/tutorials/02_neural_network_classifier_and_regressor.html
  [2]: https://quantumcomputing.stackexchange.com/questions/21568/qiskits-classifier-is-not-optimising-the-weights
  [3]: https://colab.research.google.com/drive/1diZB4EUqaM8Ku4b2y7zbjKnxIxK_7hEv?usp=sharing

jonvet:
Summary
Fixes #251  where VQC couldn't be trained on multiclass datasets. This is fixed by a change to the default parity function of VQC and ensuring CrossEntropyLoss cannot return nan.

Details and comments
Changelog: Bugfix

### DIFF
--- a/qiskit_machine_learning/algorithms/classifiers/vqc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/vqc.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021.
+# (C) Copyright IBM 2021, 2022.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
@@ -167,7 +167,6 @@ class VQC(NeuralNetworkClassifier):
         return super().fit(X, y)
 
     def _get_interpret(self, num_classes: int):
-
         def parity(x: int, num_classes: int = num_classes) -> int:
             return x % num_classes
 
--- a/qiskit_machine_learning/algorithms/classifiers/vqc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/vqc.py
@@ -166,8 +166,9 @@ class VQC(NeuralNetworkClassifier):
         )
         return super().fit(X, y)
 
-    def _get_interpret(self, num_classes):
-        def parity(x, num_classes=num_classes):
-            return f"{x:b}".count("1") % num_classes
+    def _get_interpret(self, num_classes: int):
+
+        def parity(x: int, num_classes: int = num_classes) -> int:
+            return x % num_classes
 
         return parity
--- a/test/algorithms/classifiers/test_vqc.py
+++ b/test/algorithms/classifiers/test_vqc.py
@@ -48,6 +48,67 @@ class TestVQC(QiskitMachineLearningTestCase):
         )
 
     @data(
+    # optimizer, quantum instance
+    ("cobyla", "statevector"),
+    ("cobyla", "qasm"),
+    ("bfgs", "statevector"),
+    ("bfgs", "qasm"),
+    (None, "statevector"),
+    (None, "qasm"),
+    )
+    def test_multiclass(self, config):
+        opt, q_i = config
+
+        if q_i == "statevector":
+            quantum_instance = self.sv_quantum_instance
+        else:
+            quantum_instance = self.qasm_quantum_instance
+
+        if opt == "bfgs":
+            optimizer = L_BFGS_B(maxiter=5)
+        elif opt == "cobyla":
+            optimizer = COBYLA(maxiter=25)
+        else:
+            optimizer = None
+
+        num_inputs = 2
+        feature_map = ZZFeatureMap(num_inputs)
+        ansatz = RealAmplitudes(num_inputs, reps=1)
+        # fix the initial point
+        initial_point = np.array([0.5] * ansatz.num_parameters)
+
+        # construct classifier - note: CrossEntropy requires eval_probabilities=True!
+        classifier = VQC(
+            feature_map=feature_map,
+            ansatz=ansatz,
+            optimizer=optimizer,
+            quantum_instance=quantum_instance,
+            initial_point=initial_point,
+        )
+
+        # construct data
+        num_samples = 5
+        num_classes = 5
+        # pylint: disable=invalid-name
+
+        X = algorithm_globals.random.random((num_samples, num_inputs))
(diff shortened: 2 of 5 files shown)


jonvet:
2 notes:
- I think the way normally to resolve numerical stability problems with CE is the log-sum-exp "trick" https://gregorygundersen.com/blog/2020/02/09/log-sum-exp/ but not sure if this will work here, because the output of the VQC forward pass are probabilities
- in the new CE loss test I added I thought about asserting that we get the same result as the torch CE loss. However we won't get the same result because we're not using `log-sum-exp`, and even if we did, our CE loss used log2 whereas in torch they use the natural logarithm. If we want to change the latter, that would probably be a separate PR?

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -166,8 +166,8 @@ def fit(self, X: np.ndarray, y: np.ndarray):  # pylint: disable=invalid-name
         )
         return super().fit(X, y)
 
-    def _get_interpret(self, num_classes):
-        def parity(x, num_classes=num_classes):
-            return f"{x:b}".count("1") % num_classes
+    def _get_interpret(self, num_classes: int):
+        def parity(x: int, num_classes: int = num_classes) -> int:
+            return x % num_classes
```
adekusar-drl, line 171: Can you give an example when this definition of the parity function is better? It is explained in the reno, but I'm not quite sure I follow the explanation.

test/algorithms/classifiers/test_vqc.py
```diff
@@ -154,6 +154,70 @@ def test_default_parameters(self, config):
...
+        X = X[X.sum(1).argsort()]
+        y_indices = (
+            np.digitize(np.arange(0, 1, 1 / num_samples), np.arange(0, 1, 1 / num_classes)) - 1
+        )
+        permutation = np.random.permutation(np.arange(num_samples))
+        X = X[permutation]
+        y_indices = y_indices[permutation]
+        y = np.zeros((num_samples, num_classes))
+        for e, index in enumerate(y_indices):
+            y[e, index] = 1
```
adekusar-drl, line 212: Would be nice  to have a comment explaining what type of dataset is constructed here.

adekusar-drl (commented):
Thanks a lot, look good. The only major comment is about the parity function and why the new definition is better.

adekusar-drl:
> 2 notes:
> 
>     * I think the way normally to resolve numerical stability problems with CE is the log-sum-exp "trick" https://gregorygundersen.com/blog/2020/02/09/log-sum-exp/ but not sure if this will work here, because the output of the VQC forward pass are probabilities
> 

Well, if this does not break the rest of the code, I'm fine.

>     * in the new CE loss test I added I thought about asserting that we get the same result as the torch CE loss. However we won't get the same result because we're not using `log-sum-exp` to resolve nans (I'm clipping zeroes to a small positive value instead) , and even if we did, our CE loss used log2 whereas in torch they use the natural logarithm. If we want to change the latter, that would probably be a separate PR?

CE loss implementation is not fully "standard", I'd say, if you change it to be compliant with `pytorch`, it would be nice.

qiskit_machine_learning/algorithms/classifiers/vqc.py
jonvet: say the number of classes is 4 and the number of output qubits is 2.

then the old parity function would give the following mapping:
```
0 -> "00" -> 0
1 -> "01" -> 1
2 -> "10" -> 1
3 -> "11" -> 2
```
so we only ever map to 0, 1 or 2, which means that later on when we construct the probabilities array (which will have 4 columns) the last column will be all 0, and a 0 gives `nan` in the CE computation.

the new parity function gives this mapping:
```
0 -> 0
1 -> 1
2 -> 2
3 -> 3
```
so that problem described above doesn't exist.

the comment in the reno was to say that even with the new parity function you can still end up with a 0.. unlikely for an entire column, but for individual cells in that column (if a class just has a low probability and we happen to never sample it). so even with this change we still need to fix the case where we take log(0)

adekusar-drl: Thanks for the explanation. I made some additional calculations and realized that the new function yields more uniform results across classes/labels, especially it is noticeable when number of classes increases. Maybe multiple classes should be handled completely differently, e.g. one vs rest or one vs one. But this is another story.

adekusar-drl (approved):
Thanks for another contribution!

jonvet:
> Thanks for another contribution!

my pleasure :) I was actually still gonna replace the log2 with natural log in the CrossEntropy calculation to make it more in line with pytorch, sorry was bit in a rush. maybe we can do this in another PR.

adekusar-drl:
> > Thanks for another contribution!
> 
> my pleasure :) I was actually still gonna replace the log2 with natural log in the CrossEntropy calculation to make it more in line with pytorch, sorry was bit in a rush. maybe we can do this in another PR.

No hurry, perhaps a multiplier `1/N` should be added as well :) In general, all loss functions can be revised.

PR #285, opened 2022-01-20
Add support for Python 3.10

manoelmarques:
### Summary
Following Qiskit/qiskit-terra#7102

### Details and comments
There isn't torch for python 3.10 yet. Since the tutorials use torch, I am keeping them running in the CI under python 3.7 and python 3.8 as before.

IMPORTANT:
Terra is backporting this. Once they release it to pypi, I will backport it as well. Do not backport before they release otherwise it will fail. The stable CI gets Terra from Pypi.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [tool.black]
 line-length = 100
-target-version = ['py36', 'py37', 'py38', 'py39']
+target-version = ['py36', 'py37', 'py38', 'py39', 'py310']
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [tool.black]
 line-length = 100
-target-version = ['py36', 'py37', 'py38', 'py39', 'py310']
+target-version = ['py37', 'py38', 'py39', 'py310']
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -1,3 +1,3 @@
 [tool.black]
 line-length = 100
-target-version = ['py36', 'py37', 'py38', 'py39']
\ No newline at end of file
+target-version = ['py36', 'py37', 'py38', 'py39', 'py310']
```
adekusar-drl: Do we drop support for 3.6? Or `py36` is still required for some purposes?

adekusar-drl (commented):
Thanks, left a minor question.

pyproject.toml
manoelmarques: It is 'black' settings for supported code. Terra kept it but I don't think it makes sense if python 3.6 is not supported in main anymore. I removed it but need to be careful to put it back in stable once this gets backported, since stable still supports 3.6

adekusar-drl (commented):
Looks good!

adekusar-drl (approved):

Issue #232, opened 2021-09-27
Clustering Algorithms using Quantum Computing


dquero:
With regards to ML, Qiskit 0.30 provides Quantum Neural Networks, Neural Network Classifier, Quantum Kernel Machine Learning, qGANs for Loading Random Distributions and Torch Connector and Hybrid QNNs.
I'm working to generate Clustering algorithms. At least k-means. Do you find it useful? Any suggestion?

adekusar-drl:
Hi, certainly, this repository is a good place for quantum machine learning algorithms. If you could implement a new algorithm using Qiskit machinery, then it would be a good addition to this package. Let me know if you have questions on Qiskit!

dquero:
I have my first version of the k-means classifier.
It does not use the euclidian distance but the cos of the angle, what is directly proportional to the distance as it can be seen in related literature.   I'm attaching the program here, I'm at your disposal for next actions.
[k-means.txt](https://github.com/Qiskit/qiskit-machine-learning/files/7455735/k-means.txt)

adekusar-drl:
Thanks a lot. I have a few questions:
1. Is the algorithm from a paper? If so, could you please share the paper then? If no, please provide a description of the algorithm.
2. What is the advantage of using quantum computers in this algorithm?
3. Are there any comparisons versus classical algorithms?
4. The code is in the text file. If you want to finally make a contribution to QML, please take a look at the contributing guideline: https://github.com/Qiskit/qiskit-machine-learning/blob/main/CONTRIBUTING.md

dquero:
The algorithm is one of the well known family of attempts to make clustering / non-supervised machine learning more efficient by estimating distances with quantum computing. There are many articles written about it, I even could learnt from it in Spanish. Some examples:
1. Esma Aïmeur, Gilles Brassard y Sébastien Gambs. “Aceleración cuántica para el aprendizaje no supervisado”. En: Machine Learning 90.2 (2013), pages. 261–287.
2. Nathan Wiebe, Ashish Kapoor y Krysta M. Svore. “Algoritmos cuánticos para métodos del vecino más cercano para el aprendizaje supervisado y no supervisado”. En: Cálculo de información cuántica 15.3 (2015), pages 316– 356.

In that case, the metric instead-of-distance used is P(1) of the auxiliary qbit in a cswap gate. P(1) is proportional to the euclidian distance in the Block sphere of 2 points, as demonstrated here:
[...]

PR #288, opened 2022-01-25
Implement QKMeans

prakharb10:
### Summary
Implementation of Quantum K Means as discussed in  #232.

### Details and comments
This is an initial implementation. Would appreciate your feedback to make any required changes.


dquero:
Hi, @prakharb10 
why is it necessary that the backend is an input argument?
didn't you have problems with the PEP-8 style about the names of some variables (q, qc, ...) or the length of the get_states function?

adekusar-drl:
Thanks for the PR. It is really looks good. 

But the algorithm itself has a crucial limitation that makes it not so interesting for the implementation. I studied the cited paper and they state on the page 34 "The measurement during swap test is causing the deconherence of input states, thus the quantum memory should contain multiple copies of input states". Since we don't have quantum memory and there are no preconditions for such thing, quantum states have to be prepared every time they are required. In the algorithm they use amplitude encoding, take a look at the formula (129) on the page 34. In the implementation you have the same approach, on the line 232 of `qkmeans.py` you have `init_state = Initialize(x)`. This is exactly what is supposed to do in the algorithm. Nothing wrong here, but on a large number of features such initialization becomes a problem. It scales exponentially, thus, if we consider other terms in the complexity of the algorithm, it leads to the same time complexity of the algorithm as the classical clustering algorithm has. Basically, quantum state preparation compensates all the possible benefits we may obtain from computing distances on a quantum computer. Considering the limitation, I don't think the implementation may be recognized as a contribution, unfortunately.

On the other hand, the implementation looks nice and I'd like to keep it as an example. The implementation can be a separate package. Also, there's an evolving process of setting up a community/ecosystem for the tools and software that depend on Qiskit. Once we have such a community, this implementation may become a member of it.

Let me know if you have questions. 
@stefan-woerner this is for your information.

adekusar-drl:
The example is good, but does not fit well into Qiskit Machine Learning. I'd be appreciated if you moved this PR/code to a dedicated repo and contributed it to the Qiskit Ecosystem, see here: https://qiskit.org/ecosystem/
Let me know if you have questions.

PR #301, opened 2022-02-01
Use bound_pass_manager in CircuitQNN

andrenmelo:
### Summary
This PR ensures `bound_pass_manager` is called when executing bound QNN circuits.

### Details and comments
N/A

### DIFF
--- a/releasenotes/notes/use-bound-pass-manager-in-circuitqnn-20c60effedb70d76.yaml
+++ b/releasenotes/notes/use-bound-pass-manager-in-circuitqnn-20c60effedb70d76.yaml
@@ -1,6 +1,6 @@
 ---
 features:
   - |
-    Fix in ``CircuitQNN`` that ensures unbound_pass_manager is called when caching
+    Added a new feature in ``CircuitQNN`` that ensures unbound_pass_manager is called when caching
     the QNN circuit and that ``bound_pass_manager`` is called when QNN parameters
     are assigned.
--- a/releasenotes/notes/use-bound-pass-manager-in-circuitqnn-20c60effedb70d76.yaml
+++ b/releasenotes/notes/use-bound-pass-manager-in-circuitqnn-20c60effedb70d76.yaml
@@ -1,0 +1,6 @@
+---
+fixes:
+  - |
+    Fix in ``CircuitQNN`` that ensures unbound_pass_manager is called when caching
+    the QNN circuit and that ``bound_pass_manager`` is called when QNN parameters
+    are assigned.
--- a/releasenotes/notes/use-bound-pass-manager-in-circuitqnn-20c60effedb70d76.yaml
+++ b/releasenotes/notes/use-bound-pass-manager-in-circuitqnn-20c60effedb70d76.yaml
@@ -1,5 +1,5 @@
 ---
-fixes:
+features:
   - |
     Fix in ``CircuitQNN`` that ensures unbound_pass_manager is called when caching
     the QNN circuit and that ``bound_pass_manager`` is called when QNN parameters
--- a/qiskit_machine_learning/algorithms/classifiers/vqc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/vqc.py
@@ -40,7 +40,6 @@ class VQC(NeuralNetworkClassifier):
         quantum_instance: QuantumInstance = None,
         initial_point: np.ndarray = None,
         callback: Optional[Callable[[np.ndarray, float], None]] = None,
-        cache_transpilation: bool = True,
     ) -> None:
         """
         Args:
@@ -118,7 +117,6 @@ class VQC(NeuralNetworkClassifier):
             output_shape=2,
             quantum_instance=quantum_instance,
             input_gradients=False,
-            cache_transpilation=cache_transpilation,
         )
 
         super().__init__(
--- a/qiskit_machine_learning/algorithms/classifiers/vqc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/vqc.py
@@ -40,6 +40,7 @@ class VQC(NeuralNetworkClassifier):
         quantum_instance: QuantumInstance = None,
         initial_point: np.ndarray = None,
         callback: Optional[Callable[[np.ndarray, float], None]] = None,
+        cache_transpilation: bool = True,
     ) -> None:
         """
         Args:
@@ -117,6 +118,7 @@ class VQC(NeuralNetworkClassifier):
             output_shape=2,
             quantum_instance=quantum_instance,
             input_gradients=False,
+            cache_transpilation=cache_transpilation,
         )
 
         super().__init__(
--- a/qiskit_machine_learning/neural_networks/circuit_qnn.py
+++ b/qiskit_machine_learning/neural_networks/circuit_qnn.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2020, 2021.
+# (C) Copyright IBM 2020, 2022.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
--- a/qiskit_machine_learning/neural_networks/circuit_qnn.py
+++ b/qiskit_machine_learning/neural_networks/circuit_qnn.py
@@ -208,7 +208,7 @@ class CircuitQNN(SamplingNeuralNetwork):
                         "determined as 2^num_qubits."
                     )
 
-                output_shape_ = (2 ** self._circuit.num_qubits,)
+                output_shape_ = (2**self._circuit.num_qubits,)
 
         # final validation
         output_shape_ = self._validate_output_shape(output_shape_)
@@ -250,8 +250,10 @@ class CircuitQNN(SamplingNeuralNetwork):
         measurements or not depending on the type of backend used.
         """
         self._set_quantum_instance(
-            quantum_instance, self._original_output_shape, self._original_interpret,
-            self._cache_transpilation
+            quantum_instance,
+            self._original_output_shape,
+            self._original_interpret,
+            self._cache_transpilation,
         )
 
     def _set_quantum_instance(
--- a/qiskit_machine_learning/neural_networks/circuit_qnn.py
+++ b/qiskit_machine_learning/neural_networks/circuit_qnn.py
@@ -62,11 +62,16 @@ class CircuitQNN(SamplingNeuralNetwork):
         gradient: Gradient = None,
         quantum_instance: Optional[Union[QuantumInstance, BaseBackend, Backend]] = None,
         input_gradients: bool = False,
-        cache_transpilation: bool = True,
     ) -> None:
         """
         Args:
             circuit: The parametrized quantum circuit that generates the samples of this network.
+                There will be an attempt to transpile this circuit and cache the transpiled circuit
+                for subsequent usages by the network. If for some reasons the circuit can't be
+                transpiled, e.g. it originates from
+                :class:`~qiskit_machine_learning.circuit.library.RawFeatureVector`, the circuit
+                will be transpiled every time it is required to be executed and only when all
+                parameters are bound. This may impact overall performance on the network.
             input_params: The parameters of the circuit corresponding to the input.
             weight_params: The parameters of the circuit corresponding to the trainable weights.
             sparse: Returns whether the output is sparse or not.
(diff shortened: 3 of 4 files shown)


woodsp-ibm:
I will note that this was done Qiskit/qiskit-terra#7217 which maybe can be utilized. It was to accommodate directly what you describe in that some passes need to happen after the parameters are bound.

adekusar-drl:
I agree with @woodsp-ibm. If you want to make use of the passes that work with parameter-bound circuits, then you should pass `bound_pass_manager` to `QuantumInstance` and transpile a circuit again once all parameters are bound.

andrenmelo:
Hi @woodsp-ibm and @adekusar-drl,

Thanks for the quick reply! Unfortunately `bound_pass_manager` does not work here. It is called from `quantum_instance.transpile`, which is [skipped](https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/utils/quantum_instance.py#L500-L512) because `has_transpiled` (or `circuit_qnn._circuit_transpiled`) is set to `True`.

adekusar-drl:
That's right. I think you can add something like this to the algorithms:
```
if quantum_instance.bound_pass_manager is not None:
    quantum_instance.transpile(circuits, pass_manager=quantum_instance.bound_pass_manager)
```
Thus, you make use of the cached transpiled parameterized circuit and you apply additional passes on the parameter-bound circuit as well.

andrenmelo:
Perhaps a simpler solution is to check whether `quantum_instance.bound_pass_manager` exists; if so then call `circuits = self._quantum_instance.transpile(circuits, transpiler=self._quantum_instance.bound_pass_manager)` [here](https://github.com/Qiskit/qiskit-machine-learning/blob/5a19a9f846fb4eb911c3c006ff7c7ccaaa646430/qiskit_machine_learning/neural_networks/circuit_qnn.py#L372).

andrenmelo:
This implementation is still problematic in the following scenario: a `bound_pass_manager` is set, but not a `pass_manager`. Then `QuantumInstance` [will use the `bound_pass_manager`](https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/utils/quantum_instance.py#L440-L445) when `CircuitQNN` tries to cache the transpiled circuit.

The problem there seems to be on `terra` side. Probably `transpile` should not use `bound_pass_manager` on unbound circuits.

adekusar-drl:
@andrenmelo I see your point, let me check if this is an intentional behaviour.

adekusar-drl:
@andrenmelo OK, the way how `bound_pass_manager` is called in `QuantumInstance` is implemented intentionally. I think this behavior should be changed and `bound_pass_manager` should be call explicitly via `quantum_instance.transpile(qc, pass_manager=bound_pass_manager)`.

andrenmelo:
> @andrenmelo OK, the way how `bound_pass_manager` is called in `QuantumInstance` is implemented intentionally. I think this behavior should be changed and `bound_pass_manager` should be call explicitly via `quantum_instance.transpile(qc, pass_manager=bound_pass_manager)`.

@adekusar-drl Sorry, I did not quite follow. Where do you mean this change shoud occur?

adekusar-drl:
> > @andrenmelo OK, the way how `bound_pass_manager` is called in `QuantumInstance` is implemented intentionally. I think this behavior should be changed and `bound_pass_manager` should be call explicitly via `quantum_instance.transpile(qc, pass_manager=bound_pass_manager)`.
> 
> @adekusar-drl Sorry, I did not quite follow. Where do you mean this change shoud occur?

Oh, sorry for the confusion. In Qiskit Terra, since `QuantumInstance` is there. But I'm not saying you should do this of course.
I see a QNN implementation like this:

```python
    # cache a transpiled parameterized circuit
    tqc = qi.transpile(qc)

    # ....
    for ...
        bound_circuit = tqc.assign_parameters(...)
        bound_circuit = qi.transpile(bound_circuit, pass_manager=qi.bound_pass_manager)[0]
        qi.execute(bound_circuit)
```
But this requires a change in `QuantumInstance` first.

adekusar-drl:
@andrenmelo Sorry for a delay, now, I hope, I know how you should proceed with this PR. I've looked through the code of `QuantumInstance` and `CircuitSampler` to understand how `bound_pass_manager` is applied. The latter was updated in the same PR where `bound_pass_manager` was introduced, see https://github.com/Qiskit/qiskit-terra/pull/7217 for more details.

So, we assume that there are two pass managers are passed to `QuantumInstance`, an unbound one and a bound. The first one should be invoked on parameterized circuits, the second only when all parameters are bound. A key assumption here is that both of them are not `None` and they are explicitly passed to `qi.transpile()`. Thus, if a user wants to make use of the `bound_pass_manager` feature, then both have to be passed to `QuantumInstance`. In general, this should not be a problem. Once we made this assumption the rest is pretty clear. A slightly modified snippet from my post above should work:

```python
# somewhere in the user code as an example
pm_config = PassManagerConfig.from_backend(FakeBrooklyn)
pm = level_1_pass_manager(pm_config)
qi = QuantumInstance(backend, pass_manager=pm, bound_pass_manager=...)
```
Then in `CircuitQNN`
```python
# cache a transpiled parameterized circuit
tqc = qi.transpile(qc, pass_manager=qi.unbound_pass_manager)

# evaluate circuits with parameters
for ...
    bound_circuit = tqc.assign_parameters(...)
    if qi.bound_pass_manager is not None:    # may be this if statement is not required
        bound_circuit = qi.transpile(bound_circuit, pass_manager=qi.bound_pass_manager)[0]
    else:
        bound_circuit = qc
    qi.execute(bound_circuit, had_transpiled=True)
```
If no pass managers are passed (both are effectively `None`) the code works as it is now.
If only `pass_manager` (unbound one) is passed then the behavior is the same.
If only `unbound_pass_manager` is passed then it is invoked for a parameterized circuit and for the bound circuit and the user is doomed. This case can be handled manually, e.g. a warning can be raised.

I hope it is clear now. Are you fine to proceed with this PR in this way? Let me know if you have questions.

andrenmelo:
Hi @adekusar-drl, thanks for the reply!

If I'm not mistaken, the current code already satisfies the requirements you laid out. The only difference is that I do not loop over bound circuits, and instead transpile and execute them in one go.

> If only pass_manager (unbound one) is passed then the behavior is the same.
If only unbound_pass_manager is passed then it is invoked for a parameterized circuit and for the bound circuit and the user is doomed. This case can be handled manually, e.g. a warning can be raised.

Here I got lost. `pass_manager` is `unbound_pass_manager`, right?

adekusar-drl:
> If I'm not mistaken, the current code already satisfies the requirements you laid out. The only difference is that I do not loop over bound circuits, and instead transpile and execute them in one go.

You do have to iterate over samples, bind parameters, transpile and then execute them in one go.

> Here I got lost. `pass_manager` is `unbound_pass_manager`, right?

Yes, you're right.

andrenmelo:
> > If I'm not mistaken, the current code already satisfies the requirements you laid out. The only difference is that I do not loop over bound circuits, and instead transpile and execute them in one go.
> 
> You do have to iterate over samples, bind parameters, transpile and then execute them in one go.
> 
> > Here I got lost. `pass_manager` is `unbound_pass_manager`, right?
> 
> Yes, you're right.

Indeed, I do a loop to bind parameters [here](https://github.com/Qiskit/qiskit-machine-learning/blob/12d40befb8e53c5defdbadffd224f01ffb9e7952/qiskit_machine_learning/neural_networks/circuit_qnn.py#L355-L362), but then a [loop to transpile is not necessary.](https://github.com/Qiskit/qiskit-machine-learning/blob/12d40befb8e53c5defdbadffd224f01ffb9e7952/qiskit_machine_learning/neural_networks/circuit_qnn.py#L364-L369)

adekusar-drl:
> Indeed, I do a loop to bind parameters [here](https://github.com/Qiskit/qiskit-machine-learning/blob/12d40befb8e53c5defdbadffd224f01ffb9e7952/qiskit_machine_learning/neural_networks/circuit_qnn.py#L355-L362), but then a [loop to transpile is not necessary.](https://github.com/Qiskit/qiskit-machine-learning/blob/12d40befb8e53c5defdbadffd224f01ffb9e7952/qiskit_machine_learning/neural_networks/circuit_qnn.py#L364-L369)

Ah, that's right, you pass a list of circuits to `transpile`.

andrenmelo:
> > Indeed, I do a loop to bind parameters [here](https://github.com/Qiskit/qiskit-machine-learning/blob/12d40befb8e53c5defdbadffd224f01ffb9e7952/qiskit_machine_learning/neural_networks/circuit_qnn.py#L355-L362), but then a [loop to transpile is not necessary.](https://github.com/Qiskit/qiskit-machine-learning/blob/12d40befb8e53c5defdbadffd224f01ffb9e7952/qiskit_machine_learning/neural_networks/circuit_qnn.py#L364-L369)
> 
> Ah, that's right, you pass a list of circuits to `transpile`.

So there can be a problem when only `bound_pass_manager` is set . I agree there should be a warning in this scenario, but shouldn't this logic be on the `terra` side?

adekusar-drl:
> > > Indeed, I do a loop to bind parameters [here](https://github.com/Qiskit/qiskit-machine-learning/blob/12d40befb8e53c5defdbadffd224f01ffb9e7952/qiskit_machine_learning/neural_networks/circuit_qnn.py#L355-L362), but then a [loop to transpile is not necessary.](https://github.com/Qiskit/qiskit-machine-learning/blob/12d40befb8e53c5defdbadffd224f01ffb9e7952/qiskit_machine_learning/neural_networks/circuit_qnn.py#L364-L369)
> > 
> > 
> > Ah, that's right, you pass a list of circuits to `transpile`.
> 
> So there can be a problem when only `bound_pass_manager` is set . I agree there should be a warning in this scenario, but shouldn't this logic be on the `terra` side?

I don't have a strong opinion here. Maybe documentation in `QuantumInstance` should be updated. May be such a warning should be raised in `QuantumInstance`. Feel free to raise an issue in `qiskit-terra` if you think this problem should be addressed out there. Having a warning here is purely optional, I'd say. `CircuitSampler` does not have it.

andrenmelo:
> Edited

I've opened an issue [here](https://github.com/Qiskit/qiskit-terra/issues/7646).

adekusar-drl (commented):
Can you please add:
1. A unit test that verifies `CircuitQNN` works correctly when both pass managers are passed to `QuantumInstance`
2. A reno file as stated in the Release Notes section of the contributing guide: https://github.com/Qiskit/qiskit-machine-learning/blob/main/CONTRIBUTING.md

andrenmelo:
> Can you please add:
> 
> 1. A unit test that verifies `CircuitQNN` works correctly when both pass managers are passed to `QuantumInstance`
> 2. A reno file as stated in the Release Notes section of the contributing guide: https://github.com/Qiskit/qiskit-machine-learning/blob/main/CONTRIBUTING.md

@adekusar-drl I took a stab at it in the latest commits. Let me know if the way I went about it makes sense.

andrenmelo:
Ah, the current unit tests require a [fix](https://github.com/Qiskit/qiskit-terra/pull/7333) on `terra` that's not yet merged. I will change that and also fix the copyright.

adekusar-drl:
@andrenmelo I'd suggest to run the tests locally before pushing the changes.

releasenotes/notes/use-bound-pass-manager-in-circuitqnn-20c60effedb70d76.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Fix in ``CircuitQNN`` that ensures unbound_pass_manager is called when caching
+    the QNN circuit and that ``bound_pass_manager`` is called when QNN parameters
+    are assigned.
```
adekusar-drl, line 6: Can you please move this to features? A rephrase like:
`Added support for bound pass manager in ``CircuitQNN``. This additional transpilation step ensures ``unbound_pass_manager`` is called when caching the QNN circuit and that ``bound_pass_manager`` is called when QNN parameters are assigned.`

adekusar-drl (commented):
Thanks for the PR and your patience. Looks good aside this PR introduces a new feature rather than a bug fix.

releasenotes/notes/use-bound-pass-manager-in-circuitqnn-20c60effedb70d76.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Fix in ``CircuitQNN`` that ensures unbound_pass_manager is called when caching
+    the QNN circuit and that ``bound_pass_manager`` is called when QNN parameters
+    are assigned.
```
adekusar-drl, line 6: Since this is a new feature, please don't use the word "fix":
```suggestion
    Added a new feature in ``CircuitQNN`` that ensures unbound_pass_manager is called when caching
    the QNN circuit and that ``bound_pass_manager`` is called when QNN parameters
    are assigned.
```

adekusar-drl (approved):
Thanks for a good job!

PR #302, opened 2022-02-03
Batch computation of gradients in QNNs

adekusar-drl:
### Summary
Currently gradients for a batch are evaluated one by one. Thus, for each sample in a dataset at least one job is submitted. It is possible to evaluate gradients for all samples in one go by passing a list of values to `CircuitSampler`. In this PR we make use of this batching feature to reduce a number of jobs required to compute gradients.

### Details and comments
There is a workaround in the code to overcome a problem(or a "feature"?) in `CircuitSampler`: https://github.com/Qiskit/qiskit-terra/issues/7608
There are a few refactorings as well to simplify the code.

### DIFF
--- a/qiskit_machine_learning/neural_networks/circuit_qnn.py
+++ b/qiskit_machine_learning/neural_networks/circuit_qnn.py
@@ -36,7 +36,7 @@ from scipy.sparse import coo_matrix
 
 from qiskit import QuantumCircuit
 from qiskit.circuit import Parameter
-from qiskit.opflow import Gradient, CircuitSampler, StateFn, OpflowError
+from qiskit.opflow import Gradient, CircuitSampler, StateFn, OpflowError, OperatorBase
 from qiskit.providers import BaseBackend, Backend
 from qiskit.utils import QuantumInstance
 from qiskit.exceptions import MissingOptionalLibraryError
@@ -107,6 +107,8 @@ class CircuitQNN(SamplingNeuralNetwork):
         self._weight_params = list(weight_params or [])
         self._input_gradients = input_gradients
         sparse = False if sampling else sparse
+        if sparse:
+            self._check_sparse_installed()
 
         # copy circuit and add measurements in case non are given
         # TODO: need to be able to handle partial measurements! (partial trace...)
@@ -145,7 +147,7 @@ class CircuitQNN(SamplingNeuralNetwork):
         self._construct_gradient_circuit()
 
     def _construct_gradient_circuit(self):
-        self._gradient_circuit: QuantumCircuit = None
+        self._gradient_circuit: OperatorBase = None
         try:
             # todo: avoid copying the circuit
             grad_circuit = self._original_circuit.copy()
@@ -206,7 +208,7 @@ class CircuitQNN(SamplingNeuralNetwork):
                         "determined as 2^num_qubits."
                     )
 
-                output_shape_ = (2 ** self._circuit.num_qubits,)
+                output_shape_ = (2**self._circuit.num_qubits,)
 
         # final validation
         output_shape_ = self._validate_output_shape(output_shape_)
@@ -336,9 +338,7 @@ class CircuitQNN(SamplingNeuralNetwork):
     def _sample(
         self, input_data: Optional[np.ndarray], weights: Optional[np.ndarray]
     ) -> np.ndarray:
-
-        if self._quantum_instance is None:
-            raise QiskitMachineLearningError("Sampling requires a quantum instance!")
+        self._check_quantum_instance("samples")
 
         if self._quantum_instance.is_statevector:
             raise QiskitMachineLearningError("Sampling does not work with statevector simulator!")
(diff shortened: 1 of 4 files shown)


adekusar-drl:
@woodsp-ibm Do we need a reno here?

woodsp-ibm:
I imagine this batching improves performance. Since such an improvement should be something end users might notice I think its worthwhile to state. Seeing such a release note might encourage someone to get the new version to have things run faster if this is what they were doing.

adekusar-drl:
@woodsp-ibm Could you please re-review since I've added a couple of new things to `OpflowQNN`?

woodsp-ibm:
A couple of comments/questions.

You refer to an issue in CircuitSampler and some form of workaround here. I imagine the intent is to remove the workaround but it may be that you want to merge this now with the workaround and refactor it once that's addressed. If that's the case were you thinking of raising a separate issue in regards of improving this once the CircuitSampler issue is remedied?

As you know there is some form of batch limit in QuantumKernel since that could produce a lot of circuits and without that in the past it ran out of memory. The Quantum Instance itself will not exceed the backend limit when creating jobs but the problem in this case was more in the creator of the circuits. This PR talks about evaluating the gradient for all samples in the dataset in one go. Is this likely to hit the same sort of problem or are things ok from that perspective?

adekusar-drl:
> A couple of comments/questions.
> 
> You refer to an issue in CircuitSampler and some form of workaround here. I imagine the intent is to remove the workaround but it may be that you want to merge this now with the workaround and refactor it once that's addressed. If that's the case were you thinking of raising a separate issue in regards of improving this once the CircuitSampler issue is remedied?
> 

Since I don't know when and how the issue will be solved, I agree that an issue to address this workaround should be risen in this repo.

> As you know there is some form of batch limit in QuantumKernel since that could produce a lot of circuits and without that in the past it ran out of memory. The Quantum Instance itself will not exceed the backend limit when creating jobs but the problem in this case was more in the creator of the circuits. This PR talks about evaluating the gradient for all samples in the dataset in one go. Is this likely to hit the same sort of problem or are things ok from that perspective?

Good point, let me check it out.

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -375,16 +375,12 @@ def _sample(
...
 
         # evaluate operator
         circuits = []
-        rows = input_data.shape[0]
-        for i in range(rows):
+        num_samples = input_data.shape[0]
+        for i in range(num_samples):
             param_values = {
                 input_param: input_data[i, j] for j, input_param in enumerate(self._input_params)
             }
```
dlasecki, line 390: Code duplication with lines 352-356.

releasenotes/notes/add-circuit-qnn-batch-grad-62be3eb762f0e82e.yaml
```diff
@@ -0,0 +1,11 @@
+---
+features:
+  - |
+    In the previous releases at the backpropagation stage of ``CircuitQNN`` and ``OpflowQNN``
+    gradients were computed for each sample in a dataset individually and then the obtained values
+    were aggregated into one output array. Thus, for each sample in a dataset at least one job was
+    submitted. Now, gradients are computed for all samples in a dataset in one go by passing a
+    list of values for a single parameter to ``CircuitSampler``. Therefore, a number of jobs
+    required for such computations is significantly reduced. This improvements may speed up
```
dlasecki, line 9:
```suggestion
    required for such computations is significantly reduced. This improvement may speed up
```

dlasecki (changes requested):

adekusar-drl:
> 
> > As you know there is some form of batch limit in QuantumKernel since that could produce a lot of circuits and without that in the past it ran out of memory. The Quantum Instance itself will not exceed the backend limit when creating jobs but the problem in this case was more in the creator of the circuits. This PR talks about evaluating the gradient for all samples in the dataset in one go. Is this likely to hit the same sort of problem or are things ok from that perspective?
> 

@woodsp-ibm I was thinking of this problem, even wrote some code, but I'm hesitant it is right time to implement an additional level of batches. First of all, in the algorithms we have `warm_start` parameter that allows to split large datasets into chucks and sequentially train models. This is exactly the same behaviour we can see in the classical algorithms, where models are trained across many epochs and in batches inside a single epoch. So, if someone has a large dataset then it is still possible to train a model if there's a memory limitation. Also, I expect some changes to happen in `QuantumCircuit` to reduce memory footprint since Terra is heading to support really large circuits. Of course, if we or somebody else hit this memory issue, we can introduce this additional layer, otherwise, I prefer not to do premature optimization. What do you think?

woodsp-ibm (approved):
> Also, I expect some changes to happen in QuantumCircuit to reduce memory footprint since Terra is heading to support really large circuits. Of course, if we or somebody else hit this memory issue, we can introduce this additional layer, otherwise, I prefer not to do premature optimization. What do you think?

Sure, since you describe other ways perhaps that this will be alleviated and other changes coming that may read on this is a beneficial way. Lets go with things as they are for now then.

qiskit_machine_learning/neural_networks/circuit_qnn.py
adekusar-drl: I'd suggest to review implementations of QNNs in general, there are a lot of things to be improved.

PR #304, opened 2022-02-04
Fix quantum kernel trainer web docs

### LINKED ISSUE #303: Quantum Kernel Trainer documentation does not populate in web API
### Environment

N/A

### What is happening?

Quantum Kernel Trainer documentation is absent from web API

N/A

### What should happen?

.rst file should be added to qiskit-machine-learning/docs/apidocs

### Any suggestions?

_No response_

calebj15:
### Summary
closes #303 

A new package was created for last release, `kernels.algorithms`, to hold the QuantumKernelTrainer class. An .rst file was not created for this new package, so the package and its contents do not show up in the web API.

### Details and comments
Added an .rst file so that the kernels.algorithms package (quantum kernel trainer) will populate in web docs.

### DIFF
--- a/docs/apidocs/qiskit_machine_learning.kernels.algorithms.rst
+++ b/docs/apidocs/qiskit_machine_learning.kernels.algorithms.rst
@@ -1,0 +1,6 @@
+﻿.. _qiskit-machine-learning-kernels-algorithms:
+
+.. automodule:: qiskit_machine_learning.kernels.algorithms
+   :no-members:
+   :no-inherited-members:
+   :no-special-members:
--- a/qiskit_machine_learning/kernels/__init__.py
+++ b/qiskit_machine_learning/kernels/__init__.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021.
+# (C) Copyright IBM 2021, 2022.
 #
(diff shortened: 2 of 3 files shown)


woodsp-ibm:
I think in the init file of the parent folder, ie in kernels, the init file there needs a link to algorithms as a sub module so it gets included. Take a look say at the init file in the root

calebj15:
> I think in the init file of the parent folder, ie in kernels, the init file there needs a link to algorithms as a sub module so it gets included. Take a look say at the init file in the root.
> 
> BTW, If you have the requirements-dev installed locally you should be able to do a `make html` in the root of the project which will build out things there for you - output ends up in a _build folder under docs

Thanks @woodsp-ibm. I updated the kernels init file to point to its submodule, kernels.algorithms.

woodsp-ibm:
In the kernels.algorithms init file here https://github.com/Qiskit/qiskit-machine-learning/blob/7382584428bda6ccbb0b0672284ac09abbfe587d/qiskit_machine_learning/kernels/algorithms/__init__.py#L18-L19

it has the same title as in its parent folder init, ie `Quantum Kernels` and comes out as below. Might this be better as `Quantum Kernel Algorithms`

image

adekusar-drl:
@manoelmarques could you please take a look if everything is correct for sphinx?

manoelmarques (approved):

PR #306, opened 2022-02-08
Add unit tests for QuantumKernel circuit batching

### LINKED ISSUE #242: Improve unit tests when batching is used in QuantumKernel
Recently `QuantumKernel` was improved when a statevector simulator is used, see #209. This modification makes use of `batch_size` parameter. The unit tests does not verify that this parameter is used correctly cause such check requires intercepting the calls to `QuantumInstance`. This issue is raised to improve the tests in this regard. For more details see the above mentioned PR.

declanmillar:
### Summary

This aims to improve the unit tests for QuantumKernel by testing that batch_size is used correctly for both statevector and QASM simulators. Closes #242.

This follows improvements in the QuantumKernel code related to batching circuits using the statevector simulator (#209).

### Details and comments

It adds a new test class that records the actual number of circuits being passed
to `QuantumInstance.execute` and checks that it does not exceed the `batch_size` requested by `Quantum Kernel`.
It also checks that the sum of the batch sizes matches the total number of expected circuits.
This is tested when using both the statevector and QASM simulators.

### DIFF
--- a/test/kernels/test_qkernel.py
+++ b/test/kernels/test_qkernel.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021.
+# (C) Copyright IBM 2021, 2022.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
--- a/test/kernels/test_qkernel.py
+++ b/test/kernels/test_qkernel.py
@@ -525,18 +525,40 @@ class TestQuantumKernelBatching(QiskitMachineLearningTestCase):
     to execute does not exceed the batch_size requested by the Quantum Kernel.
     Also checks that the sum of the batch sizes matches the total number of expected
     circuits.
+    Checks batching with both statevector simulator and QASM simulator.
     """
 
-    def count_circuits(self, f):
-        @functools.wraps(f)
+    def __init__(self):
+        super().__init__()
+        self.circuit_counts = []
+
+    def count_circuits(self, func):
+        """Decorator to record the number of circuits passed to QuantumInstance.execute.
+
+        Args:
+            func (Callable): execute function to be wrapped
+
+        Returns:
+            Callable: function wrapper
+        """
+
+        @functools.wraps(func)
         def wrapper(*args, **kwds):
             self.circuit_counts.append(len(args[1]))
-            return f(*args, **kwds)
+            return func(*args, **kwds)
 
         return wrapper
 
     def hook_quantum_instance(self):
+        """Hooks into QuantumInstance to monitor circuit batching.
+
+        Class factory that takes the QuantumInstance class, applies a method decorator to the
+        execute function and returns the modified class.
+        """
+
         class HookedQuantumInstance(QuantumInstance):
+            """Applies the count_circuits decorator to the QuantumInstance.execute function."""
+
             @self.count_circuits
             def execute(self, circuits, had_transpiled: bool = False):
                 return super().execute(circuits, had_transpiled)
--- a/test/kernels/test_qkernel.py
+++ b/test/kernels/test_qkernel.py
@@ -528,7 +528,7 @@ class TestQuantumKernelBatching(QiskitMachineLearningTestCase):
     Checks batching with both statevector simulator and QASM simulator.
     """
 
-    def __init__(self, methodName='runTest'):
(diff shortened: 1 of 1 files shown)


test/kernels/test_qkernel.py
```diff
@@ -515,5 +516,140 @@ def test_user_parameters(self):
...
+    The decorator function records the actual number of circuits being passed
+    to execute does not exceed the batch_size requested by the Quantum Kernel.
+    Also checks that the sum of the batch sizes matches the total number of expected
+    circuits.
+    Checks batching with both statevector simulator and QASM simulator.
+    """
+
+    def __init__(self, methodName="runTest"):
+        super().__init__(methodName)
+        self.circuit_counts = []
```
woodsp-ibm, line 533: It should be possible to put `self.circuit_counts = []` into the setup() method no? Would avoid having to do an init - which is unusual in a unit test - also it would avoid having to code it explicitly to do that in each unit test since setup is designed to run ahead of the test itself being invoked.

declanmillar: Yes, good point! Cheers. I've made that change.

test/kernels/test_qkernel.py
```diff
@@ -515,5 +516,136 @@ def test_user_parameters(self):
...
+        return wrapper
+
+    def hook_quantum_instance(self):
+        """Hooks into QuantumInstance to monitor circuit batching.
+
+        Class factory that takes the QuantumInstance class, applies a method decorator to the
+        execute function and returns the modified class.
+        """
+
+        class HookedQuantumInstance(QuantumInstance):
```
adekusar-drl, line 555: Do we really need a factory here? Can it be simplified?

test/kernels/test_qkernel.py
```diff
@@ -515,5 +516,136 @@ def test_user_parameters(self):
...
+                [3.07876080, 1.75929189],
+                [6.03185789, 5.27787566],
+                [6.22035345, 2.70176968],
+                [0.18849556, 2.82743339],
+            ]
+        )
+        self.label_train = np.asarray([0, 0, 1, 1])
+
+        self.sample_test = np.asarray([[2.199114860, 5.15221195], [0.50265482, 0.06283185]])
+        self.label_test = np.asarray([0, 1])
```
adekusar-drl, line 600: Can you please add a line how these values were obtained? E.g. if I want to extend the dataset, how can I generate more samples.

adekusar-drl (commented):
Thanks, the way how `QuantumInstance` is intercepted is a bit overcomplicated to me. I left a couple of comments.

test/kernels/test_qkernel.py
declanmillar: I absolutely agree that it seems overcomplicated, but I wasn't sure how to more simply hook into a method that we don't call explicitly. I'm probably missing something blindingly obvious. I'll revisit this.

declanmillar: I was indeed missing the blindingly obvious. Now implemented without resorting to a factory.

test/kernels/test_qkernel.py
```diff
@@ -515,5 +516,122 @@ def test_user_parameters(self):
...
+    def setUp(self):
+        super().setUp()
+
+        algorithm_globals.random_seed = 10598
+
+        self.shots = 12000
+        self.batch_size = 3
+        self.circuit_counts = []
+
+        QuantumInstance.execute = self.count_circuits(QuantumInstance.execute)
```
woodsp-ibm, line 555: This is changing the loaded class itself not an instance. This will affect any use later, in any other tests, by anything using this class. This is not a good idea. Now you can monkey patch the instance itself, rather than the class, after its created below.

declanmillar: Yep, excellent point. I've now monkey patched the two instances rather than the class.

test/kernels/test_qkernel.py
declanmillar: Sure, I've updated the data with a comment on its generation. Incidentally, I would probably have instinctively generated the data at runtime, but I note that data for unit tests is generally hardcoded throughout the module, presumably to avoid a dependency on ad_hoc_data.

adekusar-drl: If test data is hard coded you definitely know it is not a random sample. Also, `ad_hoc` may be slow in some cases.

adekusar-drl (approved):
Thanks, looks good to me.

PR #312, opened 2022-02-10
Fix VQC training with warm_start=True

### LINKED ISSUE #296: The warm_start=True attribute in the VQC class is broken
### Environment

- **Qiskit Machine Learning version**: 0.3.0
- **Python version**: 3.9.7
- **Operating system**: Red Hat 7.9

### What is happening?

Trying to perform two or more fits with the current VQC class, when `warm_start=True`, produces the following error:
```
  File "/work/deodagiu/bug_examples/vqc_warm_start.py", line 33, in <module>
    vqc.fit(data[10:, :], onehot_target[10:])
  File "/work/deodagiu/miniconda/envs/qiskit_bugs/lib/python3.9/site-packages/qiskit_machine_learning/algorithms/classifiers/vqc.py", line 167, in fit
    return super().fit(X, y)
  File "/work/deodagiu/miniconda/envs/qiskit_bugs/lib/python3.9/site-packages/qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py", line 109, in fit
    x0=self._choose_initial_point(),
  File "/work/deodagiu/miniconda/envs/qiskit_bugs/lib/python3.9/site-packages/qiskit_machine_learning/algorithms/trainable_model.py", line 204, in _choose_initial_point
    self._initial_point = self._fit_result[0]
TypeError: 'OptimizerResult' object is not subscriptable
```
**This bug was introduced with 0.3.0. In previous versions, `warm_start=True` did not produce this issue.**

Run the following code:
```
from time import perf_counter
import numpy as np

from qiskit import Aer

# Choose a seed for the quantum instance initialisation.
seed = 123

# Construct toy data set.
data = np.random.rand(20, 4)
target = [0, 0, 0, 0, 0, 1, 1, 1, 1, 1]
target = np.tile(target, 2)

# Onehot encode the target.
onehot_target = np.zeros((target.size, int(target.max() + 1)))
onehot_target[np.arange(target.size), target.astype(int)] = 1

# Initialise the VQC.
backend = Aer.get_backend("aer_simulator_statevector")
qin = QuantumInstance(backend, seed_simulator=seed, seed_transpiler=seed)
vqc = VQC(num_qubits=4, feature_map=None, ansatz=None, loss='cross_entropy',
          optimizer=None, warm_start=True, quantum_instance=qin,
          initial_point=None, callback=None)

# Fit the VQC to the constructed toy data. Time the fit.
fit_time_init = perf_counter()
vqc.fit(data[:10, :], onehot_target[:10])
fit_time_fina = perf_counter()
print(f"Fit completed in: {fit_time_fina-fit_time_init:.2e} s")
fit_time_init = perf_counter()
vqc.fit(data[10:, :], onehot_target[10:])
fit_time_fina = perf_counter()
print(f"Fit completed in: {fit_time_fina-fit_time_init:.2e} s")
```

The output of this code is shown in the **What is happening?** section.

### What should happen?

The VQC should complete the fit without any issues.

### Any suggestions?

_No response_

declanmillar:
### Summary

Fixes the extraction of the initial_point in TrainableModel from the final point of the
minimization in the existing OptimizerResult object. Closes #296.

Adds unit tests to ensure VQC correctly handles warm starts.

### Details and comments

As pointed out by @woodsp-ibm, "In Terra the optimizers were refactored and the old optimize method, that returned a tuple (subscriptable) was deprecated and a new method minimize was created that returns OptimizerResult object."

This is a very small fix to reflect that change and some related unit tests that would have caught the issue.

### DIFF
--- a/qiskit_machine_learning/algorithms/trainable_model.py
+++ b/qiskit_machine_learning/algorithms/trainable_model.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021.
+# (C) Copyright IBM 2021, 2022.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
--- a/qiskit_machine_learning/algorithms/trainable_model.py
+++ b/qiskit_machine_learning/algorithms/trainable_model.py
@@ -201,7 +201,7 @@ class TrainableModel:
             An array as an initial point
         """
         if self._warm_start and self._fit_result is not None:
-            self._initial_point = self._fit_result[0]
+            self._initial_point = self._fit_result.x
         elif self._initial_point is None:
             self._initial_point = algorithm_globals.random.random(self._neural_network.num_weights)
         return self._initial_point
--- a/releasenotes/notes/fix-vqc-warm-start-1c15d3f5742c5e84.yaml
+++ b/releasenotes/notes/fix-vqc-warm-start-1c15d3f5742c5e84.yaml
@@ -1,11 +1,11 @@
 ---
 fixes:
   - |
-    Fixes an issue where `VQC` would fail with `warm_start=True`.
-    The extraction of the `initial_point` in `TrainableModel` from the final
+    Fixes an issue where ``VQC`` would fail with ``warm_start=True``.
+    The extraction of the ``initial_point`` in ``TrainableModel`` from the final
     point of the minimization had not been updated to reflect the refactor
-    of optimizers in `qiskit-terra`; the old `optimize` method, that returned
-    a tuple was deprecated and new method `minimize` was created that returns
-    an `OptimizerResult` object. We now correctly recover the final point of
+    of optimizers in ``qiskit-terra``; the old ``optimize`` method, that returned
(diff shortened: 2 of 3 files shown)


woodsp-ibm (commented):
Can you add a `reno` release note with a fixes section. This is required for issues that impact/affect the public API.

Thanks for adding a unit test that checks this. This is something that we always want if we find our CI has a hole like this case where the tests did not catch it. I see you have modelled it with data after what I see is done elsewhere here. I will note that that data values could be more directly coded into the data e.g directly have an instance of the optimizer etc rather than having this indirect via string values. ddt can also input to multiple parameters and do unpacking via  decorator so the test does not need to have that. I am not asking for anything to be changed merely pointing this flexibility out that you may want to look at going forwards if/when using ddt. Eg from our former Aqua code https://github.com/Qiskit/qiskit-aqua/blob/c1564af8792c6664670807614a378147fd04d28f/test/aqua/test_vqe.py#L126-L131

declanmillar:
This PR should be labelled as `Changelog: Bugfix` and `stable backport potential`. I don't appear to have the necessary permissions to apply these myself.

woodsp-ibm (approved):

PR #317, opened 2022-02-14
fixes #297: fitting in batches without all labels

### LINKED ISSUE #297: Training the VQC in batches, where one batch happens to have only one label, leads to error
### Environment

- **Qiskit Machine Learning version**: 0.3.0
- **Python version**: 3.9.7
- **Operating system**: Red Hat 7.9 

### What is happening?

I am training a VQC with batches of data. Whenever I am lucky/unlucky enough that a batch only contains one of the two classes in my data set, the VQC throws the following error:
```
  File "/work/deodagiu/bug_examples/vqc_one_class.py", line 29, in <module>
    vqc.fit(data[:5, :], onehot_target[:5])
  File "/work/deodagiu/miniconda/envs/qiskit_bugs/lib/python3.9/site-packages/qiskit_machine_learning/algorithms/classifiers/vqc.py", line 167, in fit
    return super().fit(X, y)
  File "/work/deodagiu/miniconda/envs/qiskit_bugs/lib/python3.9/site-packages/qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py", line 95, in fit
    raise QiskitMachineLearningError(
qiskit_machine_learning.exceptions.QiskitMachineLearningError: 'Current settings only applicable to binary classification! Got labels: [[1. 0.]\n [1. 0.]\n [1. 0.]\n [1. 0.]\n [1. 0.]]'
```

Reproduce the issue by running the following code:
```
from time import perf_counter
import numpy as np

from qiskit import Aer

# Choose a seed for the quantum instance initialisation.
seed = 123

# Construct toy data set.
data = np.random.rand(20, 4)
target = [0, 0, 0, 0, 0, 1, 1, 1, 1, 1]
target = np.tile(target, 2)

# Onehot encode the target.
onehot_target = np.zeros((target.size, int(target.max() + 1)))
onehot_target[np.arange(target.size), target.astype(int)] = 1

# Initialise the VQC.
backend = Aer.get_backend("aer_simulator_statevector")
qin = QuantumInstance(backend, seed_simulator=seed, seed_transpiler=seed)
vqc = VQC(num_qubits=4, feature_map=None, ansatz=None, loss='cross_entropy',
          optimizer=None, warm_start=True, quantum_instance=qin,
          initial_point=None, callback=None)

# Fit the VQC to the constructed toy data. Time the fit.
fit_time_init = perf_counter()
vqc.fit(data[:5, :], onehot_target[:5])
fit_time_fina = perf_counter()
print(f"Fit completed in: {fit_time_fina-fit_time_init:.2e} s")
fit_time_init = perf_counter()
vqc.fit(data[5:10, :], onehot_target[5:10])
fit_time_fina = perf_counter()
print(f"Fit completed in: {fit_time_fina-fit_time_init:.2e} s")
```
The output is the same as the one in the **What is happening?** section.

### What should happen?

The VQC should compelete the fit and move on with it's training, given that the shape of the onehot encoded data is indicative that there's two classes in the data set and that the VQC, as it currently stand, is hardcoded to only use onehot encoded data.

### Any suggestions?

Change the conditions in the following line of code:

https://github.com/Qiskit/qiskit-machine-learning/blob/585b795caab3e080902e0741e374eb2759fca777/qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py#L94

declanmillar:
### Summary

Closes #297. Currently, VQC will throw an error if trained on batches of data where not all of the target labels that can be found in the full dataset are present. This is because VQC interprets the number of unique labels in the current batch as the number of classes. 

Currently, VQC is hard-coded to expect one-hot encoded labels. Therefore, I suggest a small change to determine the number of classes from the shape of the label array. I have also edited the docstring to add the label encoding requirements of the current VQC implementation.

### Details and comments

Related issues that I will raise separately:
- I may also add a warning when running with a warm start and fitting to data with a different number of one-hot labels, as I would expect this to invalidate any previous fit.
- While multi-label classification is not supported, the code will not raise an error when it is provided; it might be beneficial to raise an error in such cases.
- In future, it might be good for VQC to accept different label encodings, in line with what an sklearn user might expect.
- It may simplify the unit tests to add objects directly to the dtd data decorator, rather than using string tags, as discussed by @woodsp-ibm on #312.

### DIFF
--- a/qiskit_machine_learning/algorithms/classifiers/vqc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/vqc.py
@@ -9,7 +9,7 @@
 # Any modifications or derivative works of this code must retain this
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
-"""An implementation of quantum neural network classifier."""
+"""An implementation of variational quantum classifier."""
 
 from typing import Union, Optional, Callable, cast
 import numpy as np
@@ -27,7 +27,17 @@ from .neural_network_classifier import NeuralNetworkClassifier
 
 
 class VQC(NeuralNetworkClassifier):
-    """Quantum neural network classifier."""
+    """Variational quantum classifier.
+
+    The variational quantum classifier is a variational algorithm where the
+    measured expectation value is interpreted as the output of a classifier.
+
+    Only supports one-hot encoded labels;
+    e.g., data like ``[1, 0, 0]``, ``[0, 1, 0]``, ``[0, 0, 1]``.
+
+    Multi-label classification is not supported;
+    e.g., data like ``[1, 1, 0]``, ``[0, 1, 1]``, ``[1, 0, 1]``, ``[1, 1, 1]``.
+    """
 
     def __init__(
         self,
@@ -155,12 +165,12 @@ class VQC(NeuralNetworkClassifier):
 
         Args:
             X: The input data.
-            y: The target values.
+            y: The target values. Required to be one-hot encoded.
 
         Returns:
             self: returns a trained classifier.
         """
-        num_classes = len(np.unique(y, axis=0))
+        num_classes = y.shape[-1]
         cast(CircuitQNN, self._neural_network).set_interpret(
             self._get_interpret(num_classes), num_classes
         )
--- a/test/algorithms/classifiers/test_vqc.py
+++ b/test/algorithms/classifiers/test_vqc.py
@@ -12,7 +12,9 @@
 
 """ Test Neural Network Classifier """
 
+from typing import Callable
 import unittest
+import functools
 
 from test import QiskitMachineLearningTestCase
 
@@ -32,6 +34,8 @@ class TestVQC(QiskitMachineLearningTestCase):
 
     def setUp(self):
         super().setUp()
+
+        self.num_classes_by_batch = []
 
         # specify quantum instances
         algorithm_globals.random_seed = 12345
@@ -300,6 +304,16 @@ class TestVQC(QiskitMachineLearningTestCase):
         score = classifier.score(X, y)
         self.assertGreater(score, 0.5)
 
+    def get_num_classes(self, func: Callable) -> Callable:
+        """Wrapper to record the number of classes assumed when building CircuitQNN."""
+
+        @functools.wraps(func)
+        def wrapper(num_classes:int):
+            self.num_classes_by_batch.append(num_classes)
+            return func(num_classes)
+
+        return wrapper
+
     @data(
         # optimizer, quantum instance
         ("cobyla", "statevector"),
@@ -334,6 +348,7 @@ class TestVQC(QiskitMachineLearningTestCase):
         # Construct the data.
         features = algorithm_globals.random.random((15, num_inputs))
         target = np.asarray([0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2])
+        num_classes = len(np.unique(target))
 
         # One-hot encode the target.
         target_onehot = np.zeros((target.size, int(target.max() + 1)))
@@ -348,6 +363,8 @@ class TestVQC(QiskitMachineLearningTestCase):
             quantum_instance=quantum_instance,
         )
 
+        classifier._get_interpret = self.get_num_classes(classifier._get_interpret)
+
         # Fit the VQC to the first third of the data.
         classifier.fit(features[:5, :], target_onehot[:5])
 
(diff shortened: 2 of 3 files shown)


woodsp-ibm:
Could we get a test case that would fail previously that will now work with this bug fix. Since its an aspect that CI did not catch it would be good to get a unit test in there to cover it.

Also don't forget the reno release note.

declanmillar:
>

declanmillar:
> Could we get a test case that would fail previously that will now work with this bug fix. Since its an aspect that CI did not catch it would be good to get a unit test in there to cover it.
> 
> Also don't forget the reno release note.

I've added a new unit test that would fail without this fix. It may be necessary to revise it if and when we implement a wider range of valid label encodings in the VQC interface. I've also added a reno note for the fix.

adekusar-drl (dismissed):
Looks good to me!
@woodsp-ibm Do you have open question?

test/algorithms/classifiers/test_vqc.py
```diff
@@ -294,6 +294,61 @@ def test_warm_start(self, config):
...
+        (None, "qasm"),
+    )
+    def test_batches_with_incomplete_labels(self, config):
+        """Test VQC when some batches do not include all possible labels."""
+        opt, q_i = config
+
+        if q_i == "statevector":
+            quantum_instance = self.sv_quantum_instance
+        else:
+            quantum_instance = self.qasm_quantum_instance
```
dlasecki, line 313: We might want to make this `if` explicit in case someone adds other backends to the test input in the future.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -155,12 +165,12 @@ def fit(self, X: np.ndarray, y: np.ndarray):  # pylint: disable=invalid-name
 
         Args:
             X: The input data.
-            y: The target values.
+            y: The target values. Required to be one-hot encoded.
```
dlasecki, line 168: Shall we add some input validation?

dlasecki (changes requested):

qiskit_machine_learning/algorithms/classifiers/vqc.py
declanmillar: I was thinking to open a new PR for input validation, but am happy to include it in this one!

woodsp-ibm: Given this is a bug fix, and will be backported, we try to keep the code to what is needed to correct the critical bug for users in order to minimize code change and potentially destabilize code in the stable branch via other function such as new features. I think I might count validating the input - presumably checking all the labels as a new feature/improvement. If somehow you consider it vital its with this - otherwise I would consider it preferable on a separate PR that is not backported and would be part of next main release.

test/algorithms/classifiers/test_vqc.py
```diff
@@ -294,6 +294,61 @@ def test_warm_start(self, config):
...
+        )
+
+        # Fit the VQC to the first third of the data.
+        classifier.fit(features[:5, :], target_onehot[:5])
+
+        # Fit the VQC to the second third of the data with a warm start.
+        classifier.fit(features[5:10, :], target_onehot[5:10])
+
+        # Fit the VQC to the third third of the data with a warm start.
+        classifier.fit(features[10:, :], target_onehot[10:])
```
woodsp-ibm, line 350: Is there anything we can assert here to test the outcome is as expected.? I imagine this test gets to the end and completes now rather than fails as it would have before the fix - so in that regard the outcome is implicitly tested. But it would be nice if something could be asserted expllicitly to better assess correctness.

qiskit_machine_learning/algorithms/classifiers/vqc.py
adekusar-drl: I agree to introduce new input checks in a separate PR.

test/algorithms/classifiers/test_vqc.py
declanmillar: Yep, seems sensible. Done.

test/algorithms/classifiers/test_vqc.py
declanmillar: Yes, good point. I will add an assertion that the number of classes assumed by VQC is the same for each batch.

qiskit_machine_learning/algorithms/classifiers/vqc.py
declanmillar: Ok. I'll open a separate issue and PR for input validation.

adekusar-drl:
@woodsp-ibm Is there anything left on your radar? If not, I'll approve and merge this PR.

woodsp-ibm (approved):
> Is there anything left on your radar? If not, I'll approve and merge this PR.
All looks good to me. May want to create issues for the other improvements/changes that are noted in the main comment so as we don't lose them.

adekusar-drl:
@declanmillar could you please post an issue(s) mentioned in this PR?

PR #319, opened 2022-02-16
Deprecate some dataset functions

manoelmarques:
### Summary
Deprecate all dataset functions but `ad_hoc_data` still used by tutorials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,25 +46,25 @@ be classified.
 from qiskit import BasicAer
 from qiskit.utils import QuantumInstance, algorithm_globals
 from qiskit.algorithms.optimizers import COBYLA
-from qiskit.circuit.library import TwoLocal
+from qiskit.circuit.library import TwoLocal, ZZFeatureMap
 from qiskit_machine_learning.algorithms import VQC
-from qiskit_machine_learning.datasets import wine
-from qiskit_machine_learning.circuit.library import RawFeatureVector
+from qiskit_machine_learning.datasets import ad_hoc_data
 
 seed = 1376
 algorithm_globals.random_seed = seed
 
-# Use Wine data set for training and test data
-feature_dim = 4  # dimension of each data point
-training_size = 12
-test_size = 4
+# Use ad hoc data set for training and test data
+feature_dim = 2  # dimension of each data point
+training_size = 20
+test_size = 10
 
 # training features, training labels, test features, test labels as np.array,
 # one hot encoding for labels
 training_features, training_labels, test_features, test_labels = \
-    wine(training_size=training_size, test_size=test_size, n=feature_dim)
+    ad_hoc_data(
+            training_size=training_size, test_size=test_size, n=feature_dim, gap=0.3)
 
-feature_map = RawFeatureVector(feature_dimension=feature_dim)
+feature_map = ZZFeatureMap(feature_dimension=feature_dim, reps=2, entanglement="linear")
 ansatz = TwoLocal(feature_map.num_qubits, ['ry', 'rz'], 'cz', reps=3)
 vqc = VQC(feature_map=feature_map,
           ansatz=ansatz,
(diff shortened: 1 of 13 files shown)


adekusar-drl (approved):
Thanks, looks good!

woodsp-ibm:
In terms of removing these convenience methods to access/transform the data ready for the algorithms we were going to have in their stead a tutorial right? that shows using a standard data set and the rationales behind the reductions/scaling and then it use with an algorithm. This would help with questions that are often fielded about how do I use/bring in my own data set as they can follow the tutorial. It can be a separate issue to do this for sure - I am not suggesting adding it via this PR but we would want people to start doing this to avoid these now deprecated datasets. We could also reference the tutorial in the deprecation - at present it does not seem there is any guidance on what to do as a replacement, something we normally have.

On the readme sample - maybe that can go back to the Wine dataset but with explicit scaling etc so its directly used from sklearn rather that via the convenience functions, It would seem better for the sample that its closer to real world than the adhoc dataset.

I think the Adhoc dataset deprecation needed a bit more care as its used by the textbook. Maybe it could be discussed with them as the alternative was to have it in tutorials directly which maybe they could use from there or a copy could be made for the textbook to make things more independent. I guess it would nice if that were gone from here too albeit that it will still be in this repo for the tutorial here just in a different place, but no longer a support API/function of ML

manoelmarques:
The adhoc dataset is the last one remaining and it  wasn't deprecated in this PR.

If using it is not the best solution for the readme, tutorials then we should address it in an issue or separate PR.

Using the code from inside the wine utility function directly into the readme adds around 20 more lines of code counting the imports. Maybe it is a better solution but it could be done in a separate PR with the tutorial changes.

I am merging this PR but noting that @woodsp-ibm comments above are valid but probably not for this PR.

If @adekusar-drl believes we should change the readme now and use the internals of `wine` on it I can create a quick PR just for the readme change.

adekusar-drl:
As I mentioned in another discussion, the plan is to write a new tutorial, hopefully in the Qiskit Advocate Program, where we pick a dataset from scikit-learn, e.g. `iris` or `wine`, apply some basic data exploratory analysis, feature reduction, feature transformations if required, and then train a model. Basically, in this tutorial we have to apply all the steps are in these helper functions, but in a cleaner way with some explanations why we do this. Perhaps, we should have written such a tutorial from the beginning of this module to avoid these questions. Although, we tried to design the API as close as possible to scikit-learn and such questions should not be an issue.

As for the readme sample, I admit, ad hoc is not the best choice, it is artificial. But I also agree with Manoel, getting back to wine will require much more lines in the sample. So, I'd prefer to keep it simple. May be, this new tutorial should become a readme note, this may be a good choice as well.

woodsp-ibm:
@adekusar-drl Might it be worthwhile to put these things into an epic? around deprecating the datasets. The actual deprecation has already been done, but it could include line items for the tutorial as above, maybe something for the readme, and perhaps also one to look further with the textbook team that adhoc could be moved into tutorials where its used.

I agree that adding 20+ lines of code to the readme is not good. Is this the same choosing a different data set where picking something else, aside from our adhoc set, makes no difference. I guess I have to query is this what we will be asking of users each time they use any of the datasets direct from sklearn now without the convenience methods when they want to use these quantum algorithms. Is there any sort of general utility that could be provided here that did the reduction etc on whatever dataset you had. That might save a bunch of boiler plate code for users (and for us) if such a thing is possible/practical..

adekusar-drl:
I agree, I'll open an issue (can be converted to epic once somebody is assigned) with:
- consider ad_hoc deprecation
- update readme
- update tutorials
- update text book

Regarding dataset pre-processing, `scikit-learn` has a lot of transformers that can be easily used for such things as feature transformation, label encoding, dimensionality reduction and so on. The problem is that some datasets require some transformations, other datasets require other transformations and there's no a single pipeline that suits all cases. Even if we come up with such a "universal" transformation there will be a dataset that requires an additional step we missed. Moreover, labels can be encoded differently depending on the loss function we want to apply. We can provide some guidance, but in overall data preparation step is on users. The documentation can be enhanced, e.g. more snippets to show what kind of data is supported by a certain algorithm/model. And such documentation update can be another issue.

PR #332, opened 2022-02-22
Updated construct circuit unit tests for QuantumKernel

### LINKED ISSUE #149: Review the tests in TestQuantumKernelConstructCircuit
Recently, the tests in `TestQuantumKernelConstructCircuit` were broken, a quick fix (PR: https://github.com/Qiskit/qiskit-machine-learning/pull/148) showed that these tests construct circuits and validate the number of gates in the circuits. Since the way how complex gates are represented in circuits has been changed (PR: https://github.com/Qiskit/qiskit-terra/pull/6634) these tests became less valueable. This goal of this issue is to analyze these tests and come up with a decision what to do with them. There are a few options: remove the tests, rewrite it in a way to make them better, or restructure them completely, e.g. test circtuits slightly differently.

adekusar-drl:
### Summary
This PR slightly changes the unit tests that test `QuantumKernel.construct_circuit()`. Closes #149.

### Details and comments
Instead of testing the size of the output circuits, these tests check for the number of qubits, feature map operations, optionally if measurement operations are applied and inversibility of the feature maps if only one vector is passed. Should be better than the original tests.

### DIFF
--- a/test/kernels/test_qkernel.py
+++ b/test/kernels/test_qkernel.py
@@ -21,7 +21,7 @@ import numpy as np
 
 from sklearn.svm import SVC
 
-from qiskit import BasicAer
+from qiskit import BasicAer, QuantumCircuit
 from qiskit.circuit import Parameter
 from qiskit.circuit.library import ZZFeatureMap
 from qiskit.utils import QuantumInstance, algorithm_globals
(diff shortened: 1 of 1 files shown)


woodsp-ibm (approved):
It seems fine to me. One could argue that perhaps the test includes private knowledge of the exact circuit structure that the quantum kernel creates. As long as it provides a circuit back the main testing of it in my view is done indirectly when testing evaluate() since that internally ends up calling this and we get coverage from that too. But I guess its all good.

PR #343, opened 2022-03-09
Add extra installs to tox

manoelmarques:
### Summary
This allows for unit testing torch and sparse when using tox


woodsp-ibm (approved):

garrison:
I think the `minversion` should be set to `2.4` along with this change, as `extras` [was not introduced until then](https://tox.wiki/en/latest/config.html#conf-extras).

PR #349, opened 2022-03-16
Add GPU tests

adekusar-drl:
### Summary
Refactored hybrid model tests and added tests that must run on GPU.

### DIFF
--- a/test/connectors/test_cpu_torch_connector.py
+++ b/test/connectors/test_cpu_torch_connector.py
@@ -15,15 +15,10 @@ import unittest
 from test import QiskitMachineLearningTestCase
 from test.connectors.test_torch_connector import TestTorchConnector
 
-from qiskit.utils import optionals
-import qiskit_machine_learning.optionals as _optionals
-
 
 class TestCPUTorchConnector(QiskitMachineLearningTestCase, TestTorchConnector):
     """CPU based tests of the PyTorch connector."""
 
-    @unittest.skipUnless(optionals.HAS_AER, "qiskit-aer is required to run this test")
-    @unittest.skipIf(not _optionals.HAS_TORCH, "PyTorch not available.")
     def setUp(self):
         super().setup_test()
         super().setUp()
--- a/test/connectors/test_cpu_torch_connector.py
+++ b/test/connectors/test_cpu_torch_connector.py
@@ -9,32 +9,28 @@
 # Any modifications or derivative works of this code must retain this
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
-"""GPU based tests of hybrid PyTorch networks."""
+"""CPU based tests of the PyTorch connector."""
 
 import unittest
-from test import QiskitMachineLearningTestCase, gpu
-from test.connectors.test_torch_networks import TestTorchNetworks
+from test import QiskitMachineLearningTestCase
+from test.connectors.test_torch_connector import TestTorchConnector
 
 from qiskit.utils import optionals
 import qiskit_machine_learning.optionals as _optionals
 
 
-class TestGPUTorchConnector(QiskitMachineLearningTestCase, TestTorchNetworks):
-    """GPU based tests of hybrid PyTorch networks."""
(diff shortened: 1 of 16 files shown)


adekusar-drl:
@manoelmarques Could you please have a quick look if this PR makes sense? I fully re-use the existing tests and for GPU tests I force them to be executed on GPU.

manoelmarques:
@adekusar-drl The `@gpu` decorator is like Terra's `@slow_test` decorator. It is a function decorator, to be used above the method, not the class.

adekusar-drl:
@woodsp-ibm Can you please review?

test/connectors/test_gpu_torch_connector.py
```diff
@@ -0,0 +1,42 @@
...
+    @gpu
+    def setUp(self):
+        super().setup_test()
+        super().setUp()
+
+        import torch
+
+        if not torch.cuda.is_available():
+            # raise test exception
+            self.skipTest("CUDA is not available")
```
woodsp-ibm, line 36: I imagine this could have been done as a decorator too right?

test/connectors/test_cpu_torch_connector.py
```diff
@@ -0,0 +1,37 @@
...
+
+from qiskit.utils import optionals
+import qiskit_machine_learning.optionals as _optionals
+
+
+class TestCPUTorchConnector(QiskitMachineLearningTestCase, TestTorchConnector):
+    """CPU based tests of the PyTorch connector."""
+
+    @unittest.skipUnless(optionals.HAS_AER, "qiskit-aer is required to run this test")
+    @unittest.skipIf(not _optionals.HAS_TORCH, "PyTorch not available.")
```
woodsp-ibm, line 26: Just curious why one uses `skipUnless` and the other `skipIf(not ` - it just seems odd since I think they end up doing the same right (same in the similar class for GPU)

manoelmarques: Yes, it is just because the AER test was copied form other files in ML and were initially copied from Terra that uses `skipunless` for Aer and uses `skipIf` for other optional libraries but they do the same.

test/connectors/test_torch_connector.py
```diff
@@ -61,6 +62,42 @@ def setUp(self):
 
         torch.manual_seed(algorithm_globals.random_seed)
 
+    @abstractmethod
```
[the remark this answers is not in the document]
manoelmarques: If they need to derive from `TestTorchConnector` yes but that is the same solution use in Nature. Another solution could be adding a `pylint: ignore` to ignore warnings on not found methods but that would cause any method misspelled in this file to fail only on runtime.
https://github.com/Qiskit/qiskit-nature/blob/bff9c6a5fe674485c8588add27534bcad43b3276/test/drivers/second_quantization/test_driver.py#L46

test/connectors/test_cpu_torch_connector.py
woodsp-ibm: I get they do the same - and maybe in one library in different places it would seem a lot less weird than when the occurrences are back-to-back like this. It makes me wonder whether there is some reason for it since while they should work the same way why are the two different forms used here by the author. As in why is the PyTorch one not simply a skipUnless too.

test/connectors/test_torch_connector.py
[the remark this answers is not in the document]
woodsp-ibm: I guess its been a while since those driver tests and I did not remember them being done like this - I recalled they had the common parent with the tests just not that it included signatures like these too. Maybe there is no better way to do this type of thing to avoid this "awkwardness". It all works so I guess its good.

test/connectors/test_cpu_torch_connector.py
manoelmarques: I changed to skipUnless and refactored to a common base class.

test/connectors/test_torch_connector.py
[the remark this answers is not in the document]
manoelmarques: I refactored it to a common base class since they were repeated in two base classes.

woodsp-ibm (approved):

PR #361, opened 2022-03-31
Remove deprecations and deprecate unused CrossEntropySigmoidLoss

adekusar-drl:
### Summary
This PR removes support for "l1" and "l2" loss definitions, use "absolute_error" and "squared_error" instead. Also this PR deprecates `CrossEntropySigmoidLoss` as unused.

PR #362 to be merged first to main and stable.


releasenotes/notes/remove-deprecation-deprecate-sigmoid-319213068ae00ea9.yaml
```diff
@@ -0,0 +1,7 @@
+---
+deprecations:
+  - |
+    Deprecated class ``CrossEntropySigmoidLoss`` as it is not used in the package.
```
woodsp-ibm, line 4: I am trying to understand this statement. The class exists as a loss function and could be specified on TrainableModel as the loss to be used, so `it is not used in the package` is hard for me to follow, as it seems it could be used and specified as a loss where its applicable for an algorithm.

adekusar-drl: OK, may be unclear. The loss function is not called anywhere in QML. It could be used, but never used actually. We don't actually need it.

adekusar-drl: That's right. It can be used externally, but you wrote "is not useful to keep". That's the only reason I want to deprecate it.

woodsp-ibm (approved):

Issue #159, opened 2021-07-13
Extend type hints in the classifiers and regressors


adekusar-drl:
### What is the expected behavior?
Since we are going to introduce categorical labels (and perhaps categorical features), I think, we can re-consider type hints for the classifiers and regressors and make them more like in Scikit-Learn: `array_like`. This may allow to pass plain lists or other types. As an example take a look at the interface here: https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html

This issue is just for discussion.

edoaltamura:
@adekusar-drl, could you please share an example or template of how you'd match the SKlearn type hinting for QML?

adekusar-drl:
It has been a low priority for a long time, things might have changed. While scikit learn does not expose type hints by default, one of the possible options to take a look at https://numpy.org/doc/stable/reference/typing.html#module-numpy.typing. Numpy has `ArrayLike` type hint and it may work. This issue is more exploratory one rather something that can be easily implemented. Even if numpy's hinting is good enough, I'd first double check if such hinting brings in any value, not problems.

edoaltamura:
For instance
https://github.com/qiskit-community/qiskit-machine-learning/blob/c59063af02afbb56cfdb27408068aa1c0f5bc718/qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py#L128-L130
would become
```python
import numpy.typing as npt

def fit( 
     self, X: npt.ArrayLike, y: npt.ArrayLike, sample_weight: npt.ArrayLike | None = None 
 ) -> "PegasosQSVC": 
```
or, more strictly,
```python
import numpy.typing as npt

def fit( 
     self, X: npt.NDArray[np.float64], y: npt.NDArray[np.float64], sample_weight: npt.NDArray[np.float64] | None = None 
 ) -> "PegasosQSVC": 
```
While young, the new typing in Numpy is covered by tests and CI, and made it into a major release. So I'd consider the basic features relatively stable.

edoaltamura:
Mypy can also be configured to account for strongly types Numpy objects as
```
[mypy]
plugins = numpy.typing.mypy_plugin
```
See https://numpy.org/doc/stable/reference/typing.html .

edoaltamura:
Closing this issue in favour of #975, which covers the more recent Rust bindings. Please refer to #975 for follow-on details.

PR #364, opened 2022-04-04
Add Effective Dimension class

### LINKED ISSUE #340: Add Effective Dimension Class
### What should we add?

The goal of this issue is to track the development of the effective dimension class, as described in epic #318. The latest version of this class can be found in my personal fork: https://github.com/ElePT/qiskit-machine-learning/tree/eff-dim, in `qiskit_machine_learnig/algorithms/effective_dimension.py
`

ElePT:
### Summary
This PR closes #340 by adding the `LocalEffectiveDimensionClass` introduced in https://arxiv.org/abs/2112.04807, taking the original experiments' code as reference: https://github.com/amyami187/effective_dimension.

### Details and comments
The local effective dimension shares all the logic from the global effective dimension algorithm, with the only difference being the limitation in number of parameter sets used. However, it has been discussed to be conceptually different enough to become its own class. 

Please let me know if you disagree with any variable names.


qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def __init__(
+        self,
+        qnn: NeuralNetwork,
+        params: Optional[Union[List[float], np.ndarray, float]] = None,
+        inputs: Optional[Union[List[float], np.ndarray, float]] = None,
+        num_params: Optional[int] = 1,
+        num_inputs: Optional[int] = 1,
+        fix_seed=False,
+        callback: Optional[Callable] = None,
```
woodsp-ibm, line 39: I think we can do better with the Callable typehint here so as to include its parameter types. Eg this is from the vqe callback `callback: Optional[Callable[[int, np.ndarray, float, float], None]] = None`

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,337 @@
...
+    ) -> None:
+        # pylint: disable=wrong-spelling-in-docstring
```
woodsp-ibm, line 41: Rather than disable spelling it would be better to add custom word(s) to .pylintdict file

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,337 @@
...
+    """
+
+    def __init__(
+        self,
+        qnn: NeuralNetwork,
+        params: Optional[Union[List[float], np.ndarray, float]] = None,
+        inputs: Optional[Union[List[float], np.ndarray, float]] = None,
+        num_params: Optional[int] = 1,
+        num_inputs: Optional[int] = 1,
+        fix_seed=False,
```
woodsp-ibm, line 38: Unless there is good reason to have a different seed on each instance we could go with the algorithm globals random number generator. When seeded that gives a reproducible sequence, otherwise by default its seeded randomly.

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,337 @@
...
+        from the original paper."""
+        return self._model.num_weights
+
+    @property
+    def num_params(self) -> int:
+        """Returns the number of network parameter sets."""
+        return self._num_params
+
+    @num_params.setter
+    def num_params(self, num_params) -> None:
```
woodsp-ibm, line 94: Typehint for num_params missing

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,337 @@
...
+
+    @property
+    def params(self) -> np.ndarray:
+        """Returns network parameters."""
+        return self._params
+
+    @params.setter
+    def params(self, params: Union[List[float], np.ndarray, float]) -> None:
+        """Sets network parameters."""
+        if params is not None:
```
woodsp-ibm, line 106: If params is allowed to be None here then the typehint should have Optional i.e. `Optional[Union[List[float], np.ndarray, float])]`

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,337 @@
...
+            self._params = np.asarray(params)
+            self.num_params = len(self._params)
+        else:
+            if self._seed is not None:
+                np.random.seed(self._seed)
+            # random sampling from uniform distribution
+            params = np.random.uniform(0, 1, size=(self.num_params, self._model.num_weights))
```
woodsp-ibm, line 113: See above comment in regards of using algorithm globals.

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,337 @@
...
+            self._params = params
+
+    @property
+    def num_inputs(self) -> int:
+        """Returns the number of input sets."""
+        return self._num_inputs
+
+    @num_inputs.setter
+    def num_inputs(self, num_inputs) -> None:
```
woodsp-ibm, line 122: Typehint on num_inputs

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,337 @@
...
+        grads = np.zeros(
+            (
+                self.num_inputs * self.num_params,
+                self._model.output_shape[0],
+                self._model.num_weights,
+            )
+        )
+        outputs = np.zeros((self.num_inputs * self.num_params, self._model.output_shape[0]))
+
+        # could this be batched further?
```
woodsp-ibm, line 163: Are you intending to answer this and resolve one way or the other

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,337 @@
...
+
+        # post-processing in the case of OpflowQNN output, to match the CircuitQNN output format
+        if isinstance(self._model, OpflowQNN):
+            grads = np.concatenate([grads / 2, -1 * grads / 2], 1)
+            outputs = np.concatenate([(outputs + 1) / 2, (1 - outputs) / 2], 1)
+
+        return grads, outputs
+
+    def get_fishers(self, gradients: np.ndarray, model_outputs: np.ndarray) -> np.ndarray:
+        # pylint: disable=wrong-spelling-in-docstring
```
woodsp-ibm, line 187: Same comment about spelling as above

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,337 @@
...
+    """
+
+    def __init__(
+        self,
+        qnn: NeuralNetwork,
+        num_inputs: Optional[int] = 1,
+        params: Optional[Union[List, np.ndarray]] = None,
+        inputs: Optional[Union[List, np.ndarray]] = None,
+        fix_seed: bool = False,
+        callback: Optional[Callable] = None,
```
woodsp-ibm, line 316: Similar comment to the parent class in terms of some of the params, like num_inputs, callback and seed.

Here params just says List rather than List[float] in the typehint.

The order of the parameters here is different than the parent. Yes it has one less but for the others would it not make more sense to keep the same order?

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,337 @@
...
+                (num_inputs, qnn_input_size).
+
+        Raises:
+            QiskitMachineLearningError: If ``len(params) > 1``
+        """
+        params = np.asarray(params)
+        if params is not None and len(params.shape) > 1 and params.shape[0] > 1:
+            raise QiskitMachineLearningError(
+                "The local effective dimension algorithm uses only 1 set of parameters."
+            )
```
woodsp-ibm, line 335: It is helpful to include for uses and when debugging etc to include specific information about the failure in a formatted string in the message.

Though params is checked here there is a setter in the parent right so that this check would not be done later if you call the setter. Perhaps the setter should be overidden here and make the check, since params passed in the constructor is set using the setter it would protect both paths.

releasenotes/notes/add-effective-dimension-class-44fd7661f6ba4b33.yaml
```diff
@@ -0,0 +1,11 @@
+---
+features:
+  - |
+    Introduced two new classes, ``EffectiveDimension`` and ``LocalEffectiveDimension``
```
woodsp-ibm, line 4: We can make these `:class:` references for Sphinx so when people read the release notes they can just click on what will be a hyperlink so they can go to the class and its documentation easily if they are interested in more.

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,337 @@
...
+        dets_div = dets / 2
+        effective_dims = (
+            2
+            * (logsumexp(dets_div, axis=logsum_axis) - np.log(self.num_params))
+            / np.log(n / (2 * np.pi * np.log(n)))
+        )
+
+        return np.squeeze(effective_dims)
+
+    def get_eff_dim(self, n: Union[List[int], np.ndarray, int]) -> Union[np.ndarray, int]:
```
woodsp-ibm, line 278: I'll let other people comment on names. For me though, so things are more self documenting why not `get_effective_dimension`. I think get_effective_dim might be ok too as dim is regularly used for dimension. get_fhat seems obtuse, And why is it get_fishers plural - that seems to be get fisher matrix so seems singular to me - the comment says Fisher Information Matrix so it seemed singular in respect of fisher.

qiskit_machine_learning/algorithms/effective_dimension.py
ElePT: Yes, it would make more sense. Thanks for pointing it out!

test/algorithms/test_effective_dimension.py
```diff
@@ -0,0 +1,210 @@
...
+            else:
+                result = self.result3
+
+        global_ed = EffectiveDimension(
+            qnn=qnn, num_params=num_params, num_inputs=num_inputs, fix_seed=True
+        )
+
+        effdim = global_ed.get_eff_dim(self.n)
+
+        self.assertAlmostEqual(effdim, result, 5)
```
adekusar-drl, line 131: Would it be easier to read 4 independent tests instead of this sophisticated `if` statement?

adekusar-drl (commented):
Thanks for the good job!
By the way, do you use a limit of 80 characters per line?

test/algorithms/test_effective_dimension.py
woodsp-ibm: Or pass the result in with each tuple in the data so its one test with the inputs and expected result passed in.

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,371 @@
...
+            # add dimension to model outputs for broadcasting
+            model_outputs = np.expand_dims(model_outputs, axis=2)
+
+        # get grad-vectors (gradient_k/model_output_k)
+        # multiply by sqrt(model_output) so that the outer product cross term is correct
+        # after Einstein summation
+        gradvectors = np.sqrt(model_outputs) * gradients / model_outputs
+
+        # compute sum of matrices obtained from outer product of grad-vectors
+        fisher = np.einsum("ijk,lji->ikl", gradvectors, gradvectors.T)
```
dlasecki, line 209:
```suggestion
        fisher_information = np.einsum("ijk,lji->ikl", gradvectors, gradvectors.T)
```
Please check other places.

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,371 @@
...
+
+        return fisher
+
+    def get_normalized_fisher(self, normalized_fisher: np.ndarray) -> Tuple[np.ndarray, float]:
+        """
+        This method computes the normalized Fisher Information Matrix (f_hat)
```
dlasecki, line 215:
```suggestion
        This method computes the normalized Fisher Information Matrix
```

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,371 @@
...
+        and extracts its trace.
+        Args:
+            normalized_fisher: The Fisher Information Matrix to be normalized.
+        Returns:
+             normalized_fisher: The normalized Fisher Information Matrix, a numpy array
+                    of size ``(num_inputs, num_weights, num_weights)``
```
dlasecki, line 221:
```suggestion
             normalized_fisher: The normalized Fisher Information Matrix, a numpy array
                 of size ``(num_inputs, num_weights, num_weights)``
```

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,371 @@
...
+        )
+
+        # calculate normalized_normalized_fisher for all the empirical normalized_fisher
+        normalized_fisher = self._model.num_weights * fisher_avg / fisher_trace
+        return normalized_fisher, fisher_trace
+
+    def _get_effective_dimension(
+        self,
+        normalized_fisher: Union[List[float], np.ndarray],
+        n: Union[List[int], np.ndarray, int],
```
dlasecki, line 250: Let's user a better name than `n`.

test/algorithms/test_effective_dimension.py
```diff
@@ -0,0 +1,209 @@
...
+from qiskit_machine_learning.neural_networks import TwoLayerQNN, CircuitQNN
+from qiskit_machine_learning.algorithms.effective_dimension import (
+    EffectiveDimension,
+    LocalEffectiveDimension,
+)
+
+
+@ddt
+class TestEffDim(QiskitMachineLearningTestCase):
+    """Test the Effective Dimension algorithm"""
```
dlasecki, line 36:
```suggestion
    """Test the Effective Dimension algorithm."""
```

test/algorithms/test_effective_dimension.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def setUp(self):
+        super().setUp()
+
+        # fix seeds
```
dlasecki, line 41:
```suggestion
```

test/algorithms/test_effective_dimension.py
```diff
@@ -0,0 +1,209 @@
...
+    @unittest.skipUnless(optionals.HAS_AER, "qiskit-aer is required to run this test")
+    def test_alg_results(self, config):
+        """Test that the algorithm results match the original code's."""
+
+        qnn_name, num_inputs, num_params, result = config
+
+        if qnn_name == "circuit2":
+            qnn = self.circuit_qnn_2
+        else:
+            qnn = self.circuit_qnn_1
```
dlasecki, line 111: How about we get rid of `if` by using a class dict.

test/algorithms/test_effective_dimension.py
```diff
@@ -0,0 +1,209 @@
...
+        ("circuit1", 10, 10, 4.62355184),
+        ("circuit1", 1, 1, 1.39529449),
+        ("circuit1", 10, 1, 4.92825034),
+        ("circuit2", 10, 10, 5.93064171),
+    )
+    @unittest.skipUnless(optionals.HAS_AER, "qiskit-aer is required to run this test")
+    def test_alg_results(self, config):
+        """Test that the algorithm results match the original code's."""
+
+        qnn_name, num_inputs, num_params, result = config
```
dlasecki, line 106:
```suggestion
    @data(
        # num_inputs, num_params
        ("circuit1", 10, 10, 4.62355184),
        ("circuit1", 1, 1, 1.39529449),
        ("circuit1", 10, 1, 4.92825034),
        ("circuit2", 10, 10, 5.93064171),
    )
    @unpack
    @unittest.skipUnless(optionals.HAS_AER, "qiskit-aer is required to run this test")
    def test_alg_results(self, qnn_name, num_inputs, num_params, result):
        """Test that the algorithm results match the original code's."""
```
Unpack also needs an import.

dlasecki (changes requested):

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,370 @@
...
+    @params.setter
+    def params(self, params: Optional[Union[List[float], np.ndarray, float]]) -> None:
+        """Sets network parameters."""
+        if params is not None:
+            self._params = np.asarray(params)
+            self._num_params = len(self._params)
+        else:
+            # random sampling from uniform distribution
+            np.random.seed(self._seed)
+            self._params = np.random.uniform(0, 1, size=(self._num_params, self._model.num_weights))
```
woodsp-ibm, line 116: As I commented before I think this should be using the algorithm_globals random generator as is done elsewhere. I.e. call `algorithm_globals.random.uniform(....)` which is a the random source, which can be seeded for reproducibility and can be used just the same way as above (its backed by a numpy Generator). The code above is using the legacy random function of numpy - if it is really essential to have instances of these on different generators, i.e. using a common core generator as is done elsewhere is not appropriate, then the above code should be changed from legacy. To me I do not see a good reason not to use the algorithm globals generator.

releasenotes/notes/add-effective-dimension-class-44fd7661f6ba4b33.yaml
```diff
@@ -0,0 +1,12 @@
+---
+features:
+  - |
+    Introduced two new classes, :class:`~qiskit_machine_learning.algorithms.EffectiveDimension`
```
woodsp-ibm, line 4: You should update the see algorithms \_\_init\_\_.py file to import these at this level and to have them in the docstring section too there so these new classes are linked into the API Ref. (Your unit test can then potentially import from here as we expect a user more commonly to do - rather than via the module itself as is currently done)

ElePT:
@woodsp-ibm. About https://github.com/Qiskit/qiskit-machine-learning/pull/364#discussion_r846785188, I have switched to `algorithm_globals` but I need to keep explicitly setting the seed when calling the setter method. If I set the seed only once during `init`, it doesn't fix the seed for calls to `algorithm_globals.random` if they are done inside the setter methods (this is why I was doing it, not because I need different seeds for each).

woodsp-ibm:
> but I need to keep explicitly setting the seed when calling the setter method

@ElePT I am not sure why you are setting the seed in the class itself at all. The random generator either gives purely random samples, if nothing (or None) is set, where a system-level random valued seed is used for the generator. If a seed is set (non-None value) then it gives reproducible results. So if any algorithm/code uses the random generator one can choose to set the algorithm globals random seed if you want reproducibility for say testing, otherwise you would not set it (or set it to None) and leave it to some system generated random seed whereby it will now be random and not reproducible.

I do not see why you want to have a seed setter and set it each time within the class itself. This is affecting the shared global generator and will have the effect of making everything reproducible even though it appears one if just seeding a given instance. If chatting out-of-band with you would help work things out here I can do that.

qiskit_machine_learning/algorithms/__init__.py
```diff
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021.
+# (C) Copyright IBM 2021,2022.
```
woodsp-ibm, line 3:
```suggestion
# (C) Copyright IBM 2021, 2022.
```

qiskit_machine_learning/algorithms/__init__.py
```diff
@@ -117,4 +117,6 @@
     "PyTorchDiscriminator",
     "QuantumGenerator",
     "QGAN",
+    "EffectiveDimension",
+    "LocalEffectiveDimension",
```
woodsp-ibm, line 121: These should also be added to the docstring section above so the are linked into the API reference material. Maybe they come under some new section 'Dimensionality Reduction' or something?

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,355 @@
...
+            params: An array of neural network parameters (weights), of shape
+                ``(num_params, num_weights)``.
+            inputs: An array of inputs to the neural network, of shape
+                ``(num_inputs, qnn_input_size)``.
+            num_params: If ``params`` is not provided, the algorithm will
+                randomly sample ``num_params`` parameter sets from a
+                uniform distribution. By default, ``num_params = 1``.
+            num_inputs:  If ``inputs`` is not provided, the algorithm will
+                randomly sample ``num_inputs`` input sets
+                from a normal distribution. By default, ``num_inputs = 1``.
```
woodsp-ibm, line 63: I must admit I find this a bit confusing to sort out. The qnn has `num_inputs` and `num_weights` properties so at times text here is referring to properties of the qnn, other times its a property here say `num_params` and `num_inputs` is ambigious and I think has different meanings. I think here num_params and num_inputs refer to a length/size of arrays that will be randomly created for params and inputs if not supplied.

I wonder whether params could be say `params: l[Union[List[float], np.ndarray, int]] = 1`, where by default, or with an int supplied, will build out a random set. Otherwise via List or array you supply your own. This would remove the need for `num_params` and `num_inputs` here and any confusion that may arise particularly over the latter. Also for params and inputs these are two dimensional array right as in params it an array of different weight values, num_params of them. Is List[float] correct or should it be a List[List[float]] as in I do not see any reshaping if it were a single list. The inputs one says `(num_inputs, qnn_input_size)` which I think is more correctly `(num_inputs, qnn.num_inputs)`

ElePT:
> In terms of location - is this something that we want at the algorithms level, a collection of ml algorithms of different sorts. Or something that perhaps could be in with the neural networks since its really about metrics around these right?

Yes, I added it to the global algorithms folder because I didn't really find a suitable one.... any thoughts @adekusar-drl ?

adekusar-drl:
> > In terms of location - is this something that we want at the algorithms level, a collection of ml algorithms of different sorts. Or something that perhaps could be in with the neural networks since its really about metrics around these right?
> 
> Yes, I added it to the global algorithms folder because I didn't really find a suitable one.... any thoughts @adekusar-drl ? The neural networks one could be an option. Once we agree on one, I can update the `__init__` file as mentioned in the comment above.

I think the `neural_networks` package is a good place. As I understand `ED` deals with neural networks only and it is not a classifier/regressor or any sort of a ML algorithm, but rather an addition to neural networks. Let's move to `neural_networks` if there's no strong objections.

qiskit_machine_learning/algorithms/effective_dimension.py
adekusar-drl: @ElePT please, be careful with variable names. I suggest to use `weights`, `num_weights` for weights, `num_inputs` for the number of features, same as in QNNs, `num_samples` for the number of samples in input data. Thus, `(num_inputs, qnn_input_size)` should become `(num_samples, num_inputs)`. I hope, I did not bring more confusion.

adekusar-drl: Although, I'd prefer to rename `num_inputs` to `num_features` in QNNs. Maybe, next time.

qiskit_machine_learning/algorithms/effective_dimension.py
```diff
@@ -0,0 +1,355 @@
...
+                self._callback(msg)
+
+            back_pass = np.asarray(
+                self._model.backward(input_data=self.inputs, weights=param_set)[1]
+            )
+            t_after_backward = time.time()
+
+            if self._callback is not None:
+                msg = f"iteration {i}, time backward pass: {t_after_backward - t_after_forward}"
+                self._callback(msg)
```
adekusar-drl, line 155: I wonder why you decided to pass a message string instead of numbers to the callback?

test/algorithms/test_effective_dimension.py
```diff
@@ -0,0 +1,179 @@
...
+from qiskit.opflow import PauliSumOp
+from qiskit_machine_learning.neural_networks import TwoLayerQNN, CircuitQNN
+from qiskit_machine_learning.algorithms import (
+    EffectiveDimension,
+    LocalEffectiveDimension,
+)
+
+
+@ddt
+class TestEffDim(QiskitMachineLearningTestCase):
```
adekusar-drl, line 35:
```suggestion
class TestEffectiveDimension(QiskitMachineLearningTestCase):
```

adekusar-drl (commented):
I'd add tests:
- where setters `params` and `inputs` are tested, especially what happens if you pass wrong-shaped arrays.
- callback is used.
see coverage here: https://coveralls.io/builds/48248750/source?filename=qiskit_machine_learning%2Falgorithms%2Feffective_dimension.py

qiskit_machine_learning/algorithms/effective_dimension.py
woodsp-ibm: I personally find it confusing with the qnn haviing num_weights and num_inputs and names here being the same/similar but from what I can see different meanings. From what I see there an array of different values for weights and one for inputs. These have number of different value/samples in them. And are of dimension M x qnn.num_weights and N x qnn.num_inputs. As far as I can see M and N do not need to be the same. Then how to do that, Could it be as I was alluding to just a single field where you either supply your own array or you tell us the size M or N and we build a random array. Whatever we do, as you say we need to be careful with names - if I look at the end of the tutorial (I did not mention it there yet pending any out come from here - it plots 'trained_params` but the text says `We can see from the graph above that the trained weights lead to a higher local effective dimension,` where it talks about weights. The params field here is a set of values for weights right.

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,349 @@
...
+    **References**
+        [1]: Abbas et al., The power of quantum neural networks. `The power of QNNs
+            <https://arxiv.org/pdf/2011.00027.pdf>`_
+    """
+
+    def __init__(
+        self,
+        qnn: NeuralNetwork,
+        params: Optional[Union[List[float], np.ndarray, int]] = 1,
+        inputs: Optional[Union[List[float], np.ndarray, int]] = 1,
```
woodsp-ibm, line 41: These no longer need to be Optional as None is not a valid value any more for these right?

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,349 @@
...
+        self._inputs = None
+        self._num_params = 1
+        self._num_inputs = 1
+        self._model = qnn
+        self._callback = callback
+
+        # Define inputs and parameters
+        self.params = params  # type: ignore
+        # input setter uses self._model
+        self.inputs = inputs  # type: ignore
```
woodsp-ibm, line 73: Is the type ignore here needed? From what I can see the `params` type in the constructor matches the setter (which it should)

ElePT:
@adekusar-drl. In the end, I have decided to go for `num_samples` (for the old `num_inputs`, as suggested), `num_params` and `num_data` (for the old `n`, later refactored to `num_samples`). Hopefully these names are more clear. I have added more checks in the setters, the corresponding unit tests, I have rolled back the callback signature to take in the time values directly, instead of strings (I believe it's better for data collection/analysis). If the sphynx checks don't fail, I believe I have addressed all the comments. If you approve of the new names, I will update the tutorial accordingly.

adekusar-drl:
> @adekusar-drl. In the end, I have decided to go for `num_samples` (for the old `num_inputs`, as suggested), `num_params` and `num_data` (for the old `n`, later refactored to `num_samples`). Hopefully these names are more clear. I have added more checks in the setters, the corresponding unit tests, I have rolled back the callback signature to take in the time values directly, instead of strings (I believe it's better for data collection/analysis). If the sphynx checks don't fail, I believe I have addressed all the comments. If you approve of the new names, I will update the tutorial accordingly.

Can you please tell me what `num_data` stands for now? I understand we are walking in circles, but these names are important to avoid confusions.

ElePT:
> > @adekusar-drl. In the end, I have decided to go for `num_samples` (for the old `num_inputs`, as suggested), `num_params` and `num_data` (for the old `n`, later refactored to `num_samples`). Hopefully these names are more clear. I have added more checks in the setters, the corresponding unit tests, I have rolled back the callback signature to take in the time values directly, instead of strings (I believe it's better for data collection/analysis). If the sphynx checks don't fail, I believe I have addressed all the comments. If you approve of the new names, I will update the tutorial accordingly.
> 
> Can you please tell me what `num_data` stands for now? I understand we are walking in circles, but these names are important to avoid confusions.

Yes, so for the effective dimension you need to provide sets of inputs (now called samples) and parameters for the Monte Carlo sampling, but the actual formula also depends on a parameter `n`, that is referred to in the original paper as "number of data samples", or `n`. This is what I now call `num_data` (and I previously called `num_samples`).

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,354 @@
...
+        """
+        This method computes the model's Monte Carlo sampling for a set of input samples and
+        parameters (weights).
+
+        Returns:
+             grads: QNN gradient vector, result of backward passes, of shape
+                ``(num_samples * num_params, output_size, num_weights)``.
+             outputs: QNN output vector, result of forward passes, of shape
+                ``(num_samples * num_params, output_size)``.
+        """
```
woodsp-ibm, line 145: This is still referring to `parameters` and `num_samples`, `num_params`

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,354 @@
...
+        else:
+            # there is a weird mypy error if we keep the same variable name, so there's 'weights'
+            weights = np.asarray(weight_samples)
+            # additional check to accept 1D arrays
+            if len(weights.shape) < 2:
+                weights = np.expand_dims(weight_samples, 0)
+            if weights.shape[0] != 1 or weights.shape[1] != self._model.num_weights:
+                raise QiskitMachineLearningError(
+                    f"The Local Effective Dimension class expects"
+                    f" a parameter array of shape (1, qnn.num_weights) or (qnn.num_weights)."
```
woodsp-ibm, line 349:
```suggestion
                    f" a weight_samples array of shape (1, qnn.num_weights) or (qnn.num_weights)."
```

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,354 @@
...
+        **References**
+        [1]: Abbas et al., The power of quantum neural networks.
+        `The power of QNNs <https://arxiv.org/pdf/2011.00027.pdf>`__.
+    """
+
+    def __init__(
+        self,
+        qnn: NeuralNetwork,
+        weight_samples: Union[List[float], np.ndarray, int] = 1,
+        input_samples: Union[List[float], np.ndarray, int] = 1,
```
woodsp-ibm, line 42: This I brought up before, but I am not sure its resolved. The type hint here says `List[float]` but should it be List[List[float]] since its really a two dimensional affair i.e. `An array of neural network parameters (weights), of shape               (num_weight_samples, num_weights), or an int`. It could be reshaped from a flat list but I did not see that. 

What I see in the setter for instance is this
```
                len(weight_samples.shape) > 2
                or weight_samples.shape[1] is not self._model.num_weights
```
That checks the dimensions (size of the shape) being > 2 but if 1 will end up failing on the next line as it will index shape[1] which will be out of bounds for a flat list.

I suggested adding unit tests to ensure it ran with plain lists as expected being passed in to make sure of things.

test/neural_networks/test_effective_dimension.py
```diff
@@ -0,0 +1,260 @@
...
+        )
+
+        self.qnns = {"circuit1": circuit_qnn_1, "circuit2": circuit_qnn_2, "opflow": opflow_qnn}
+
+        # define sample numbers
+        self.n_list = [5000, 8000, 10000, 40000, 60000, 100000, 150000, 200000, 500000, 1000000]
+        self.n = 5000
+
+    @data(
+        # qnn_name, num_inputs, num_params, result
```
woodsp-ibm, line 99:
```suggestion
        # qnn_name, num_inputs, num_weights, result
```
I think thats more correct than `params` now right. Its fairly minor being a comment in test case but may as well make it better for future readers

test/neural_networks/test_effective_dimension.py
```diff
@@ -0,0 +1,260 @@
...
+        global_ed2 = EffectiveDimension(
+            qnn=qnn2,
+            weight_samples=num_weight_samples,
+            input_samples=num_input_samples,
+        )
+
+        effdim1 = global_ed1.get_effective_dimension(self.n)
+        effdim2 = global_ed2.get_effective_dimension(self.n)
+
+        self.assertAlmostEqual(effdim1, effdim2, 5)
```
woodsp-ibm, line 137: This asserts two results are the same - in my mind it would be good to have an expected reference value at least one of these was compared against to ensure that they both don't go bad together and we never notice. I see this is done in other tests too, but I will just comment here.

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,354 @@
...
+from qiskit.utils import algorithm_globals
+from qiskit_machine_learning import QiskitMachineLearningError
+
+from .opflow_qnn import OpflowQNN
+from .neural_network import NeuralNetwork
+
+
+class EffectiveDimension:
+    """
+    This class computes the global effective dimension for Qiskit
```
woodsp-ibm, line 29:
```suggestion
    This class computes the global effective dimension for a Qiskit
```
This is how it currently reads
> This class computes the global effective dimension for Qiskit NeuralNetwork following the definition used in [1]

So just adding the `a` so it reads a little easier
>This class computes the global effective dimension for a Qiskit NeuralNetwork fo...

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,354 @@
...
+
+        # step 4: compute eff. dim
+        effective_dimensions = self._get_effective_dimension(normalized_fisher, dataset_size)
+
+        return effective_dimensions
+
+
+class LocalEffectiveDimension(EffectiveDimension):
+    """
+    This class computes the local effective dimension for Qiskit
```
woodsp-ibm, line 314:
```suggestion
    This class computes the local effective dimension for a Qiskit
```

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,354 @@
...
+        self._num_input_samples = 1
+        self._model = qnn
+        self._callback = callback
+
+        # Define weight samples and input samples
+        self.weight_samples = weight_samples  # type: ignore
+        # input setter uses self._model
+        self.input_samples = input_samples  # type: ignore
+
+    def num_weights(self) -> int:
```
woodsp-ibm, line 78: This ends up being a method i.e. you would call it as `ed.num_weights()` versus accessing it as property on the neural network i.e.` self._model.num_weights` Would we prefer a property getter here if this functionality is needed?

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,354 @@
...
+        """Returns the dimension of the model according to the definition from [1]."""
+        return self._model.num_weights
```
woodsp-ibm, line 80: `Returns the dimension of the model...`  So far we have given a NeuralNetwork as a qnn into the constructor. Yes it was stored as _model but somehow I imagine the term model to be something I have to read the reference and its not something we refer qnn to be in the constructor docstring. This is also call num_weights not dim or dimension. I think maybe, something more like this perhaps
`Returns the number of weights of the supplied neural network. This is the dimension of the model according to the definition from [1].` where we explain it in terms of the code here and state the correspondence to the paper rather than the latter that being the main thing.  Since we passed in qnn and ought to know num_weights from that, this indirect access to it - is it really useful?

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,354 @@
...
+
+    @property
+    def weight_samples(self) -> np.ndarray:
+        """Returns network parameters (weights)."""
+        return self._weight_samples
+
+    @weight_samples.setter
+    def weight_samples(self, weight_samples: Union[List[float], np.ndarray, int]) -> None:
+        """Sets network parameters (weights)."""
```
woodsp-ibm, line 89: Since the input samples setter/getter is more direct and says `Sets network input samples.` Would this and the getter above be more direct as `Sets network weights samples.` - why do we need to use the term parameters here and then have to qualify it with (weights) in order for it to presumably be understood in context here.

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,354 @@
...
+        if isinstance(weight_samples, int):
+            # random sampling from uniform distribution
+            self._weight_samples = algorithm_globals.random.uniform(
+                0, 1, size=(weight_samples, self._model.num_weights)
+            )
+        else:
+            weight_samples = np.asarray(weight_samples)
+            if (
+                len(weight_samples.shape) > 2
+                or weight_samples.shape[1] is not self._model.num_weights
```
woodsp-ibm, line 99: `is not` - do we mean `!=` here

test/neural_networks/test_effective_dimension.py
adekusar-drl: Added an explicit assert.

test/neural_networks/test_effective_dimension.py
```diff
@@ -0,0 +1,260 @@
...
+            qnn=qnn,
+            weight_samples=weights,
+            input_samples=inputs,
+        )
+
+        effdim1 = global_ed1.get_effective_dimension(self.n_list)
+        effdim2 = global_ed2.get_effective_dimension(self.n_list)
+
+        print(effdim1)
+        print(effdim2)
```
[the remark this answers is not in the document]
adekusar-drl: Yes, I forgot to remove prints while was working on tests.

qiskit_machine_learning/neural_networks/effective_dimension.py
adekusar-drl: I'm not sure it is required. I removed this method.

qiskit_machine_learning/neural_networks/effective_dimension.py
adekusar-drl: I believe `num_weights` can be easily retrieved from qnn, so I don't see a real need for this method.

qiskit_machine_learning/neural_networks/effective_dimension.py
adekusar-drl: Updated.

qiskit_machine_learning/neural_networks/effective_dimension.py
adekusar-drl: Oh, good point, of course it must be `!=`.

qiskit_machine_learning/neural_networks/effective_dimension.py
adekusar-drl: While you are absolutely right, I'm not sure I want to keep `List` in type hints at all. They tend to cause many problems, especially when multidimensional arrays are required. I'd prefer to remove support for `List` completely in favor of plain `np.ndarray`. I think I know where this type hint originates from. There's the same problem in `NeuralNetwork`, same list type hint and it is also incorrect, unfortunately. In `NeuralNetwork` a multidimensional array is expected. Usually it is of shape `(num_samples, num_features)`. 

Therefore, `input_samples` may be a single value, a plain list or a list of list at least. I can extend type hints, but prefer to remove support for lists and this is a good chance to start it off. There's an old issue to consider as well: https://github.com/Qiskit/qiskit-machine-learning/issues/159.

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,347 @@
...
+    def weight_samples(self, weight_samples: Union[List[float], np.ndarray, int]) -> None:
+        """Sets network weight samples."""
+        if isinstance(weight_samples, int):
+            # random sampling from uniform distribution
+            self._weight_samples = algorithm_globals.random.uniform(
+                0, 1, size=(weight_samples, self._model.num_weights)
+            )
+        else:
+            weight_samples = np.asarray(weight_samples)
+            if len(weight_samples.shape) > 2 or weight_samples.shape[1] != self._model.num_weights:
```
woodsp-ibm, line 93:
```suggestion
            if len(weight_samples.shape) != 2 or weight_samples.shape[1] != self._model.num_weights:
```
I think this should be better as != 2 rather than > 2 right. Since an single dimension array would go on to the test of shape[1] and will fail since the shape will only have a single entry.

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -0,0 +1,347 @@
...
+    def input_samples(self, input_samples: Union[List[float], np.ndarray, int]) -> None:
+        """Sets network input samples."""
+        if isinstance(input_samples, int):
+            # random sampling from normal distribution
+            self._input_samples = algorithm_globals.random.normal(
+                0, 1, size=(input_samples, self._model.num_inputs)
+            )
+        else:
+            input_samples = np.asarray(input_samples)
+            if len(input_samples.shape) > 2 or input_samples.shape[1] != self._model.num_inputs:
```
woodsp-ibm, line 118:
```suggestion
            if len(input_samples.shape) != 2 or input_samples.shape[1] != self._model.num_inputs:
```
Same as the weight_samples comment above

adekusar-drl: OK, I see the point. Then, it's easier to restrict to 2D (other shapes are pretty rare and can be converted to 2D anyway) rather than changing the algorithm with unknown benefits.

qiskit_machine_learning/neural_networks/effective_dimension.py
adekusar-drl: OK, I removed lists from the type hints.

qiskit_machine_learning/neural_networks/effective_dimension.py
adekusar-drl: Updated

woodsp-ibm (approved):

PR #365, opened 2022-04-04
Quantum Kernel fix for duplicate input vectors with QASM simulator

### LINKED ISSUE #336: Exception while evaluating QuantumKernel on a QASM simulator when duplicate entries in the dataset
### Environment

- **Qiskit Machine Learning version**: latest sources
- **Python version**: all
- **Operating system**: all
 

### What is happening?

Exception is raised when a quantum kernel is evaluated for a dataset with duplicate entries on a qasm simulator:
```
FAILURE: Can not get job id, Resubmit the qobj to get job id. Error: max() arg is an empty sequence 
.... This line appears many times ...
FAILURE: Can not get job id, Resubmit the qobj to get job id. Error: max() arg is an empty sequence 
  File ".../qiskit-machine-learning/_sandbox/qk_bug.py", line 14, in <module>
    qasm_kernel.evaluate(x_vec=train_features)
  File ".../qiskit-machine-learning/qiskit_machine_learning/kernels/quantum_kernel.py", line 535, in evaluate
    results = self._quantum_instance.execute(circuits, had_transpiled=True)
  File ".../envs/dev-qml/lib/site-packages/qiskit/utils/quantum_instance.py", line 811, in execute
    if circuit_job
  File ".../envs/dev-qml/lib/site-packages/qiskit/utils/run_circuits.py", line 530, in run_circuits
    max_job_retries=max_job_retries,
  File ".../envs/dev-qml/lib/site-packages/qiskit/utils/run_circuits.py", line 710, in _safe_submit_circuits
    raise QiskitError("Max retry limit reached. Failed to submit the qobj correctly")
qiskit.exceptions.QiskitError: 'Max retry limit reached. Failed to submit the qobj correctly'
```

Run the script:

```python
import numpy as np
from qiskit import Aer

from qiskit_machine_learning.kernels import QuantumKernel

qasm_backend = QuantumInstance(Aer.get_backend("qasm_simulator"))   # statevector_simulator

feature_map = ZZFeatureMap(feature_dimension=2)
qasm_kernel = QuantumKernel(feature_map=feature_map, quantum_instance=qasm_backend)

train_features = np.array([[1, 2], [1, 2]])
qasm_kernel.evaluate(x_vec=train_features)
```

If you replace `qasm_simulator` with `statevector_simulator` no exception is raised.

### What should happen?

No exception should be raised.

### Any suggestions?

This problem happens due to this check: https://github.com/Qiskit/qiskit-machine-learning/blob/main/qiskit_machine_learning/kernels/quantum_kernel.py#L524-L526

There's a check if two samples are identical and if this is true these samples are not used for quantum kernel estimation and since in the original dataset there are just two identical samples, nothing can be estimated and an empty list is passed to `QuantumInstance` for execution and it fails.

May be a warning should be raised.

gentinettagian:
Fixes #336

### Summary
The ``QuantumKernel`` class raised an exception when evaluating the matrix with duplicate entries using a qasm simulator. Whenever ``x = y`` in ``k(x,y)`` appeared, the code would skip the calculation and not provide a result. This has been fixed by hard coding ``k(x,x) = 1``, as the overlap of a quantum state with itself is always equal to ``1``.


woodsp-ibm:
Can you update your black install as the latest version no longer requires the extra space around the ** operator. If you look at recent history of one of these other changed files you will see that extra space was there until recently when Qiskit updated which  version of black was being used. This is why build is failing with black.

pbark:
Hi @gentinettagian , the fix of the issue is right. I have one conceptual concern remaining here. In the presence of noise the calculation of the overlaps will suffer from the measurement problem that will introduce some noise in the kernel matrix values. In the case that we introduce 1 for identical datapoints (as you rightly did) we will not introduce at all noise and the kernel may have 'spike' on the diagonal elements. We notice in many hardware runs with increased number if qubits that the off diagonal values are significantly lower than the statevector predicted ones and in this case removing the check for identical elements and calculating the actual overlap would be of benefit. The cost will be linear for the training kernel matrix and the kernel test matrix (should be `N` and `<<N` respectively). I think that removing the check for identical elements would be a possible solution, although your solution also addresses the problem.

gentinettagian:
> Hi @gentinettagian , the fix of the issue is right. I have one conceptual concern remaining here. In the presence of noise the calculation of the overlaps will suffer from the measurement problem that will introduce some noise in the kernel matrix values. In the case that we introduce 1 for identical datapoints (as you rightly did) we will not introduce at all noise and the kernel may have 'spike' on the diagonal elements. We notice in many hardware runs with increased number if qubits that the off diagonal values are significantly lower than the statevector predicted ones and in this case removing the check for identical elements and calculating the actual overlap would be of benefit. The cost will be linear for the training kernel matrix and the kernel test matrix (should be `N` and `<<N` respectively). I think that removing the check for identical elements would be a possible solution, although your solution also addresses the problem.

The issue I have with this is that we add additional computational cost (even if it's not much) just to get a noisy result and generally I would think that an exact result is superior. But if in practice a noisy kernel entry is desirable then we could just remove the check, as you suggest.

PR #369, opened 2022-04-08
Average objective functions

### LINKED ISSUE #327: Review implementation of the loss and objective functions

Review implementation of the loss and objective functions and verify their correctness. At least, objective values are not scaled as `1/N` and this causes questions why loss/objective values are large.

adekusar-drl:
### Summary
Objective functions constructed by the neural network classifiers and regressors now include an averaging factor that is evaluated as `1 / number_of_samples`. Computed averaged objective values are passed to a user specified callback if any. Users may notice a dramatic decrease in the objective values in their callbacks. This is due to this averaging factor.
Closes #327.

### Details and comments
Updated:
- Objective functions
- Increased number of samples from 5 to 6 in the tests to improve test convergence and stability.

### DIFF
--- a/qiskit_machine_learning/algorithms/objective_functions.py
+++ b/qiskit_machine_learning/algorithms/objective_functions.py
@@ -52,6 +52,7 @@ class ObjectiveFunction:
         """
         super().__init__()
         self._X = X
+        self._num_samples = X.shape[0]
         self._y = y
         self._neural_network = neural_network
         self._loss = loss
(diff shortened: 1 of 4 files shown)


woodsp-ibm:
Is there a defined benefit to this as well. I can perhaps see that the values can remain more or less the same if you change the number of samples a little rather than as things were before. This presumably will only change the scale of the plots in the 02_xxx tutorial - the curve will presumably be the same.

woodsp-ibm (approved):

Issue #371, opened 2022-04-08
Hardcoded test size in breast cancer and digits dataset


bopardikarsoham:
### Environment

-

### What is happening?

test_size being hardcoded to 0.3 in [#Line36](https://github.com/Qiskit/qiskit-machine-learning/blob/main/qiskit_machine_learning/datasets/breast_cancer.py#L36) is not giving the user the flexibility to set the desirable train-test-split ratio.

### How can we reproduce the issue?

The variable `test_size` is already present as an input in the function's parameters. As a result, replacing 0.3 with test_size resolves the problem. 
`sample_train, sample_test, label_train, label_test = train_test_split(
        data, target, test_size=test_size, random_state=12
    )`

### What should happen?

-

### Any suggestions?

I also believe that the random_state variable should be defined as a function parameter. So that the user has the option of randomizing the samplings.

adekusar-drl:
The datasets are deprecated and will be removed in the release after the upcoming `0.4` release. 

The `test_size` variable is used to subset test data here: https://github.com/Qiskit/qiskit-machine-learning/blob/main/qiskit_machine_learning/datasets/breast_cancer.py#L61

adekusar-drl:
I close the issue since these datasets will be removed soon.

PR #372, opened 2022-04-08
Fix hardcoded test size in breast cancer and digits dataset

bopardikarsoham:
As per Issue #371 , I've done the appropriate modifications.

### DIFF
--- a/qiskit_machine_learning/datasets/breast_cancer.py
+++ b/qiskit_machine_learning/datasets/breast_cancer.py
@@ -33,7 +33,7 @@ def breast_cancer(training_size, test_size, n, plot_data=False, one_hot=True):
     class_labels = [r"A", r"B"]
     data, target = datasets.load_breast_cancer(return_X_y=True)
     sample_train, sample_test, label_train, label_test = train_test_split(
-        data, target, test_size=0.3, random_state=12
+        data, target, test_size=test_size, random_state=12
     )
 
     # Now we standardize for gaussian around 0 with unit variance
--- a/qiskit_machine_learning/datasets/digits.py
+++ b/qiskit_machine_learning/datasets/digits.py
@@ -34,7 +34,7 @@ def digits(training_size, test_size, n, plot_data=False, one_hot=True):
     data = datasets.load_digits()
     # pylint: disable=no-member
     sample_train, sample_test, label_train, label_test = train_test_split(
-        data.data, data.target, test_size=0.3, random_state=22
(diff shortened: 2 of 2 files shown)


adekusar-drl:
Please take a look at my reply in the issue. This is wrong usage of the `test_size` parameter. 

Nevertheless, if you want to fix this, please take a look at the contributing guide here https://github.com/Qiskit/qiskit-machine-learning/blob/main/CONTRIBUTING.md and consider similar changes to other datasets.

bopardikarsoham:
Yes, I just saw your response to the issue and saw where I was going wrong. Really sorry for the inconvenience.

adekusar-drl:
> Yes, I just saw your response to the issue and saw where I was going wrong. Really sorry for the inconvenience.

@bopardikarsoham my apologies, you are right. In the `wine` and `iris` datasets the `test_size` is applied in the same way as you mentioned in the issue. So, the fix looks correct.

bopardikarsoham:
I also had a query about why test_size is not set to 0.3 in datasets like wine and iris, as shown at https://github.com/Qiskit/qiskit-machine-learning/blob/main/qiskit_machine_learning/datasets/iris.py#L36.

bopardikarsoham:
> > Yes, I just saw your response to the issue and saw where I was going wrong. Really sorry for the inconvenience.
> 
> @bopardikarsoham my apologies, you are right. In the `wine` and `iris` datasets the `test_size` is applied in the same way as you mentioned in the issue. So, the fix looks correct.

Yes, l saw that too. That's why I made the change.

adekusar-drl:
> I also had a query about why test_size is not set to 0.3 in datasets like wine and iris, as shown at https://github.com/Qiskit/qiskit-machine-learning/blob/main/qiskit_machine_learning/datasets/iris.py#L36.

I'd say either `test_size` parameter should be set to `0.3` or to the passed `test_size` value everywhere.

woodsp-ibm:
I am not sure if this is something we would want to do at this point. The dataset is initially split such that 0.3 of it is kept for test. Then out of the data that is split 0.7/0.3 ration,  training size and test_size amounts are taken. Maybe the ratio of 0.7 to 0.3 as per the dataset split is different than training_size/test_size split such that its not so great. Either way I think this will end up as breaking change and with the datasets being deprecated for removal I think it may be better to live with it. A user can directly use breast cancer etc if the behavior here does not work for them.

adekusar-drl:
I close this PR as the datasets are deprecated.

Issue #342, opened 2022-03-09
Add Effective Dimension Tutorial


ElePT:
### What should we add?

The goal of this issue is to track the development of a tutorial for the effective dimension class and runtime program, as described in epic https://github.com/Qiskit/qiskit-machine-learning/issues/318.

PR #373, opened 2022-04-08
Effective Dimension Tutorial

ElePT:
### Summary
This PR addresses #342, with a tutorial that shows the relevance and usage of the effective dimension class.

### Details and comments
This PR depends on #364.


adekusar-drl (commented):
I had a quick look and have a few comments:
- Rename the file `09_*`, to `10_*` since I had `09` in another PR :)
- Could it be better merge import cells into one?
- Add some comments between code cells to be sure we have not only code, but some explanations as well?
- Fix a random seed to be sure the results are reproducible.
- Think of a thumbnail image/figure/plot for the tutorial.

woodsp-ibm:
In section 3 you compute `d` in the first cell there and print it. You then compute it again in the next cell. And in a subsequent cell print it again where it seems unneeded and one needs to look back to see what this number 40 means as there is no description in the print - in the first cell its not of course needed since the code is right there.

Also the final section in the notebook seems to not have run properly showing an error in unpack.

ElePT:
I believe I have addressed all of your comments @adekusar-drl, @woodsp-ibm, and I have tried to use a more interesting example for to demo how to use the local effective dimension.

adekusar-drl:
@ElePT you have a conflict in `.pylintdict`, could you please resolve it?

.pylintdict
```diff
@@ -84,6 +84,7 @@ kwarg
 kwargs
 langle
 macos
+main
```
woodsp-ibm, line 87: `main` is not in the base dictionary - given that's a very common word that's hard to believe

docs/tutorials/10_effective_dimension.ipynb
```diff
@@ -0,0 +1,763 @@
...
+    }
+   ],
+   "source": [
+    "num_qubits = 3\n",
+    "# create a feature map\n",
+    "feature_map = ZFeatureMap(feature_dimension=num_qubits, reps=1)\n",
+    "# create a variational circuit\n",
+    "ansatz = RealAmplitudes(num_qubits, reps=1)\n",
+    "\n",
+    "# create quantum circuit\n",
```
woodsp-ibm, line 133: Rather than this more abstract comment, would it be better to sat combine the feature map and ansatz into a in single circuit.

Now this looks like a `two layer qnn`. Is there some rationale why we chose to do that as a CircuitQNN here rather than use the that. If so would this benefit from some explanation here. I assume this can work fine with a TwoLayerQNN.

docs/tutorials/10_effective_dimension.ipynb
```diff
@@ -0,0 +1,763 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 3.1 Define the Problem\n",
+    "\n",
+    "In order to define the problem, we need a series of sets of input samples (`num_input_samples`) and weights (`num_weight_samples`), as well as the total number of data samples available in a dataset (`n`). The `input_samples` and `weight_samples` are set in the class constructor, while the number of data samples is given during the call to the effective dimension computation to be able to test and compare how this measure changes with different dataset sizes.\n",
```
woodsp-ibm, line 196: > (`num_input_samples`) and weights (`num_weight_samples`)

Here there are what look like variables/properties e..g `num_input_samples` that might be accessed. These don't exist though right its just the number of input samples ie the first dimension of the input samples array etc.

docs/tutorials/10_effective_dimension.ipynb
```diff
@@ -0,0 +1,763 @@
...
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "## 4. Local Effective Dimension Example\n",
+    "As explained in the introduction, the local effective dimension algorithm only uses **one** set of weights. The `LocalEffectiveDimension` class enforces this constraint to ensure that these calculations are conceptually separate, but the rest of the implementation is shared with the `EffectiveDimension` class.\n",
+    "\n",
+    "Let's see how the `LocalEffectiveDimension` changes when we train a QNN for classification.\n",
+    "\n",
+    "First, we create a classification dataset, similarly to what you may see in other tutorials."
```
woodsp-ibm, line 470: Does this `similarly to what you may see in other tutorials` add any value here. There is no specific link where perhaps I might gain further insight; or is it trying to say its not uncommon to 'make up' i.e. create a dataset, others do it

docs/tutorials/10_effective_dimension.ipynb
```diff
@@ -0,0 +1,763 @@
...
+    "        plt.plot(x[0], x[1], \"go\")\n",
+    "plt.plot([-1, 1], [1, -1], \"--\", color=\"black\")\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The next step is to create a QNN, an instance of `TwoLayerQNN` in our case. Since we pass only the number of inputs, the network will continue with the default values for feature map and ansatz."
```
woodsp-ibm, line 516: I guess this is done around TwoLayerQNN which maybe is why the former above is CircuitQNN to show that can be done too?

docs/tutorials/10_effective_dimension.ipynb
```diff
@@ -0,0 +1,763 @@
...
+   "cell_type": "code",
+   "execution_count": 18,
+   "metadata": {
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "trained_weights = circuit_classifier._fit_result.x\n",
```
woodsp-ibm, line 614: I guess this is part of the rationale for #385 to avoid accessing private field

.pylintdict
adekusar-drl: > `main` is not in the base dictionary - given that's a very common word that's hard to believe
oh, wow, removed.

docs/tutorials/10_effective_dimension.ipynb
adekusar-drl: > I guess this is part of the rationale for #385 to avoid accessing private field

This is correct.

docs/tutorials/10_effective_dimension.ipynb
ElePT: Yes, this was the idea, to have examples with both.

docs/tutorials/10_effective_dimension.ipynb
ElePT: Yes, this is basically a TwoLayerQNN, and the only reason why it's shown with a Circuit QNN is to have both examples. I can add a remark pointing this out.

docs/tutorials/10_effective_dimension.ipynb
ElePT: True, they were accessible at the beginning but not any more. Will change.

docs/tutorials/10_effective_dimension.ipynb
```diff
@@ -0,0 +1,692 @@
...
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Iteration: 0, forward time: 0.03782296180725098, backward time: 0.36411023139953613\n",
+      "Iteration: 1, forward time: 0.01554107666015625, backward time: 0.45325207710266113\n",
+      "Iteration: 2, forward time: 0.01328730583190918, backward time: 0.2095649242401123\n",
+      "Iteration: 3, forward time: 0.014077186584472656, backward time: 0.25399303436279297\n",
```
woodsp-ibm, line 310: I am wondering about the value of this effective dimension callback. In other callbacks elsewhere we have things like current expectation value, current loss value as these are computed at each step that these might be used by the user to track/plot algorithm progression etc. Other information that the user might use is also passed, such as current job for execution, that they can then user other calls to look up the job and see more detail etc. Timing information like this I would at first have thought more think about logging, so if they wanted it they would enable logging . I guess  each "iteration" here, where there are 10 of them,  relates to the 10 samples, so its more an index into the sample being processing in order? 

Also as it stands I see the parameters of the callback are not documented in the algorithm constructor - I looked there to see what it said about this.

If its felt that this is desirable to keep, I think in this example I would format the times to say limit them to say 3 decimal places rather than have so many decimal places that I imagine mean little as I imagine these times vary quite a bit from run to run depending what else is going on in the machine running these.

ElePT: The reasoning behind the callback is that iteration times are interesting for monitoring the performance, particularly when ran inside a runtime program, and logging cannot be accessed in runtime programs. 

I see your point with using the name "iterations", each iteration corresponds to a sampled set of weights (there are 10 samples in this examples), so maybe I can substitute "iteration" with "weights sample". I will also apply the limit on decimal places.

woodsp-ibm: In these other callbacks normally the user can do something like tune some hyperparameters, change the anstaz and/or optimizer and can see how that affects say the convergence via the callback.

In the case of timing, aside from perhaps being of interest, I don't think there is anything you can do right. If you are computing the effective dimension of a given network it is going to be whatever it is whether local or on the runtime.

If its something you found useful and its something users would want then ok, but...

> Also as it stands I see the parameters of the callback are not documented in the algorithm constructor - I looked there to see what it said about this.

But I think the parameters ought to be documented as to what they are on the class, and that change could be added to this PR. What it has at present is 

```
        callback: Optional[Callable[[int, float, float], None]] = None,
...        
        callback: A callback function for the Monte Carlo sampling.
```
that does not spell things out. If it can say the int is the current weight sample and the floats are the forward and backward times on seconds for evaluating that sample   - or whatever the right wording is to describe this.

ElePT: @woodsp-ibm It's true that the callback itself is not used for anything practical, and while I did find it useful in certain experiments, I now also see how it makes sense to just turn it into a logging message (it adds noise to the documentation). I can later look into using a callback for the runtime program if I find it necessary, but I agree that the class code is cleaner without it. Sorry for going in circles for so long about this (and thanks for you patience). I now have removed the callback and turned into a debug message. I hope it makes sense now!

qiskit_machine_learning/neural_networks/effective_dimension.py
```diff
@@ -336,6 +340,7 @@ def weight_samples(self, weight_samples: Union[np.ndarray, int]) -> None:
             # there is a weird mypy error if we keep the same variable name, so there's 'weights'
             weights = np.asarray(weight_samples)
             # additional check to accept 1D arrays
+            # additional check to accept 1D arrays
```
woodsp-ibm, line 343: Did you accidentally double-up this comment?

ElePT: yes, that is a typo. Fixing it now.

docs/tutorials/10_effective_dimension.ipynb
```diff
@@ -0,0 +1,770 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 2. The Effective Dimension Algorithm\n",
+    "\n",
+    "Both the global and local effective dimension algorithms use the Fisher Information matrix to provide a measure of complexity. The details on how this matrix is calculated are provided in the [reference paper](https://arxiv.org/pdf/2011.00027.pdf), but in general terms, this matrix captures how sensitive a neural network's output is to changes in the network's parameter space.\n",
+    "\n",
+    "In particular, this algorithm follows 4 main steps:\n",
+    "\n",
+    "1. **Monte Carlo simulation:** the forward and backward passes (gradients) of the neural network are computed for each pair of input samples-parameter sets.\n",
```
woodsp-ibm, line 39: `are computed for each pair of input samples-parameter sets` should we change this to something like `each pair or input and weight samples` so things are more clear based on how we ended up calling things.

docs/tutorials/10_effective_dimension.ipynb
```diff
@@ -0,0 +1,770 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Effective Dimension of Qiskit Neural Networks\n",
+    "In this tutorial, we will take advantage of the `EffectiveDimension` and `LocalEffectiveDimension` classes to evaluate the power of Quantum Neural Network models. These are metrics based on information geometry that connect to notions such as trainability, expressibility or ability to generalize.\n",
```
woodsp-ibm, line 8: Here we introduce these two classes, one is simply EffectiveDimension and the other LocalEffectiveDimension. In subsequent text is has "**global** effective dimension" and "**local** effective dimension" with the aspects highlighted. Now the local one seems to match to the LocalEffectiveDimension and I guess we leave it to readers to infer that the global one is simply the EffectiveDimension. Maybe we could explicitly make the connection in the text between the concept i.e; global effective dimension and the class implementing it i.e. EffectiveDimension.

docs/tutorials/10_effective_dimension.ipynb
```diff
@@ -0,0 +1,770 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 3.1 Define the Problem\n",
+    "\n",
+    "In order to define the problem, we need a series of sets of input samples and weights, as well as the total number of data samples available in a dataset. The `input_samples` and `weight_samples` are set in the class constructor, while the number of data samples is given during the call to the effective dimension computation to be able to test and compare how this measure changes with different dataset sizes."
```
woodsp-ibm, line 198: What does it mean here "to define the problem". It seems the problem is that we want to measure the (local) effective dimension of the network and to do that we need input+weight samples as data for the algorithm. It defines some sort of test space perhaps the set of samples, but does it define the problem as such?

woodsp-ibm (commented):
I figured I would read through the tutorial it one last time

docs/tutorials/10_effective_dimension.ipynb
ElePT: It's true, I have changed to "3.2 Set up Effective Dimension calculation". I have also added titles to the local effective dimension example to make it more consistent with the global one.

woodsp-ibm (approved):
All looks good to me now. Thanks for sorting things out from all my comments.

PR #374, opened 2022-04-11
Add input validation to VQC and refactor all unit tests

### LINKED ISSUE #320: Allow categorical encoding in VQC

Currently, VQC only accepts one-hot encoded input. It might be beneficial to accept categorical labels (and perhaps features) to limit friction for new adopters used to scikit-learn—ideally, they could just replace their classical code with the quantum equivalent without explicitly performing the encoding themselves. The only downside is this does contribute to code bloat.

This was raised previously in #62 for VQC and NeuralNetworkClassifier. It looks like this capacity was incorporated into the parent class in #143, but, as far as I can tell, VQC is still limited to one-hot encoding.

There might be a good reason for this, but, if not, we should ensure that VQC inherits this behaviour properly.

### LINKED ISSUE #322: Warn when fitting new data with a different number of classes

One can train VQC in batches using the warm_start=True flag. If one then provides data with a different number of total classes, the code will process the data using the initial_point reached in the previous batch without complaint.

While certainly not critical, it might be friendly to warn the user that the data they are providing probably invalidates the parameters reached during the previous fit.

### LINKED ISSUE #323: VQC input validation

Multi-label classification is not supported by VQC; however, the code will not raise an error when it is provided. It might be beneficial to users to raise an error in such cases, either due to unclean data or due to them mistakenly thinking that multi-label input will be comprehended by the model.

### LINKED ISSUE #325: Simplify VQC unit tests

Currently, the VQC unit tests use the DDT library to pass a combination of strings to tests in order to check execution with a range of optimizers and quantum instances. Rather than using string tags, it is possible to pass this data directly, which may minimize and, perhaps, clarify the necessary code for tests.

declanmillar:
### Summary

Raise an error when warm starting using targets with a different number of classes to the previous dataset. Closes #322.

Raise an error when a user attempts multi-label classification with VQC, which is not supported. Closes #323.

Refactors all VQC unit tests to (hopefully) make them more digestible, shorter, simpler, and more comprehensive. Closes #325.

### Details and comments

- [x] Error when fitting new data with a different number of classes.
- [x] Add unit tests to test a warning/error is raised.
- [x] Validate VQC input. e.g., ensure input isn't multi-label (only multi-class)
- [x] Ensure an error is raised for bad input using unit tests.
- [x] Simplify/shorten all VQC unit tests.
- [x] Add release note.


qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -159,25 +168,118 @@ def num_qubits(self) -> int:
...
+        Returns:
+            The suitably transformed and validated target values.
+        """
+        targets = self._label_transformer.fit_transform(y)
+        if len(targets[0]) == 1:
+            self._one_hot = False
+            targets = targets.flatten()
+        else:
+            self._one_hot = True
+            self._validate_one_hot_targets(targets)
```
adekusar-drl, line 239: So, `self._one_hot` becomes a part of the object state in `VQC` since it is modifiable. While `one_hot` is a read-only parameter in `NeuralNetworkClassifier`. Is it possible to move the label encoding functions from `VQC` to `NeuralNetworkClassifier` then?

test/algorithms/classifiers/resources/test_vqc_binary_data.json
```diff
@@ -0,0 +1,19 @@
+{
```
adekusar-drl, line 1: I'd move the resources to the package with the test scripts.

test/algorithms/classifiers/test_vqc.py
```diff
@@ -14,184 +14,141 @@
...
-        classifier.fit(X, y)
+        initial_point = np.array([0.5] * ansatz.num_parameters)
 
-        # score
-        score = classifier.score(X, y)
-        self.assertGreater(score, 0.5)
+        x = np.asarray(self.binary_data["X"])
+
+        for y in ["y", "y_one_hot", "y_str"]:
+            with self.subTest(f"{y}"):
```
adekusar-drl, line 123: Considering that you have @data, I'd suggest to move the `for` loop to @data and to avoid long declarations you can use `itertools.product` like here:
```python
    @idata(itertools.product(OPTIMIZERS, L1L2_ERRORS, QUANTUM_INSTANCES, CALLBACKS))
    @unpack
    def test_classifier_with_opflow_qnn(self, opt, loss, q_i, cb_flag):
```

adekusar-drl (commented):
Thanks for the good job! What makes me worry is that I don't see how you transform predicted labels to the original format. May be I missed that somewhere in the code. The rest, I guess, minor things.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -175,3 +172,13 @@ def _validate_input(self, X: np.ndarray, y: np.ndarray = None) -> Tuple[np.ndarr
...
+
+    def _validate_binary_labels(self, y: np.ndarray) -> np.ndarray:
+        if not self._has_valid_binary_labels(y):
+            raise QiskitMachineLearningError(
+                f"Current settings only applicable to binary classification! Got labels: {y}"
+            )
+        return y
+
+    def _has_valid_binary_labels(self, y: np.ndarray) -> bool:
+        return len(y.shape) == 1 and len(np.unique(y)) == 2
```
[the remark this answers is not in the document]
declanmillar: Yeah, I just wrapped it in a function to make the intention of the `if` statement clear.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -132,13 +136,18 @@ def __init__(
...
-            one_hot=True,
             optimizer=optimizer,
             warm_start=warm_start,
             initial_point=initial_point,
             callback=callback,
         )
 
+        # For checking the number of classes is the same for different batches.
+        self.warm_start: bool = warm_start
+        self.num_classes: int | None = None
```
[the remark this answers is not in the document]
declanmillar: - I originally had it as a read-only protected member, but then just decided to expose it directly. I'm not sure why; I think because I was working on the factories in Nature which are being refactored to directly expose their attributes. I'll change it back to read-only.
- This is to address #322. The instance-level one is just to compare against prior batches, and for the user to have read-only access for debugging, etc (when I've protected it). It should stay the same for all batches if `warm_start=True`. If `warm_start=False` then it's just overridden by the next fit (as it should represent a whole new trained model anyway in that case, I think).
- Yes, I think so due to the logic in the bullet point above, which is encoded in `_validate_num_classes`.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -132,13 +136,18 @@ def __init__(
...
+
+        self._label_transformer = LabelBinarizer()
```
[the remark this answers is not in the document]
declanmillar: That's a very good point. I'll review that.

qiskit_machine_learning/algorithms/classifiers/vqc.py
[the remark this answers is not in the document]
declanmillar: If you're happy with the need for it, I could add some comments to make it clearer?

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -159,25 +168,118 @@ def num_qubits(self) -> int:
...
+        Returns:
+            The number of inferred classes.
+        """
+        if self._one_hot:
+            num_classes = y.shape[-1]
+        else:
+            num_classes = sum(np.unique(y))
+
+        self.num_classes = self._validate_num_classes(num_classes)
+        return self.num_classes
```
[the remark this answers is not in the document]
declanmillar: That was a last-minute refactor and I bungled it. I'll fix this up.

adekusar-drl:
One small addition here. Please also consider an additional section in the `NeuralNetworkClassifier/Regressor` tutorial where you could show new features of `VQC`.

test/algorithms/classifiers/test_vqc.py
declanmillar: That's a nice suggestion. Thanks! Will do.

qiskit_machine_learning/algorithms/classifiers/vqc.py
declanmillar: > One small addition here. Please also consider an additional section in the `NeuralNetworkClassifier/Regressor` tutorial where you could show new features of `VQC`.

Yep, I'm very happy to do that :).

declanmillar: Yes, I think so. I'll look into doing that.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -29,27 +33,27 @@
...
 
     def __init__(
         self,
         num_qubits: int = None,
         feature_map: QuantumCircuit = None,
         ansatz: QuantumCircuit = None,
-        loss: Union[str, Loss] = "cross_entropy",
-        optimizer: Optional[Optimizer] = None,
+        loss: str | Loss = "cross_entropy",
+        optimizer: Optimizer | None = None,
```
woodsp-ibm, line 52: I am wondering about this. Yes, this seems to be the way forward with its recent introduction, but we now have a mix of code using Union and the vertical bar. This will make things inconsistent over the API Ref. Also I see it changes the formatting for some reason
image
versus now its 
image
Plus somehow before it made the first three params Optional - the should be typehinted in the code that way since None is valid as its the default! In the new version that fact has been lost.

I am wondering if it would be "better" to not do this change incrementally and stick to Union for now and perhaps make some overall change later to have consistency through the code/docs rather than some one way some another.

adekusar-drl: Interesting. @manoelmarques could you please also take a look at these weird artifacts in the generated documentation?
I like the new syntax, but if it confuses sphinx, then it is better to keep the old one.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -154,30 +163,123 @@ def circuit(self) -> QuantumCircuit:
...
+            QiskitMachineLearningError: If the number of classes differs from
+            the previous batch when using a warm start.
+
+        Returns:
+            The validated number of classes.
+        """
+        if self._warm_start and self._num_classes is not None and self._num_classes != num_classes:
+            raise QiskitMachineLearningError(
+                f"Number of classes: {num_classes},\
+                differs from the previous batch: {self.num_classes}."
```
woodsp-ibm, line 265: Normally we would have the multiple strings so they end up being joined as one 
```
                f"Number of classes: {num_classes}, "
                f"differs from the previous batch: {self.num_classes}."
```
With using a line continuation does that not end up including all that leading space of the subseqent line into the final string.

declanmillar: 😅 sure does. Fixed.

qiskit_machine_learning/algorithms/classifiers/vqc.py
woodsp-ibm: I also noted that the current way "fixed" things in regard of some parameters like this should have been declared optional.
> num_qubits: int = None,

where you can see it came out as Optional[int] (I did not know it would do that so was kinda surprised) but it does not do it as int | None in the new way/docs above as you can see - the None aspect is lost. For either way the code typehinting should ideally be fixed of course!

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -29,27 +31,27 @@
...
+    classifier.
 
-    Only supports one-hot encoded labels;
-    e.g., data like ``[1, 0, 0]``, ``[0, 1, 0]``, ``[0, 0, 1]``.
+    Constructs a quantum circuit and corresponding neural network, then uses it
+    to instantiate a neural network classifier.
 
-    Multi-label classification is not supported;
-    e.g., data like ``[1, 1, 0]``, ``[0, 1, 1]``, ``[1, 0, 1]``, ``[1, 1, 1]``.
+    Requires one-hot encoded labels. Multi-label classification is not supported.
```
adekusar-drl, line 41: Would it be better to restore the documentation with examples?

qiskit_machine_learning/algorithms/classifiers/vqc.py
adekusar-drl: @woodsp-ibm So, how do we proceed with the new "|" design in type hints? Keep the old style for definitions and then change them all in one go or update type hints sequentially? I'm favor of updating them in one go.

test/algorithms/classifiers/test_vqc.py
```diff
@@ -12,390 +12,251 @@
...
-
-        # fit to data
-        classifier.fit(X, y)
-
-        # score
-        score = classifier.score(X, y)
+        num_classes, x, y = dataset
+        classifier.fit(x, y)
+        score = classifier.score(x, y)
         self.assertGreater(score, 1 / num_classes)
```
adekusar-drl, line 148: Honestly, I don't recall why it was `1 / num_classes`, but I think this can safely replaced with `0.5`. Otherwise, we may have a not so good classifier.

test/algorithms/classifiers/test_vqc.py
```diff
@@ -12,390 +12,251 @@
...
-        # fix the initial point
-        initial_point = np.array([0.5] * ansatz.num_parameters)
+    @unpack
+    def test_VQC(self, quantum_instance, num_qubits, feature_map, ansatz, optimizer, dataset):
+        """
+        Test VQC with binary and multiclass data using a range of quantum
+        instances, numbers of qubits, feature maps, and optimizers.
+        """
+        if num_qubits is None and feature_map is None and ansatz is None:
+            return
```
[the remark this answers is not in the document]
manoelmarques: skipTest with a comment explaining that the arguments are None is better because then you know the full test didn't run, otherwise it will just show success in both cases.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -87,21 +89,29 @@ def __init__(
...
         self._target_encoder = OneHotEncoder(sparse=False) if one_hot else LabelEncoder()
 
-    def fit(self, X: np.ndarray, y: np.ndarray):  # pylint: disable=invalid-name
+        # For ensuring the number of classes matches those of the previous
+        # batch when training from a warm start.
+        self._num_classes: int | None = None
+
+    @property
+    def num_classes(self) -> int:
+        """The number of classes found in the most recent fit."""
```
woodsp-ibm, line 98: If I call this before doing 'fit' then it will return None as set in the constructor. We could state that and indicate that in the return type;  or alternatively raise an exception stating that instance has not been fitted. I think something needs to be done here in that regard

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -174,4 +184,58 @@ def _validate_input(self, X: np.ndarray, y: np.ndarray = None) -> Tuple[np.ndarr
...
+
+    def _validate_one_hot_targets(self, targets: np.ndarray) -> None:
+        """Validate one-hot encoded targets.
+
+        Ensure one-hot encoded data is valid and not multi-label.
+
+        Raises:
+            QiskitMachineLearningError: If targets are invalid.
+        """
+        if not np.isin(targets, [0, 1]).all():
```
woodsp-ibm, line 208: In this, and other error messages here,  the formatted value in the string is source from an numpy array. I am wondering if in general this is going to make these error messages rather large - depends on the array size and how numpy is set to format them I guess. If they are large by default numpy truncates the data which may end up not revealing the problem anyway. Here for instance can we not just say 'Invalid on-hot targets, data must contain only 0's and 1's' We could try and find some aspect of the data that fails and include that, I am just thinking the whole of the array is maybe too much.

qiskit_machine_learning/algorithms/classifiers/vqc.py
woodsp-ibm: > So, how do we proceed with the new "|" design

Terra has started using this on the primitives - though looking at their API ref docs of them the typehinting does not seem to come out. The docs here seem ok. I wonder if updating everything in one go, though ideal, is practical enough. What we could do is add a prelude section to the release note here about this, i.e say the intent is to migrate to this new style, where previously Optional was used, and some documentation has already been updated in this regard. This would allow things to be a bit more incremental perhaps.

test/algorithms/classifiers/test_vqc.py
declanmillar: Makes sense. I've replaced it with `0.5`. I did have to increase the number of training samples for the score to reliably exceed this on all devices, so the tests might take a bit longer than previously.

qiskit_machine_learning/algorithms/classifiers/vqc.py
declanmillar: Yeah, I notice that now (at least locally) the docs are styled as before, even when coded in the new style (see attached images). Also, given that the changes are already happening incrementally in Optimization and elsewhere, I just left the new-style hints in place.
vqc-parameters-docs
vqc-parameters-code

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
declanmillar: Good point. I've now documented this in the return type and docstring.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
declanmillar: Yeah, good point. Those error messages weren't too helpful. I've edited them to, hopefully, make them clearer. Let me know if you think they still need some work.

qiskit_machine_learning/algorithms/classifiers/vqc.py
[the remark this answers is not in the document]
declanmillar: These changes will now be part of the now-separate PR to fix #320. (This PR's name is now a misnomer.)

qiskit_machine_learning/algorithms/classifiers/vqc.py
declanmillar: These changes will now be part of the now-separate PR to fix #320. (This PR's name is now a misnomer.)

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -29,27 +31,28 @@
...
-    measured expectation value is interpreted as the output of a classifier.
+    The variational quantum classifier (VQC) is a variational algorithm where the measured
+    expectation value is interpreted as the output of a classifier.
+
+    Constructs a quantum circuit and corresponding neural network, then uses it to instantiate a
+    neural network classifier.
 
-    Only supports one-hot encoded labels;
-    e.g., data like ``[1, 0, 0]``, ``[0, 1, 0]``, ``[0, 0, 1]``.
+    Requires one-hot-encoded labels. E.g., ``[[1,0,0],[0,1,0],[0,0,1]]``.
```
adekusar-drl, line 40:
```suggestion
    Requires one-hot-encoded labels. E.g., ``[[1, 0, 0], [0, 1, 0], [0, 0, 1]]``.
```

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -29,27 +31,28 @@
...
 
-    Multi-label classification is not supported;
-    e.g., data like ``[1, 1, 0]``, ``[0, 1, 1]``, ``[1, 0, 1]``, ``[1, 1, 1]``.
+    Multi-label classification is not supported. E.g., ``[[1,1,0],[0,1,1],[1,0,1]]``.
```
adekusar-drl, line 42:
```suggestion
    Multi-label classification is not supported. E.g., ``[[1, 1, 0], [0, 1, 1], [1, 0, 1]]``.
```

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
```diff
@@ -174,4 +185,65 @@ def _validate_input(self, X: np.ndarray, y: np.ndarray = None) -> Tuple[np.ndarr
             elif scipy.sparse.issparse(y):
                 y = cast(spmatrix, y).toarray()  # cast is required by mypy
 
-        return X, y
+        return np.asarray(X), np.asarray(y)
```
adekusar-drl, line 188: May be I missed this in the previous review, but why `np.asarray()` is requried? Is this caused by pylint/mypy/other problems?

test/algorithms/classifiers/test_vqc.py
```diff
@@ -12,390 +12,253 @@
...
-        ("cobyla", "statevector"),
-        ("cobyla", "qasm"),
-        ("bfgs", "statevector"),
-        ("bfgs", "qasm"),
-        (None, "statevector"),
-        (None, "qasm"),
+    # Set-up the datasets.
+    binary_x, binary_y = _generate_classification_data(6, 2, num_features)
+    binary_datasets = [(binary_x, _one_hot_encode(binary_y))]
+    multiclass_x, multiclass_y = _generate_classification_data(9, 3, num_features)
```
adekusar-drl, line 111: Curious, why you choose these sizes, 6 and 9. Is there a reason for that? I don't say this wrong, I think, in the initial tests sizes were different.

declanmillar: Ok, so partly this is due to OCD. I found that I had to increase the number of samples to 9 in the 3 class classification so it can score above `0.5`, rather than `1 / num_classes`. I'm not totally sure it's guaranteed to pass as I think there is still stochasticity in the tests, despite fixing the seed. So, then I made the 2 class classification 6 so that it was linearly proportional. I can probably reduce it.

qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
declanmillar: So, I think this was required by mypy, but that was when the fix for #320 was part of this PR. It might not be strictly necessary now. I will check.

test/algorithms/classifiers/test_vqc.py
adekusar-drl: Usually more samples increases stability. But too many samples increases training time, so there's a trade off in the tests. If it works, fine with me.

declanmillar: Is there any way we remove all the randomness from the optimization? Are there additional seeds we can set?

adekusar-drl: There are two seeds in `QuantumInstance`, randomness in the data generation and `initial_point`. This is what comes to my mind. Seems like everything is fixed.

adekusar-drl: The failing test does not have randomization in the code. So everything is fixed upfront. Fixing a seed in `setUp()` may or may not help.

releasenotes/notes/vqc-input-validation-d5550e6bcd531f22.yaml
```diff
@@ -0,0 +1,10 @@
+---
+features:
+  - |
+    :class:`~qiskit_machine_learning.algorithms.classifiers.vqc.VQC` will now
```
woodsp-ibm, line 4: Can you check in the built documentation that this resolves as a link (and the one below), i.e. you can click it and it takes you to the class. Since its building still in the CI I cannot at present. If you have not done this go to any build Details so you are in the CI results and goto Summary. Then scroll down and you should see a list of build artifacts (zip files) One is the `documentation` so if you download and unzip/untar that you will end up with the built docs. You can start out in the index.html if you like or go directly to the release_notes.html and find this entry there. I am thinking it will need to be
``:class:`~qiskit_machine_learning.algorithms.VQC` `` as this is where we document this class as its imported at this level for users convenience.

test/algorithms/classifiers/test_vqc.py
woodsp-ibm: There is a seed there set up front as a class level variable is what I was referring to above (that was a copy of line in question) in moving it. Where it currently is anything later that uses the algorithms random generator will be subject to loading/execution order e.g. _generate_classification_data that uses it.  I saw comments around random behvaior, and maybe the discussion was over a specific test for which this comment does not specifically apply. I think it does in general though.

adekusar-drl: There's another issue with the tests. Old tests (main branch) take roughly 1:30, new tests 5:10 on my laptop. There are more new tests here, but I suspect the datasets are not great and fitting them, especially with LBFGS, takes quite some time.

declanmillar: I've hardcoded the data with some (hopefully consistently) working examples with two data points for each class. If we want to make it quicker, we could remove some of the other test variations, if they're not providing much extra value.

declanmillar: Ok, I've decided to change 0.5 back to 1 / num_classes, as it was before. Given the unit test speed requirements, I think it's ok just to check that the score is somewhat reasonable given the minimal sample size. It's probably best to evaluate performance properly outside the unit tests. Let me know if you disagree though.

adekusar-drl: I suggest to generate data somehow differently. The original dataset seems to be a bit noisy.
```python
from sklearn.datasets import make_classification
from sklearn.preprocessing import MinMaxScaler

BINARY_X, BINARY_Y = make_classification(n_samples=6, n_features=2, n_classes=2, n_redundant=0, n_clusters_per_class=1, class_sep=5.0, random_state=algorithm_globals.random_seed)
BINARY_X = MinMaxScaler().fit_transform(BINARY_X)

MULTICLASS_X, MULTICLASS_Y = make_classification(n_samples=9, n_features=2, n_classes=3, n_redundant=0, n_clusters_per_class=1, class_sep=5.0, random_state=algorithm_globals.random_seed)
MULTICLASS_X = MinMaxScaler().fit_transform(MULTICLASS_X)
```

You can increase `class_sep` to make classes more separable.

I replaced the original data generation lines with this snippet and in a few cases I get score of `0.88`, mostly it is `1.0`. Reducing the number of samples to 4 and 5 respectively adds more instability in the test results. In the warm start test I got a score of 0.5. So, I think 6 and 9 are reasonable. With these datasets you may test a score against `0.5`, not `1 / num_classes`.

Also, I suggest to remove `None` from a list of optimizers and have a non-parameterized dedicated unit test where we don't set optimizer. The reason is that if optimizer is not set then a default one is created with no `maxiter` set. And this optimizer may cause some tests to run for a long time. Even without `None` the run time is not close to the original time, but somewhat lesser, around 3:20 on my laptop. Reducing `maxiter` in `LBFGS` and `COBYLA` to 3 and 15 respectively is still reasonable. I observe quite a number of scores of `0.77` and still a lot of `1.0` and the run time is 2:10. I don't pretend that my numbers are statistically correct. In fact, there are more tests now than in the original code, so some performance degradation seems inevitable.

You may replace `_one_hot_encode` with `OneHotEncoder(sparse=False)` just for consistency, since I have `MixMaxScaler` in my snippet, but this is a minor.

Let me know if you have questions.

declanmillar: Thanks, Anton. That helps a lot :).

test/algorithms/classifiers/test_vqc.py
```diff
@@ -12,21 +12,100 @@
...
+    seed_simulator=algorithm_globals.random_seed,
+    seed_transpiler=algorithm_globals.random_seed,
+)
+QASM_QUANTUM_INSTANCE = QuantumInstance(
+    Aer.get_backend("aer_simulator"),
+    shots=100,
+    seed_simulator=algorithm_globals.random_seed,
+    seed_transpiler=algorithm_globals.random_seed,
+)
+QUANTUM_INSTANCES_ = [SV_QUANTUM_INSTANCE, QASM_QUANTUM_INSTANCE]
```
adekusar-drl, line 52: A silly question, why some variables end with "_" and some not? 
It looks weird here:
```python
        itertools.product(
            QUANTUM_INSTANCES_, NUM_QUBITS_LIST, FEATURE_MAPS, ANSATZES, OPTIMIZERS, DATASETS
        )
```

declanmillar: This was just because I defined `COBYLA_` to avoid it clashing with the class import `COBYLA`. Then I renamed them all to match, then I thought that looked stupid and reverted the names. But, I missed that one.

I'm probably going to remove these globals to address Steve's comment below anyway.

test/algorithms/classifiers/test_vqc.py
```diff
@@ -35,367 +114,172 @@ class TestVQC(QiskitMachineLearningTestCase):
 
     def setUp(self):
         super().setUp()
-
+        algorithm_globals.random_seed = RANDOM_SEED
```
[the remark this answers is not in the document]
declanmillar: Ah, yes, I see. So, the reason I defined them outside setup is so I could use them directly with `ddt`. I'd actually prefer to refactor the tests to avoid `ddt` and just use native `unittest` methods. But, `ddt` is used elsewhere in the package with globals, so I wanted to keep in line with those (though they may be mutable in these cases). Do you know off the top of your head if I can use variables from setup directly with ddt (without using string flags). Or else are there any objections to refactoring away from ddt?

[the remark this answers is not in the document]
woodsp-ibm: You could define functions, say on the test class, and have these be called by ddt to create the instances on demand. You could also do it in setup like before but say place the instances into dictionary and use places a key in ddt to lookup the instance you want - these could be string keys. This would then be a simple key lookup rather than the if tests, based on the strings, that were there before.

We use ddt where we have the same test in effect but different values for it - the values can be inputs as well as expected results given the different inputs if the expected result varies. Given a want to exercise some code the same way over a set of values is why we use it to decorate the tests - it then under the covers essentially expands out everything to create all the needed tests each one run separately. Yes you can use for loops along with sub tests to approximate this so the test loops over all conditions and each being a sub-test will still run even if one of them fails. We used to use a different package to do this in the past but went over to ddt which is what Terra decided to use.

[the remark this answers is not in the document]
declanmillar: So, I opted to use `subTest()`s instead of `ddt` functions. Mostly because I think it's slightly cleaner and easier to read in this case. I've also moved all of the setups into `setUp`. Hopefully, this means the subtests are set up correctly now.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -159,25 +168,118 @@ def num_qubits(self) -> int:
...
-            y: The target values. Required to be one-hot encoded.
+            y: The target values.
 
         Returns:
-            self: returns a trained classifier.
+            Trained classifier.
         """
-        num_classes = y.shape[-1]
+        y = self._validate_y(y)
+        num_classes = self._get_num_classes(y)
```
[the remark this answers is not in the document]
declanmillar: There have been quite a few changes since you made your comment, but, just to address it now. As far as I can tell the result is preserved in `self._fit_result`, which is only nullified if `warm_start` is `False`. Let me know if I'm missing something though.

adekusar-drl (commented):
I guess `320_nnc_categorical_encoding.py` and `320_vqc_categorical_encoding.py` are added by mistake?

test/algorithms/classifiers/test_vqc.py
```diff
@@ -51,351 +67,222 @@ def setUp(self):
...
-        initial_point = np.array([0.5] * ansatz.num_parameters)
-
-        # construct classifier - note: CrossEntropy requires eval_probabilities=True!
+    def test_VQC(self):
+        """
+        Test VQC with binary and multiclass data using a range of quantum
+        instances, numbers of qubits, feature maps, and optimizers.
+        """
+
+        for (
```
woodsp-ibm, line 128: Since setup is called ahead of the test, things like the quantum_instances etc will be re-used in this for loop. Sure the whole test, that has the for loop, will be reproducible, but subsequent loops in the for will be re-using any objects. E.g if you had the same parameterization in two different tests you might not get the exact same result because of the random sampling in the qasm simulator would not have been reseeded. Maybe overall here that's not too much of a concern here. With the ddt decorator the test will have setup run for each parameterization; i.e. for each loop of this test so they would be independent of one another.

declanmillar: Ah, right. I guess I could call setUp() and tearDown() manually at the start and end of the subTest, respectively. Though that doesn't seem all that elegant either. Perhaps it would be best to go back to DDT then...

declanmillar: As it was a quick fix, I added the setUp and tearDown calls where they're required to ensure consistency. @adekusar-drl, do you have a preference for using DDT or subTests as I have?

adekusar-drl: Personally, I don't mind having both ddt and subTests, but considering that we have a lot of ddt tests, I think, it is better to have ddt for consistency.

declanmillar: Sure, no problem. There are some advantages to using DDT. For example, one can run a specific data-generated test individually without having to loop over all of them, which is nice. For some reason, I didn't like calling a function with String arguments to instantiate the immutable classes, but it seems this is the best way to do things. I'll refactor it again ASAP.

declanmillar: Ok, I've refactored it in the first way that you suggested above, @woodsp-ibm.

> You could define functions, say on the test class, and have these be called by ddt to create the instances on demand. You could also do it in setup like before but say place the instances into a dictionary and put the keys in ddt to lookup the instance you want in the test - these could be string keys. This would then be a simple key lookup rather than the if tests, based on the strings, that were there before.

Hopefully, I implemented it correctly this time :sweat_smile:.

test/algorithms/classifiers/test_vqc.py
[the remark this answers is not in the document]
declanmillar: Please see [this](https://github.com/Qiskit/qiskit-machine-learning/pull/374#discussion_r875142649) comment for further discussion.

test/algorithms/classifiers/test_vqc.py
woodsp-ibm: Great, thanks for hanging in there with this.

woodsp-ibm (approved):

adekusar-drl:
Thanks a lot!

PR #377, opened 2022-04-13
Add a section on saving/loading hybrid models to the tutorials

adekusar-drl:
### Summary
Added two sections to the tutorials:
- in the `TorchConnector` tutorial added a few lines showing how to save/load a hybrid model.
- in the Save/Load tutorial added a reference to the torch connector tutorial and added a pseudo code as an example.


docs/tutorials/09_saving_and_loading models.ipynb
```diff
@@ -744,6 +745,32 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "acknowledged-freight",
+   "metadata": {},
+   "source": [
+    "## 4. PyTorch hybrid models\n",
+    "\n",
+    "In general, if you follow PyTorch recommendations on saving and loading models you should not have issues with hybrid models. For more details please refer to the PyTorch Connector tutorial [here](https://qiskit.org/documentation/machine-learning/tutorials/05_torch_connector.html) where a short snippet shows how to do it.\n",
```
woodsp-ibm, line 755: > In general, if you follow PyTorch recommendations on saving and loading models you should not have issues with hybrid models.

Is this more `To save and load hybrid models, when using the TorchConnector, follow the PyTorch recommendations of saving and loading the models`. For me the `In general... you should not have issues` kinda suggests there are cases where you will have some sort of issue. Is that the case? (The text in the TorchConnector tutorial is similar in this regard)

adekusar-drl: I updated as you suggested. I wrote that cause users may try to run many crazy things and while the connector works on low level pytorch interfaces they may hit cases we never tried or will never support. But I guess, we can cope such cases later if any.

woodsp-ibm (approved):

PR #391, opened 2022-04-29
Rename user parameter

chetmurthy:
### Summary

Rename QuantumKernel parameter "user_parameters" to "training_parameters".
rename property methods, instvars analogously.
update other depending files
update unit-tests
update IPYNB tutorial
run lint, style, tox -elint

### Details and comments

Renamed parameter, using "deprecated_argument" decorator so the old parameter will continue
working for the moment.  Renamed property methods, instance-variables analogously.


adekusar-drl:
@chetmurthy Seem like you have commits from two your accounts and the second one is "Chet Murthy" which is "seems not to be a GitHub user", so this account can't sign CLA. Thus, the PR won't pass the CLA check. I'd suggest to just **copy** new sources to a new clean branch and commit the changes under the public github account and re-open a new PR. 
@calebj15 I guess you had a similar issue?

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -47,13 +52,15 @@ class QuantumKernel:
     algorithms such as support vector classification, spectral clustering or ridge regression.
     """
 
+    @deprecate_arguments("0.5.1", {"user_parameters": "training_parameters"})
```
woodsp-ibm, line 55: This is the version at which it was deprecated. This would be 0.5.0 as per the [VERSION.txt](https://github.com/Qiskit/qiskit-machine-learning/blob/main/qiskit_machine_learning/VERSION.txt) (0.5.1 would be a bug fix release and deprecation is not something that would be done on a bug fix release). Same for the other deprecated aspects in this PR.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -47,13 +52,15 @@ class QuantumKernel:
...
     def __init__(
         self,
         feature_map: Optional[QuantumCircuit] = None,
         enforce_psd: bool = True,
         batch_size: int = 900,
         quantum_instance: Optional[Union[QuantumInstance, Backend]] = None,
-        user_parameters: Optional[Union[ParameterVector, Sequence[Parameter]]] = None,
+        *,
+        training_parameters: Optional[Union[ParameterVector, Sequence[Parameter]]] = None,
```
woodsp-ibm, line 63: Is there a reason to force training_parameters to be a keyword parameter - it could still be used positionally where user_parameters was before right, if that's how someone called this, since its only a name change.

woodsp-ibm (commented):
Is there any context for this change? The PR summary does not include anything in that regard and I don;t see an issue. I know there was discussion over the name when this was created.

Changes here, that affect end user, need a [release note to be created](https://github.com/Qiskit/qiskit-machine-learning/blob/main/CONTRIBUTING.md#release-notes) and included in the PR. You can find examples in the `releasenotes` folder here is that's helpful.

qiskit_machine_learning/kernels/quantum_kernel.py
chetmurthy: Hi, this is a "starter PR" (joined the S3 team); @calebj15 suggested it, there's an issue ( https://github.com/qiskit-community/prototype-quantum-kernel-training/issues/33 ) and I'm working thru it.  I'll correct these two problems posthaste.

qiskit_machine_learning/kernels/quantum_kernel.py
chetmurthy: two thoughts:
1. YES, you're right, and the code as-written almost-certainly has a bug: if user_parameters is supplied *without* a keyword-argument, the code simply won't work!  Oy, I overlooked this, and will fix.
2. My reasoning was: 
3. I found no examples of user_parameters being used as a positional argument.  And thought: "would be better to make training_parameters a keyword".  But sure, can retain it as positional.
4. *But*, once user_parameters is deprecated, training_parameters can return to being a possibly-positional argument.  Right now, it would be bad to allow users to write code that supplied both user_parameters and training_parameters as positional arguments -- because that would allow users to write code that would not work, once user_parameters was fully-removed.

qiskit_machine_learning/kernels/quantum_kernel.py
chetmurthy: Oops, also, forgot about the release-note, will fix. Also, I don't know what release user_parameters will get removed at?  Do you have a suggestion?

chetmurthy: I'm writing a release-note, and it isn't clear to me what the file should be named.  That is, sure, it should be named "rename-user_parameters-to-training_parameters-7d8cf1e056b967a7.yaml", but should I put it in the releasenotes/notes directory, or in a new "0.5" subdirectory ?

chetmurthy: Also, "reno report" stops at version "0.3", does not proceed to format releasenotes for version "0.4".  Not sure why.

woodsp-ibm: > Also, I don't know what release user_parameters will get removed at? Do you have a suggestion?

The version in the deprecation message is the version at which point the deprecation was made (commenced) so in the messages it needs to be 0.5.0 which will be the next version released and the point when this code/PR is included. We do not state the point at which it will be removed, beyond saying at least 3 months or 2 releases - see https://qiskit.org/documentation/contributing_to_qiskit.html#deprecation-policy for more info - where bullet point 3a talks about the 2 release aspect.

> I'm writing a release-note, and it isn't clear to me what the file should be named. 

You just need to do `reno new some_name_for_your_fix` from the root of the project - see here for more info https://github.com/Qiskit/qiskit-machine-learning/blob/main/CONTRIBUTING.md#release-notes And that will create a yaml filem, with the commit reference in the name too, in the `releasenotes` folder. You should edit it and commit it there. The notes are only moved later into version sub-folder just prior to release in order to keep them organized.

If you want to see the release notes in formatted form then the easiest way is to do `make html` in the project root to build the documentation for ML. This will create a `_build` folder under `docs`, and in there will be an `html` folder where you can see release_notes.html. You can click on it directly there is an index.html that is the root of the docs, and has a link to release notes in that.

qiskit_machine_learning/kernels/quantum_kernel.py
woodsp-ibm: The idea of deprecation is that most importantly any working code they have written still works when they install the new version that introduced the deprecation and to give time for a user to migrate their code to the new way.

I agrees its unlikely someone has done all positional arguments but if they have their code should ideally continue to work. So far we have not forced users to do keyword arguments so I think I would leave it to still allow positional. If they have done positional they they will not notice the change in name. For sure someone should not be supplying both the old and newly named parameter - that's really a user error.

qiskit_machine_learning/kernels/quantum_kernel.py
woodsp-ibm: > Hi, this is a "starter PR" (joined the S3 team)

Hi, I heard you had just started. I imagined you had discussed it with someone but I was lacking any context and its helpful to have that. For instance in this case, as you did in the comment, the issue in the other repo could have been included as context.

chetmurthy: OK, I've fixed the various problems; should I delete this PR and resubmit with a single commit (to minimize clutter) ?  Or leave as-is?

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -52,13 +52,14 @@ class QuantumKernel:
...
-    @deprecate_arguments("0.5.1", {"user_parameters": "training_parameters"})
+    @deprecate_arguments("0.5.0", {"user_parameters": "training_parameters"})
     def __init__(
         self,
         feature_map: Optional[QuantumCircuit] = None,
         enforce_psd: bool = True,
         batch_size: int = 900,
         quantum_instance: Optional[Union[QuantumInstance, Backend]] = None,
+        user_parameters: Optional[Union[ParameterVector, Sequence[Parameter]]] = None,
         *,
```
woodsp-ibm, line 63: I thought the deprecate arguments decorator renamed the parameter to the new one. Hence I don't think its needed in the constructor. Any new code should be dealing with training_parameters. @manoelmarques please correct me if am wrong since you are more familiar with the decorators than I. Hence aside from the addition of the decorator, the only only change I would expect to see over the old init is just the renaming of user_parameters -> training_parameters since their typehint/function is unaltered otherwise. I think you can also drop the text about user_parameters from the docstring since this really only applies to supporting old code and if you have no old code, where you will see the deprecation messages emitted when running it (assuming you called it via keyword arg).

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -74,6 +75,9 @@ def __init__(
                  quantum gates on the feature map circuit which may be tuned. If users intend to
                  tune feature map parameters to find optimal values, this field should be set.
             user_parameters: deprecated keyword alias for training_parameters
+
+        Raises:
+            QiskitMachineLearningError: if invoker supplies both user_parameters and training_parameters.
```
woodsp-ibm, line 80: See above comment - I don't think. You should never see user_parameters in the algo as it would have been renamed by decorator.

releasenotes/notes/rename-user_parameters-to-training_parameters-7d8cf1e056b967a7.yaml
```diff
@@ -0,0 +1,16 @@
...
+    ``bind_user_parameters``, ``get_unbound_user_parameters``, to
+    ``training_parameters``, ``trainin_param_binds``, ``assign_training_parameters``
+    ``bind_training_parameters``, ``get_unbound_training_parameters``.
+
+deprecations:
+  - |
+    The existing keyword-argument ``user_parameters`` and the properties
+    ``user_parameters``, ``user_param_binds``, the methods ``assign_user_parameters``
+    ``bind_user_parameters``, ``get_unbound_user_parameters`` are all deprecated
+    and will be removed after three months.
```
woodsp-ibm, line 16: We should just end this after `are all deprecated` I think - the policy is no sooner than three months and it needs to be in there at least one more version so in practice it cannot be removed any sooner than 0.7.0 or 3 months, whichever comes first,

qiskit_machine_learning/kernels/quantum_kernel.py
chetmurthy: Aha, right.  No need for user_parameters at all.  Been a long time since I coded Python, all this decorator stuff is new to me.  Nuked user_parameters in the constructor.

qiskit_machine_learning/kernels/quantum_kernel.py
chetmurthy: roger that.

releasenotes/notes/rename-user_parameters-to-training_parameters-7d8cf1e056b967a7.yaml
chetmurthy: done.

chetmurthy:
Re: the CLA, I don't know what's going wrong.  "Chet Murthy" is, I think, the name that I'm getting from my IBM email ID (which is associated to my public github account .... b/c I tried to join the IBM "organization" (but that didn't seem to go thru ... have to try again).  But I only have one public Github account, "chetmurthy", and I signed the CLA with that.  I also tried to sign again just now, and it again signed with my "chetmurthy" account.

Not sure what's wrong.

woodsp-ibm:
> Re: the CLA, I don't know what's going wrong. 

It comes down to the gitconfig and what ends up in the commits here. The name/email should match the github account for which you have signed the CLA. What happens sometimes is people work on more than one machine and the gitconfig is not set the same way, resulting in what github here picks up as different people

adekusar-drl:
> Re: the CLA, I don't know what's going wrong. "Chet Murthy" is, I think, the name that I'm getting from my IBM email ID (which is associated to my public github account .... b/c I tried to join the IBM "organization" (but that didn't seem to go thru ... have to try again). But I only have one public Github account, "chetmurthy", and I signed the CLA with that. I also tried to sign again just now, and it again signed with my "chetmurthy" account.
> 
> Not sure what's wrong.

There are two users as you can on the screenshot.
"Chet Murthy" is the user that is not on public github, this is an internal one. It is not anyhow associated with [chetmurthy](https://github.com/chetmurthy) which is public one. If you started working in an internal repo and then pushed the same commits to the public repo, then you have user under the one PR.
image

I don't think you can sign the CLA using the internal user. So, as I said the easiest way is to make a plain copy of the changed files to a new branch and then create a new clean PR.

chetmurthy:
OK:
1. make a clean copy of the PR and resubmit
2. with my external github ID, *not* with my IBM ID.

woodsp-ibm:
Make sure the gitconfig on the machine you are committing from is set with the user name and the email of the public github account you are using. I see two different emails in the commits from your PR, one is the company one and the other gmail. It seems these emails and one of them is the `chetmurphy` account, the other it has you name in the gitconfig but the email does not match.

PR #395, opened 2022-05-02
Fix declaration of Variable in GradientPenalty

ebermot:
### Summary
In the `gradient_penalty` function in `pytorch_discriminator` of the qGAN structure, the `z` variable was not properly initialised. 

### Details and comments

In this function, the `z` variable was being initialised using `Variable` module from `torch.autograd.Variable`. However, in this function, only `torch` was imported and not `torch.autograd.Variable`. In this pull request, I went from `z=Variable(x+delta_, requires_grad=True)` to `z=torch.autograd.Variable(x+delta_, requires_grad=True)`.


woodsp-ibm:
Thanks for noticing/fixing this. I am wondering did you encounter it during running or just saw it? Reason I am asking is that the path that would include that gradient penalty does not seem to be easy to get at. Running the qgan it calls train on the discriminator but leaves the penalty param as False with no way, that I see, to change that. Only if you call train directly on the discriminator do I believe you can make the code go down that path.

ebermot:
I encountered this issue when I initially wanted to add the gradient penalty to the `numpy_discriminator`. In order to enable it, in my personal repo, I added in the qGAN definition a penalty parameter which enables/disables the gradient penalty technique for the discriminator. However, for non-Wasserstein qGANs, the gradient penalty is not needed but can be added to fasten the convergence.

ElePT:
Hey @ebermot, thanks for pointing out this issue. About the use of `Variable`, it has actually been deprecated in favor of `Tensor` since PyTorch 0.4. You can check the migration guide [here](https://pytorch.org/blog/pytorch-0_4_0-migration-guide/). I think that we could take this chance to update the code and just get rid of `Variable` and other outdated PyTorch syntax. 

About setting the gradient penalty, I agree with @woodsp-ibm, it is currently not possible. We could maybe move the `penalty` flag from the discriminator `train()` method to `__init__()`, or include a `set_penalty()` method, so that it can be set by the user and then passed to the `QGAN` instance doing `QGAN.set_discriminator(discriminator_with_penalty)`.

adekusar-drl:
@ElePT I would not put too much efforts into this fix, I'm totally fine if you use `Variable` as the rest code does.

ElePT:
> @ElePT I would not put too much efforts into this fix, I'm totally fine if you use `Variable` as the rest code does.

Ok, this makes sense, but let's keep it in mind for future PyTorch refactoring @Zoufalc .

woodsp-ibm:
Can we do a reno release note for this bug fix and a unit test for CI that covers this. Potentially the latter can call method(s) directly on the discriminator in order to cover this aspect.

That migration guide is now over 4 years old! In terms of the comment around Variable this is the only file I think where its used - aside from being part of the optionals.py file here in regards of checking Torch install. It all still works and presumably would work with an older Torch still if someone still has this - there is no constraint we have on Torch version - so I think I would leave this as-is for now.

> Variable wrapping continues to work as before but returns an object of type [torch.Tensor](https://pytorch.org/docs/0.4.0/tensors.html). This means that you don’t need the Variable wrapper everywhere in your code anymore.

releasenotes/notes/gradient-penalty-parameter-5cd6ecc3e78ce187.yaml
```diff
@@ -0,0 +1,12 @@
...
+features:
+  - |
+    In the previous releases, in the ``QGAN`` class, the gradient penalty could not be enabled to train the discriminator
+    with a penalty function. Thus, a gradient penalty parameter was added during the initialization of the QGAN algorithm. 
+    This parameter indicates whether or not penalty function is applied to the loss function of the discriminator during training.
+
+fixes:
+  - |
+    Previously in the ``pytorch_discriminator`` class of the ``QGAN`` algorithm, if the gradient penalty parameter was enabled, 
+    the latent variable z was not properly initialised : Variable module was used instead of torch.autograd.Variable.
```
woodsp-ibm, line 12:
```suggestion
    the latent variable z was not properly initialized : Variable module was used instead of torch.autograd.Variable. 
```

adekusar-drl (commented):
Nice work as a first contribution. Thank you.
@woodsp-ibm do you have anything left in the PR?

woodsp-ibm (approved):

PR #396, opened 2022-05-02
rename user_parameters to training_parameters

chetmurthy:
Starter project: issue https://github.com/qiskit-community/prototype-quantum-kernel-training/issues/33

Rename QuantumKernel parameter "user_parameters" to "training_parameters".
rename property methods, instvars analogously.
update other depending files
update unit-tests
update IPYNB tutorial
run lint, style, tox -elint
add release-notes

# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
#
# On branch rename-user_parameters-to-training_parameters
# Your branch is up to date with 'my-fork/rename-user_parameters-to-training_parameters'.
#
# Changes to be committed:
#	modified:   docs/tutorials/08_quantum_kernel_trainer.ipynb
#	modified:   qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
#	modified:   qiskit_machine_learning/kernels/quantum_kernel.py
#	modified:   qiskit_machine_learning/utils/loss_functions/kernel_loss_functions.py
#	new file:   releasenotes/notes/rename-user_parameters-to-training_parameters-7d8cf1e056b967a7.yaml
#	modified:   test/kernels/algorithms/test_qkernel_trainer.py
#	new file:   test/kernels/test_deprecated_qkernel_methods.py
#	modified:   test/kernels/test_qkernel.py
#


chetmurthy:
Not sure what's wrong.
1. I reran the single testcase, passed w/o problems on Ubuntu 20.04
2. reran the entire test suite "make test"; again passed w/o problems

Is there a way for me to rerun the *precise* environment that the CI uses?  Maybe a Docker recipe I could use to do so?

woodsp-ibm:
@chetmurthy On occasion it seems that sometimes CI has a random failure and re-running the tests pass. This is intermittent and personally I have not seen one for a while, apart from the failure here under 3.8 you just had, though others have I believe. We try to sort these but the cause is not always apparent - we have seeds etc setup in order to allow tests to be reproducible.

As to running the tests you can do so on your fork. It used to be the github actions was enabled there by default - I turned it off so it did not run there as I do tests locally though not the exact same environment.

docs/tutorials/08_quantum_kernel_trainer.ipynb
```diff
@@ -286,12 +1293,12 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "{   'optimal_parameters': {ParameterVectorElement(θ[0]): 1.2984045829959046},\n",
-      "    'optimal_point': array([1.29840458]),\n",
-      "    'optimal_value': 11.733763373742224,\n",
+      "{   'optimal_parameters': {ParameterVectorElement(θ[0]): 1.6595658861350273},\n",
+      "    'optimal_point': array([1.65956589]),\n",
+      "    'optimal_value': 12.928464717873807,\n",
```
woodsp-ibm, line 1298: These values seem quite different from before - as well as the accuracy below which as gone from 1.0 to 0.5. If you run this does it change each time? The changes that you did, simple renames, should of course not affect this. However I recall SPSA having some random aspects so maybe the algorithm_globals seed should be set to allow the results here to be reproducible as having the results vary each time can be unhelpful for end-users unless that is truly the intent.

test/kernels/test_deprecated_qkernel_methods.py
```diff
@@ -0,0 +1,709 @@
...
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# NOTE WELL that this file of tests should disappear when we remove the
+# deprecated ``user_param`` argument/property/methods in quantum_kernel.py.
+# These tests exist ONLY to ensure that these deprecated objects correctly.
+
+""" Test QuantumKernel """
```
woodsp-ibm, line 17: Can we subset this down to just testing the user_parameter renaming related logic. Ie that using the deprecated parameter and the deprecated properties/methods work as expected. This, I would think, would be more just the original tests that were changed over to the new naming. 
 
Also when we test function that we expect to raise deprecated warnings we also suppress these. If you look in the CI results we otherwise collect all deprecated messages so as to ensure we are not ourselves using deprecated logic in the code which we need to fix. So we avoid having messages from CI when we know we are explicitly calling such deprecated logic just to test it.

docs/tutorials/08_quantum_kernel_trainer.ipynb
```diff
@@ -240,13 +1238,22 @@
...
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/tmp/ipykernel_100988/4072072807.py:5: DeprecationWarning: __init__: the training_parameters argument is deprecated as of version 0.4.1 and will be removed no sooner than 3 months after the release. Instead use the training_parameters argument.\n",
+      "  quant_kernel = QuantumKernel(fm, training_parameters=training_params, quantum_instance=backend)\n"
+     ]
```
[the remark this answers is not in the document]
chetmurthy: Steve, the notebook seems corrupted.  I made what should have been innocuous changes manually, and yet somehow a *ton* of changes have appeared, and I am a little mystified as to how that happened, b/c I certainly didn't make these changes.  I'm reverting and re-applying my minimal changes.

test/kernels/test_deprecated_qkernel_methods.py
chetmurthy: 1. oh, aha, yes, I'll suppress the deprecation warnings

2. I don't understand what you're suggesting:
`This, I would think, would be more just the original tests that were changed over to the new naming.`
Do you mean that I should just copy over the original test-file and rename the few places to get it working, so it will test that the originally-named(-now-deprecated) methods still work?  B/c that's what I did.  Or do you mean that you'd like to see just a few tests for these deprecated methods?  The reason I chose to do it this way, is that it is guaranteed to test all the relevant behaviour, and once the deprecated methods are removed, we can just remove the entire test-file.

The `test_deprecated.....py` file is literally "start with test_qkernel.py and rename the few spots that need renaming so it'll run, and leave all the deprecated method-names in-place unmodified".

Does that convince?

docs/tutorials/08_quantum_kernel_trainer.ipynb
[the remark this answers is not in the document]
woodsp-ibm: Some changes we might expect when re-running the notebook as it will re-generate the output cells, and if thats an image it may well end up with a huge diff. Also, depending on how the notebook was created and what version you are running it may well end up adding tags to the cell, or removing them. Generally I just look for difference that seems to be outside of what I might expect.

test/kernels/test_deprecated_qkernel_methods.py
woodsp-ibm: I think for part 2 I was thinking all that was needed was just those specific tests that you needed to update with the new names. Here they would be just as they were with the old names. I think this represents quite a significant sub-set of all the tests in this file. If the test does not explicitly deal with the old deprecated names/properties I think we should be able to rely on the original updated file to cover everything right?

woodsp-ibm: It should just be a copy of those tests from the original file that were renamed right - at least in theory. I did not check in detail but was hoping they would be easy enough to isolate.  Yes the file will go away but ahead of then it will increase the CI time by doubling up on all the other tests - it will be around at least 3 months and more likely 6 or more given we tend to release roughly quarterly or so.

chetmurthy: Fair enough: I'm removing everything that doesn't concern the renaming.

releasenotes/notes/rename-user_parameters-to-training_parameters-7d8cf1e056b967a7.yaml
```diff
@@ -0,0 +1,15 @@
...
+    ``user_parameters``, ``user_param_binds``, the methods ``assign_user_parameters``
+    ``bind_user_parameters``, ``get_unbound_user_parameters``, to
+    ``training_parameters``, ``trainin_param_binds``, ``assign_training_parameters``
+    ``bind_training_parameters``, ``get_unbound_training_parameters``.
+
+deprecations:
+  - |
+    The existing keyword-argument ``user_parameters`` and the properties
+    ``user_parameters``, ``user_param_binds``, the methods ``assign_user_parameters``
+    ``bind_user_parameters``, ``get_unbound_user_parameters`` are all deprecated.
```
woodsp-ibm, line 15: The `issues` section is to cover known bugs/problems that are in the release. I think this note can be just done as a deprecations section. And I am thinking something like this which more or less combines the above as well as provides a bit of explanation.

```
    For the class :class:`~qiskit_machine_learning.kernels.QuantumKernel`, to
    improve usability and better describe the usage, ``user_parameters`` has been
    renamed to ``training_parameters``; current behavior is retained. For this
    change the constructor parameter ``user_parameters`` is now deprecated and
    replaced by ``training_parameters``. The related properties and methods are
    renamed to match. Thus the properties ``user_parameters`` and ``user_param_binds`
    are both deprecated and replaced by ``training_parameters`` and
    ``training_param_binds`` respectively. The methods ``assign_user_parameters``,
    ``bind_user_parameters`` and ``get_unbound_user_parameters`` are deprecated and
    replaced by ``assign_training_parameters``, ``bind_training_parameters`` and
    ``get_unbound_training_parameters``.
```
Its possible to link the methods  but I figured linking to the class would suffice. I also corrected a typo in training i.e. `trainin_param_binds` 

While changing the naming do we want to consider `training_param_binds? We have parameters spelled out in the others and some are longer e.g `get_unbound_user_parameters` would not this be better as say `training_parameters_binding` or something more consistent like that? I note this is a property versus get_unbound_user_parameters as a method. The latter could be made a property as `unbound_user_parameters` as its hard for me to see really why one is a property and the other method.  @calebj15 for any thoughts here

calebj15: > While changing the naming do we want to consider `training_param_binds? We have parameters spelled out in the others and some are longer e.g `get_unbound_user_parameters`would not this be better as say`training_parameters_binding`or something more consistent like that? I note this is a property versus get_unbound_user_parameters as a method. The latter could be made a property as`unbound_user_parameters` as its hard for me to see really why one is a property and the other method. @calebj15 for any thoughts here

unbound_user_parameters isn't really a field that is kept up with, so I didn't see it as a property. It is something that is calculated on the fly and returned, so I felt that a class method was more appropriate there

woodsp-ibm: > so I felt that a class method was more appropriate there

Sure thats fine. For me its more how the end-user engages with the class rather than if there is computation that goes on behind the scenes for the result. If its significant computation then a method though is probably the better choice as accessing a property should feel no so different than accessing a public instance variable for me.

chetmurthy: Re: release-note, I added a bullet-list.  I agree that it's much better than the text; shall I remove most of the textual list in favor of the bullet-list?

chetmurthy: OK, I cleaned-up the release-note (removed most of the text), the bullet-list is much nicer.

woodsp-ibm: Any comment on this aspect from above? Its the only name that has `parameters` abbreviated to `param`

> While changing the naming do we want to consider training_param_binds? We have parameters spelled out in the others and some are longer e.g get_unbound_user_parameters Would not this be better as say training_parameters_binding or something more consistent like that?

woodsp-ibm: The parameter is already being renamed under this PR from `user_param_binds` to `training_param_binds`. All this would do is to change the new name a little more, thats all I think!

calebj15: Yes, I agree. Let's go with training_parameter_binds!

chetmurthy: Roger that, will make the change.

woodsp-ibm (approved):
Thanks for doing this - and the release note is indeed a lot easier to read with the bullet list of changed names.

adekusar-drl:
@woodsp-ibm Thanks for reviewing the PR!

PR #397, opened 2022-05-04
Randomize file names in the model save/load tests

adekusar-drl:
### Summary
Sometimes CI unit tests fail on access to a temp where a test model was saved. Likely this happens due to some concurrent test execution. This PR randomizes file names in the tests.

### Details and comments
Exception that may happen now:
```
test.algorithms.classifiers.test_neural_network_classifier.TestNeuralNetworkClassifier.test_save_load_2_circuit_qnn
-------------------------------------------------------------------------------------------------------------------

Captured traceback:
~~~~~~~~~~~~~~~~~~~

      File "/home/runner/work/qiskit-machine-learning/qiskit-machine-learning/test/algorithms/classifiers/test_neural_network_classifier.py", line 320, in test_save_load
    classifier_load = NeuralNetworkClassifier.load(file_name)

      File "/home/runner/work/qiskit-machine-learning/qiskit-machine-learning/qiskit_machine_learning/algorithms/serializable_model.py", line 54, in load
    with open(file_name, "rb") as handler:

    FileNotFoundError: [Errno 2] No such file or directory: '/tmp/classifier.model'
```

### DIFF
--- a/test/algorithms/classifiers/test_neural_network_classifier.py
+++ b/test/algorithms/classifiers/test_neural_network_classifier.py
@@ -314,7 +314,9 @@ class TestNeuralNetworkClassifier(QiskitMachineLearningTestCase):
         original_predicts = classifier.predict(test_features)
 
         # save/load, change the quantum instance and check if predicted values are the same
-        file_name = os.path.join(tempfile.gettempdir(), "classifier.model")
+        file_name = os.path.join(
+            tempfile.gettempdir(), "classifier" + str(np.random.randint(1000)) + ".model"
+        )
         classifier.save(file_name)
         try:
             classifier_load = NeuralNetworkClassifier.load(file_name)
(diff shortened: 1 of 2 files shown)


adekusar-drl:
Closed by mistake

adekusar-drl:
> This approach will reduce the probability of a race condition, but it will not eliminate it. For the qiskit-machine-learning tests, that is probably good enough. If one wanted to eliminate entirely the race (as would be essential [if writing code that must be secure](https://cwe.mitre.org/data/definitions/377.html), e.g. on a multi-user machine), one should use [`tempfile.TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) with a deterministic filename instead of `gettempdir` plus randomness.

Thanks, probably this is a better solution. Will update as you suggested.

woodsp-ibm (approved):
It looks fine. For a single file though, an alternative, that we use elsewhere is `mkstemp` which just returns a randomly named file, but like before does require it to be explicitly removed.

PR #398, opened 2022-05-04
Remove SVC inheritance for PegasosQSVC, fixes #387

### LINKED ISSUE #387: A problem with PegasosQSVC function when using decision_function function
### Environment

- **Qiskit Machine Learning version**: 0.4.0
- **Python version**: 3.7.1
- **Operating system**: Ubuntu 20.04

### What is happening?

Using the PegasosQSVC example [here](https://qiskit.org/documentation/machine-learning/tutorials/07_pegasos_qsvc.html), and calculating the score using decision_function, I got the following error:

`AttributeError: 'PegasosQSVC' object has no attribute '_sparse'`

 I use the example provided here https://qiskit.org/documentation/machine-learning/tutorials/07_pegasos_qsvc.html

```python
from qiskit_machine_learning.algorithms import PegasosQSVC

pegasos_qsvc = PegasosQSVC(quantum_kernel=qkernel, C=C, num_steps=tau)
#pegasos_qsvc = PegasosQSVC(quantum_kernel=qkernel)

pegasos_qsvc.fit(train_features, train_labels)

pegasos_score = pegasos_qsvc.score(test_features, test_labels)

y_score_qsvc = pegasos_qsvc.decision_function(test_features) # Just adding this line which is the cause of the problem. 
#y_score_qsvc = pegasos_qsvc.predict(test_features)

print(y_score_qsvc)
print(f"PegasosQSVC classification test score: {pegasos_score}")
```

### What should happen?

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_8351/1609182878.py in <module>
      9 pegasos_score = pegasos_qsvc.score(test_features, test_labels)
     10 
---> 11 y_score_qsvc = pegasos_qsvc.decision_function(test_features)
     12 #y_score_qsvc = pegasos_qsvc.predict(test_features)
     13 

~/.anaconda3/envs/qiskit/lib/python3.7/site-packages/sklearn/svm/_base.py in decision_function(self, X)
    754         transformation of ovo decision function.
    755         """
--> 756         dec = self._decision_function(X)
    757         if self.decision_function_shape == "ovr" and len(self.classes_) > 2:
    758             return _ovr_decision_function(dec < 0, -dec, len(self.classes_))

~/.anaconda3/envs/qiskit/lib/python3.7/site-packages/sklearn/svm/_base.py in _decision_function(self, X)
    512         # NOTE: _validate_for_predict contains check for is_fitted
    513         # hence must be placed before any other attributes are used.
--> 514         X = self._validate_for_predict(X)
    515         X = self._compute_kernel(X)
    516 

~/.anaconda3/envs/qiskit/lib/python3.7/site-packages/sklearn/svm/_base.py in _validate_for_predict(self, X)
    599             )
    600 
--> 601         if self._sparse and not sp.isspmatrix(X):
    602             X = sp.csr_matrix(X)
    603         if self._sparse:

AttributeError: 'PegasosQSVC' object has no attribute '_sparse'
```

### Any suggestions?

_No response_

gentinettagian:
### Summary
As discussed in #387, inheriting from `sklearn.svm.SVC` leads to major issues in `PegasosQSVC`. The main methods (`fit` and `predict`) are overridden and calling `decision_function` raises an error. As far as I can see, the only method that we obtain for free is `score`, which can also be inherited from the `ClassifierMixin` class.

This PR removes the `SVC` inheritance and instead inherits from `ClassifierMixin` to keep support for the `score` method. In addition, a `decision_function` method is added following the signature used in `scikit-learn`.

### Details and comments
The logic for the `decision_function` method was already present in the `predict` method. The change simply moves the logic into a new method. `predict` now calls `decision_function` and then makes the predictions by taking the sign of the returned values.


qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -27,7 +27,7 @@
 logger = logging.getLogger(__name__)
 
 
-class PegasosQSVC(SVC, SerializableModelMixin):
+class PegasosQSVC(ClassifierMixin, SerializableModelMixin):
     """
     This class implements Pegasos Quantum Support Vector Classifier algorithm developed in [1]
     and includes overridden methods ``fit`` and ``predict`` from the ``SVC`` super-class. This
```
woodsp-ibm, line 33: The class docstring ought to be updated since it had/has `and includes overridden methods ``fit`` and ``predict`` from the ``SVC`` super-class`

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -102,6 +102,11 @@ def __init__(
         if seed is not None:
             algorithm_globals.random_seed = seed
 
+        if C > 0:
+            self.C = C
+        else:
+            raise ValueError("C has to be a positive number.")
```
woodsp-ibm, line 108:
```suggestion
            raise ValueError(f"C has to be a positive number, found {C}.")
```
Normally we like to include the failing value in case that can be helpful for a user seeing the error. SO maybe something like the suggestion

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
gentinettagian: Thank you for noticing! I fixed it.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -215,6 +221,34 @@ def predict(self, X: np.ndarray) -> np.ndarray:
...
+               ``(m_samples, n_samples)``. Where ``m`` denotes the set to be predicted and ``n`` the
+               size of the training set. In that case, the kernel values in X have to be calculated
+               with respect to the elements of the set to be predicted and the training set.
+
+        Returns:
+            An array of the shape (n_samples), the decision function of the sample.
+
+        Raises:
+            QiskitMachineLearningError:
+                - decision_function is called before the model has been fit.
```
adekusar-drl, line 248: Is it better?
```suggestion
                - the method is called before the model has been fit.
```

releasenotes/notes/independent-pegasosQSVC-495100889ef28098.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
+  - |
+    Calling `PegasosQSVC.decision_function()` raises an error. Fixed by writing
+    own method instead of inheriting from `SVC`. The inheritance from `SVC` in
+    the `PegasosQSVC` class is removed. To keep the `score` method, inheritance
+    to the mixin class `ClassifierMixin` from scikit-learn is added.
+
```
adekusar-drl, line 8: I would also add a section `upgrade` and explain the inheritance issue and what users should do in this case.

adekusar-drl (commented):
Looks good, I'd emphasize the changes in the `upgrade` section of the release note.

releasenotes/notes/independent-pegasosQSVC-495100889ef28098.yaml
```diff
@@ -0,0 +1,17 @@
...
+upgrade:
+  - |
+    Inheriting from `sklearn.svm.SVC` resulted in errors when calling some inherited
+    methods such as `decision_function` due to the overridden `fit` implementation. 
+    For that reason, the inheritance has been replaced by a much lighter inheritance
+    from `ClassifierMixin` providing the `score` method and a new method `decision_function`
+    has been implemented. The class is still `sklearn` compatible due to duck typing.
+    This means that for the user everything that has been working in the previous
+    release still works. The only methods that are no longer supported (such as `predict_proba`)
+    were only raising errors in the previous release in practice.
```
adekusar-drl, line 17: These two sections may be quite far apart from each other in the final release notes. So, they should be clear on their own.

```suggestion
    Inheriting `PegasosQSVC` from `sklearn.svm.SVC` resulted in errors when calling some 
    inherited methods such as `decision_function` due to the overridden `fit` implementation. 
    For that reason, the inheritance has been replaced by a much lighter inheritance
    from `ClassifierMixin` providing the `score` method and a new method `decision_function`
    has been implemented. The class is still `sklearn` compatible due to duck typing.
    This means that for the user everything that has been working in the previous
    release still works, except the inheritance. The only methods that are no longer 
    supported (such as `predict_proba`) were only raising errors in the previous release 
    in practice.
```

adekusar-drl:
@woodsp-ibm is there anything left on your side in this PR?

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -27,11 +27,12 @@
...
 
-class PegasosQSVC(SVC, SerializableModelMixin):
-    """
+class PegasosQSVC(ClassifierMixin, SerializableModelMixin):
+    r"""
     This class implements Pegasos Quantum Support Vector Classifier algorithm developed in [1]
-    and includes overridden methods ``fit`` and ``predict`` from the ``SVC`` super-class. This
-    implementation is adapted to work with quantum kernels.
+    and includes methods ``fit``, ``predict`` and ``decision_function`` following the signatures
+    of `sklearn.svm.SVC <https://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html>`.
```
woodsp-ibm, line 34:
```suggestion
    of `sklearn.svm.SVC <https://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html>`_.
```
This does not format out as expected as a link - missing the trailing underscore

woodsp-ibm (approved):

PR #404, opened 2022-05-20
Get access to trained weights and optimization result

### LINKED ISSUE #385: Provide access to the trained weights in QNN models

Right now, if a user wants to get access to the trained weights, they must directly access a private field of the model, like here:
```python
trained_weights = classifier._fit_result.x
```
While it is still possible in python, it is not documented and breaks the model contract. Also, since a QNN for the model is provided by the user, I think, it is a good idea to have public access to the trained weights and may be to other properties as well once the model is fit.

adekusar-drl:
### Summary
In this PR we introduce two properties in the `TrainableModel` class. All algorithms/models like `NeuralNetworkClassifier`/`NeuralNetworkRegressor`/`VQC` that inherit from this class benefit of this PR:
- `fit_result` contains a result of the optimization process
- `weights` contains trained weights.

Both properties are available only once models are fit. Closes #385


releasenotes/notes/weights-dae0759ea750cffd.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Added two new properties to the `TrainableModel` class:
+      - ``fit_result`` returns a resulting object from the optimization procedure. Please refers to
```
woodsp-ibm, line 5: We could make Trainable model a link, or perhaps instead both the `fit-result` an `weights` be links to the new properties themselves so that a reader of the release note can go direct there.

qiskit_machine_learning/algorithms/trainable_model.py
```diff
@@ -141,9 +142,42 @@ def initial_point(self, initial_point: np.ndarray) -> None:
...
+        """
+        self._is_fit()
+        return np.asarray(self._fit_result.x)
+
+    @property
+    def fit_result(self) -> OptimizerResult:
+        """
+        Returns:
+            A resulting object from the optimization procedure. Please refer to the documentation of
+            the `OptimizerResult
```
woodsp-ibm, line 163: Similar issue to weights in regards of sphinx output

qiskit_machine_learning/algorithms/trainable_model.py
```diff
@@ -141,9 +142,42 @@ def initial_point(self, initial_point: np.ndarray) -> None:
...
+        """
+        Returns:
+            an array of trained weights. This property is for convenience only, it does the same
+            as `model.fit_result.x`.
+
+        Raises:
+            QiskitMachineLearningError: If the model has not been fit.
+        """
+        self._is_fit()
+        return np.asarray(self._fit_result.x)
```
woodsp-ibm, line 156: "This property is for convenience only, it does the same as `model.fit_result.x`." I guess I have to as why not here then why not simply do `return self.fit_result.x` since that would do the fit check and avoid it being replicated here. Maybe you want to do the np.asarray so ``return np.asarray(self.fit_result.x)`  but then I'd argue " it does the same as `model.fit_result.x`."" is not exactly right -   since doing the np,asarray can change things given .x as per OptimizerResult, is a `float | np.ndarray`

qiskit_machine_learning/algorithms/trainable_model.py
```diff
@@ -95,7 +96,7 @@ def __init__(
         self.optimizer = optimizer
 
         self._warm_start = warm_start
-        self._fit_result = None
+        self._fit_result: OptimizerResult | None = None
```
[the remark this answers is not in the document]
adekusar-drl: Let me think if this issue can be addressed without changing the algorithms a lot.

qiskit_machine_learning/algorithms/trainable_model.py
```diff
@@ -141,9 +142,42 @@ def initial_point(self, initial_point: np.ndarray) -> None:
...
+
+        Raises:
+            QiskitMachineLearningError: If the model has not been fit.
+        """
+        self._is_fit()
+        return self._fit_result
+
+    def _is_fit(self) -> bool:
+        if self._fit_result is None:
+            raise QiskitMachineLearningError("The model has not been fit yet")
```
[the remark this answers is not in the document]
adekusar-drl: Updated.

releasenotes/notes/weights-dae0759ea750cffd.yaml
adekusar-drl: Added links.

qiskit_machine_learning/algorithms/trainable_model.py
```diff
@@ -141,9 +142,42 @@ def initial_point(self, initial_point: np.ndarray) -> None:
         """Sets the initial point"""
         self._initial_point = initial_point
 
+    @property
+    def weights(self) -> np.ndarray:
+        """
+        Returns:
+            an array of trained weights. This property is for convenience only, it does the same
+            as `model.fit_result.x`.
+
```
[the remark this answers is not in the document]
adekusar-drl: Updated

qiskit_machine_learning/algorithms/trainable_model.py
adekusar-drl: Updated

qiskit_machine_learning/algorithms/trainable_model.py
adekusar-drl: Rephrased, please check it out if the text is clear enough.

woodsp-ibm: Seems fine - as I mentioned this could have been just one line `return np.asarray(self.fit_result.x)` since going via the property would do a fit check avoiding this needing to be done here too. But I know you prefer to access the instance vars more directly, so I guess you prefer it as it is. Tha'ts fine too.

woodsp-ibm (approved):
There is one open question, that you wanted to think about. But I approved this on the basis that if that seems an improvement that is desired and can be done it can be another PR as its really a different subject than this addressed.

adekusar-drl:
I'm going to merge this PR and address the question in the next one, perhaps with other improvements.

Issue #394, opened 2022-05-02
Unable to set custom gradient method in `QGAN` generator


ElePT:
### Environment

- **Qiskit Machine Learning version**: 0.4.0
- **Python version**: 3.10.0
- **Operating system**: macOS Monterrey (12.2.1)

### What is happening?

When specifying a custom gradient in the QGAN generator, a `ValueError` gets raised during training.

### How can we reproduce the issue?

On [04_qgans_for_loading_random_distributions.ipynb](https://github.com/Qiskit/qiskit-machine-learning/blob/main/docs/tutorials/04_qgans_for_loading_random_distributions.ipynb), try to set a custom gradient with either finite differences or parameter shift method, by typing the following code into the **Initialize the qGAN** section:

```
from qiskit.opflow import Gradient

custom_grad = Gradient("fin_diff")
# custom_grad = Gradient("param_shift")

# Set quantum generator
qgan.set_generator(generator_circuit=g_circuit, generator_init_params=init_params,
                   generator_gradient=custom_grad
                   )
```

This cell should run without issues, but if you continue with the tutorial, the cell with:

```
result = qgan.run(quantum_instance)
```

Will raise the following error:

```
ValueError: shapes (3,1) and (4,4) not aligned: 1 (dim 1) != 4 (dim 0)
```

### What should happen?

The user should be able to set a custom gradient method. 

### Any suggestions?

I did not have time to look into the issue, but the error traceback points to the loss calculation in [quantum_generator.py](https://github.com/Qiskit/qiskit-machine-learning/blob/main/qiskit_machine_learning/algorithms/distribution_learners/qgan/quantum_generator.py)

The exposed traceback is the following:

```
ValueError                                Traceback (most recent call last)
File ~/Desktop/qiskit-machine-learning/qiskit_machine_learning/algorithms/distribution_learners/qgan/quantum_generator.py:353, in QuantumGenerator.loss(self, x, weights)
    351 try:
    352     # pylint: disable=no-member
--> 353     loss = (-1) * np.dot(np.log(x).transpose(), weights)
    354 except Exception:  # pylint: disable=broad-except

File <__array_function__ internals>:5, in dot(*args, **kwargs)

ValueError: shapes (1,3) and (4,4) not aligned: 3 (dim 1) != 4 (dim 0)

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
Input In [120], in <cell line: 2>()
      1 # Run qGAN
[...]

Issue #393, opened 2022-05-02
`QGAN` generator training issue with `qasm_simulator` backend


ElePT:
### Environment

- **Qiskit Machine Learning version**: 0.4.0
- **Python version**: 3.10.0
- **Operating system**: macOS Monterrey (12.2.1)

### What is happening?

When running the qGAN tutorial code with `backend=qasm_simulator` and fixed seeds (simulator/transpiler), the generator parameters do not change between iterations, nor does the relative entropy (as seen in the plot), and the final distribution does not match the results from `backend=statevector _simulator`. This is overseen by the unit tests due to the bug reported in #392. 

### How can we reproduce the issue?

To reproduce the issue, run the notebook [04_qgans_for_loading_random_distributions.ipynb](https://github.com/Qiskit/qiskit-machine-learning/blob/main/docs/tutorials/04_qgans_for_loading_random_distributions.ipynb) with:
```
# Set quantum instance to run the quantum generator
quantum_instance = QuantumInstance(
    backend=BasicAer.get_backend("qasm_simulator")
    , seed_transpiler=seed, seed_simulator=seed
)
```
instead of:

```
# Set quantum instance to run the quantum generator
quantum_instance = QuantumInstance(
    backend=BasicAer.get_backend("statevector_simulator")
   , seed_transpiler=seed, seed_simulator=seed
 )
```
And check the plots for the relative entropy and final distribution to see that they don't match the statevector plots. 

You can also note that if you do not sed the seeds of the quantum instance, i.e:

```
# Set quantum instance to run the quantum generator
quantum_instance = QuantumInstance(
     backend=BasicAer.get_backend("statevector_simulator")
 )
```

The relative entropy is not constant anymore, but the learned distribution is still not correct.

### What should happen?

The results from training with different backends should match to a certain extent, and the relative entropy should not be constant.

### Any suggestions?

This issue has been discussed with @Zoufalc and @ebermot, the solution is not clear. So far, we have discovered that increasing the batch size (i.e setting `N=10000`, `batch_size = 5800`) and tuning the gradient optimizer with some "crazy" values, such as:

```
optimizer = ADAM(
                 maxiter=1,
                 tol=1e-6,
                 lr=1e-2,
                 beta_1=0.7,
                 beta_2=0.99,
                 noise_factor=1e-5,
                 eps=1, # not a realistic value
                 amsgrad=True,
             )
```

[...]

PR #405, opened 2022-05-23
Fix qGAN custom gradients

ebermot:
### Summary

This pull request is solving the issues #393 and #394 : previously, a custom gradient method could not be used for the generator of the `QGAN` algorithm without creating a dimensionality incompatibility.

### Details and comments

In the `quantum_generator` class of the `QGAN` algorithm, in order to use a `qasm_simulator`, a custom gradient method had to be used (as Parameter shift). However, due to the sparse structure of the dictionary output, it was impossible to compute the loss function. Therefore, the result output was converted into a dense array. Plus, to solve the issue #394 , the gradient method was not being properly used : it was always the `statevector_simulator` backend which was used to compute the gradient. Therefore, a `CircuitSampler` structure was used to solve this issue.


releasenotes/notes/custom-gradients-35923fa70af97d8d.yaml
```diff
@@ -0,0 +1,20 @@
+---
+
+fixes:
+  - |
+    This release note solves two issues : 
+    - Previously in the :class:`~qiskit_machine_learning.algorithms.QuantumGenerator` of 
+    the :class:`~qiskit_machine_learning.algorithms.QGAN` algorithm, if we used
+    a simulator other than the `statevector_simulator` the result dictionary had not the 
+    correct size to compute both the gradient and the loss functions. Now,
```
woodsp-ibm, line 9:
```suggestion
    This release note solves two issues : 
    - Previously in the :class:`~qiskit_machine_learning.algorithms.QuantumGenerator` of 
    the :class:`~qiskit_machine_learning.algorithms.QGAN` algorithm, if we used
    a simulator other than the `statevector_simulator` the result dictionary had not the 
    correct size to compute both the gradient and the loss functions. Now, 
```

releasenotes/notes/custom-gradients-35923fa70af97d8d.yaml
```diff
@@ -0,0 +1,20 @@
+---
+
+fixes:
+  - |
+    This release note solves two issues : 
+    - Previously in the :class:`~qiskit_machine_learning.algorithms.QuantumGenerator` of
```
woodsp-ibm, line 6:
```suggestion
    This release note solves two issues : 

    - Previously in the :class:`~qiskit_machine_learning.algorithms.QuantumGenerator` of 
```
There needs to be blank line preceding any list definition for Sphinx otherwise all the text is wrapped up as a single paragraph as you can see from this output
image

qiskit_machine_learning/algorithms/distribution_learners/qgan/quantum_generator.py
```diff
@@ -417,10 +421,23 @@ def gradient_function(current_point):
...
+                    CircuitSampler(quantum_instance.backend).convert(grad_object, value_dict).eval()
+                )
+                if quantum_instance.is_statevector:
+                    analytical_gradients = np.array(grad_object)
+                else:
+                    analytical_gradients = np.zeros((len(grad_object), grad_object[0].shape[1]))
+                    for i, _ in enumerate(grad_object):
+                        sparse_gradients = grad_object[i].tocoo()
+                        analytical_gradients[i, sparse_gradients.col] = sparse_gradients.data
+
```
woodsp-ibm, line 438: Seems strange to me to have a blank line here in the code.

test/algorithms/distribution_learners/qgan/test_qgan.py
```diff
@@ -377,6 +377,42 @@ def test_qgan_gradient_penalty_pytorch(self):
...
+    def test_qgan_training_custom_gradients_statevector(self, gradient_method: str):
+        """
+        Test qGAN different gradient methods for the statevector simulator
+
+        Args:
+            gradient_method (str): custom gradient method
+        """
+        if gradient_method is None:
+            gradient_method = "param_shift"
+
```
woodsp-ibm, line 390: This code seems unnecessary. The typehint in the method says just 'str' and does not include Optional. All values are passed to the test by ddt anyway so it should never be done unless someone changes something. I think if you wanted to guard against someone altering the ddt params or whatever then instead do a self.skipTest("gradient method must be provided") or something like this. Beyond if you want to default gradient_method to "param_shift" I think it would be better just to put it in the parameter as a default value.

test/algorithms/distribution_learners/qgan/test_qgan.py
```diff
@@ -377,6 +377,42 @@ def test_qgan_gradient_penalty_pytorch(self):
...
+    def test_qgan_training_custom_gradients_qasm(self, gradient_method: str):
+        """
+        Test qGAN different gradient methods for the qasm simulator
+
+        Args:
+            gradient_method (str): custom gradient method
+        """
+        if gradient_method is None:
+            gradient_method = "param_shift"
+
```
woodsp-ibm, line 408: Similar comment to above

woodsp-ibm:
This change seems to be to properly support counts based returns, i.e. not statevector, so there is no impact on the output when using a statevector i.e. its the same as before. So this notebook will still output the exact same result, correct? https://github.com/Qiskit/qiskit-machine-learning/blob/main/docs/tutorials/04_qgans_for_loading_random_distributions.ipynb

ebermot:
Using `statevector_simulator` with this notebook could output the same result with this change, but it will yield an error if you specify a `Gradient` method when initializing the generator.

ebermot:
Yes exactly, both the problem with the custom gradients and with the non-evolving relative entropy occurred when using a simulator other than `statevector_simulator`

ebermot:
Not exactly, because the error didn't appear when launching the notebook when we used `statevector_simulator` and without specifying any `Gradient` method. However, if we select one `Gradient` like `param_shift` or `lin_comb`, this is when we actually get an error due to dimensionality incompatibility. The reason why we didn't see it before is because when we don't specify any `generator_gradient`, we never enter the function `convert_to_grad` in `quantum_generator` and so not enabling the computation of the gradient/loss function with the specify method. When we don't enter this function, the gradient is computed using the optimizer method to do it and so never yielding any computational error.

qiskit_machine_learning/algorithms/distribution_learners/qgan/quantum_generator.py
```diff
@@ -417,10 +421,22 @@ def gradient_function(current_point):
                 quantum_instance, params=current_point, shots=self._shots
             )
             prediction_generated = discriminator.get_label(generated_data, detach=True)
-            op = ~CircuitStateFn(primitive=self.generator_circuit)
+            op = CircuitStateFn(primitive=self.generator_circuit)
             grad_object = gradient_object.convert(operator=op, params=free_params)
             value_dict = {free_params[i]: current_point[i] for i in range(len(free_params))}
-            analytical_gradients = np.array(grad_object.assign_parameters(value_dict).eval())
+            if quantum_instance.backend is not None:
```
adekusar-drl, line 427: According to the hierarchy of calls, `quantum_instance` can be `None` here. Should it be a test like `if quantum_instance is not None:`?

qiskit_machine_learning/algorithms/distribution_learners/qgan/quantum_generator.py
```diff
@@ -417,10 +421,22 @@ def gradient_function(current_point):
...
+                grad_object = (
+                    CircuitSampler(quantum_instance.backend).convert(grad_object, value_dict).eval()
```
adekusar-drl, line 429: In general you can simplify the call to `CircuitSampler(quantum_instance)`.

qiskit_machine_learning/algorithms/distribution_learners/qgan/quantum_generator.py
```diff
@@ -417,10 +421,22 @@ def gradient_function(current_point):
...
+                )
+                if quantum_instance.is_statevector:
+                    analytical_gradients = np.array(grad_object)
```
adekusar-drl, line 432: Preferable way is to use `np.asarray()`. This function does not copy data if possible and a call to `np.array()` like here copies data.
```suggestion
                    analytical_gradients = np.asarray(grad_object)
```

qiskit_machine_learning/algorithms/distribution_learners/qgan/quantum_generator.py
```diff
@@ -417,10 +421,22 @@ def gradient_function(current_point):
...
+                else:
+                    analytical_gradients = np.zeros((len(grad_object), grad_object[0].shape[1]))
+                    for i, _ in enumerate(grad_object):
+                        sparse_gradients = grad_object[i].tocoo()
+                        analytical_gradients[i, sparse_gradients.col] = sparse_gradients.data
```
adekusar-drl, line 437: Could it be simpler?
```suggestion
                        analytical_gradients[i, :] = grad_object[i].todense()
```

adekusar-drl (commented):
Thanks a lot for taking care of QGANs, really appreciated! I left a few comments, mostly about how to simplify the code. I leave to @woodsp-ibm to decide whether these suggestions are essential or not since he reviewed the PR first.

qiskit_machine_learning/algorithms/distribution_learners/qgan/quantum_generator.py
ebermot: The reason why it was made like this was to avoid the use of any structure like `todense()` which could scale up poorly for a high number of qubits. Here I used the whole structure without having to use right after the `CircuitSampler` , `to_matrix_op()`. If the scalability using `todense()` is okay, then yes it can be simpler!

adekusar-drl: I understand, but you collect rows from `grad_object` into `analytical_gradients` which is dense by creation! So, there's no reason to keep intermediate object as sparse objects. Correct me if I missed anything here.

adekusar-drl (approved):
@woodsp-ibm I'm fine with the PR.

PR #406, opened 2022-05-26
Enhancements in the QNN algorithms

adekusar-drl:
### Summary
This PR addresses a question raised here https://github.com/Qiskit/qiskit-machine-learning/pull/404#pullrequestreview-985078287 and introduces some other improvements.
- Now `TrainableModel.fit()` is not abstract and delegates actual fit to sub-classes.
- Fixed broken `warm_start` support in `NeuralNetworkRegressor`
- Some tests are optimized, new added.


test/algorithms/regressors/test_neural_network_regressor.py
```diff
@@ -122,6 +110,24 @@ def callback(objective_weights, objective_value):
             callback=callback,
         )
 
+        return regressor, regression_opflow_qnn, ansatz
+
+    @idata(itertools.product(OPTIMIZERS, QUANTUM_INSTANCES, CALLBACKS))
+    @unpack
+    def test_regressor_with_opflow_qnn(self, opt, q_i, cb_flag):
+        """Test Neural Network Regressor with Opflow QNN (Two Layer QNN)."""
+        if cb_flag is True:
```
woodsp-ibm, line 119: Arguably this would be better as simply `if cb_flag` since we really only need a flag that evaluates to True - if I set CALLBACKS to [1, 0] even though they evaluate to True, False, since it does not have the actual True object (singleton) there it would fail to ever test the callback.

woodsp-ibm, line 119: Arguably this would be better as simply `if cb_flag` since we really only need a flag that evaluates to True - if I set CALLBACKS to [1, 0] even though they evaluate to True, False, since it does not have the actual True object (singleton) there it would fail to ever test the callback.

woodsp-ibm (commented):
I am thinking that perhaps we should have a release note. On the off chance that any 3rd party have some implementation based on TrainingModel it will currently fail right since though they would have a fit method, they would not have implemented a _fit_internal which is now required.

test/algorithms/regressors/test_neural_network_regressor.py
adekusar-drl: Replaced with `if cb_flag`. This is the old code.

adekusar-drl:
> I am thinking that perhaps we should have a release note. On the off chance that any 3rd party have some implementation based on TrainingModel it will currently fail right since though they would have a fit method, they would not have implemented a _fit_internal which is now required.

Added reno.

adekusar-drl:
@woodsp-ibm A general comment. While right now introduction of `weights` and `fit_result` looks ok, it may be not so good idea in general. For instance, there may be a different type for `fit_result`, or there may be a more complex structure instead of plain `fit_result`. Same for weights. I don't have a good solution right away, but what we have now may change.

woodsp-ibm (approved):

PR #411, opened 2022-06-09
Enable default feature map for 1 qubit in the ``TwoLayerQNN`` (+ other small improvements)

Cryoris:
### Summary

Enable the default construction of the ``ZFeatureMap`` in the ``TwoLayerQNN` if the number of qubits is 1. Previously, not providing a feature map for the single qubit case raised an error as default construction assumed 2 or more qubits.

### Details and comments

This PR also includes some additional small improvements:
* Fix typehints in the initializer (`Optional` was missing in some places)
* Fix formatting in docs (math and code-blocks were not properly displayed) 
image
* Use ``if thing is not None`` instead of ``if thing`` for safety, this came back to bite us in the tail a few times in the past
* Improve the error messages to describe a bit more clearly what's wrong in the initializer
* Avoid using `~StateFn(op)` and use `is_measurement` flag instead to avoid complex conjugating


qiskit_machine_learning/neural_networks/two_layer_qnn.py
```diff
@@ -111,11 +124,11 @@ def __init__(
 
         # construct observable
         self.observable = (
-            observable if observable else PauliSumOp.from_list([("Z" * num_qubits_, 1)])
+            observable if observable is not None else PauliSumOp.from_list([("Z" * num_qubits_, 1)])
         )
 
         # combine all to operator
-        operator = ~StateFn(self.observable) @ StateFn(self._circuit)
+        operator = StateFn(self.observable, is_measurement=True) @ StateFn(self._circuit)
```
adekusar-drl, line 131: Does the code do the same? A bit non-obvious replacement.

Cryoris: No it doesn't do exactly the same, the new one is correct while the old one isn't 😉 

If you use `~` to indicate a measurement, then the operator will wrongly be complex conjugated. That doesn't happen if you directly use `is_measurement=True`. Since most our observables are hermitian that's usually not a problem but if you have a non-hermitian observable for some reason (e.g. `iX`) then it silently gives wrong results.

qiskit_machine_learning/neural_networks/two_layer_qnn.py
```diff
@@ -63,37 +64,49 @@ def __init__(
...
+        if num_qubits is not None:
             num_qubits_ = num_qubits
-            if feature_map:
+            if feature_map is not None:
                 if feature_map.num_qubits != num_qubits:
-                    raise QiskitMachineLearningError("Incompatible num_qubits and feature_map!")
+                    raise QiskitMachineLearningError(
+                        f"Mismatching num_qubits ({num_qubits}) and number of qubits "
+                        f"in the feature_map ({feature_map.num_qubits})!"
+                    )
```
woodsp-ibm, line 79: For the feature map and anstaz, if these are blueprint ccts we could set the num qubits similar to what is done in VQE right where of course we cannot do this for plain ccts. Just wondering if perhaps it could be more flexible in this regard.

Cryoris: Yeah I think we could!

dlasecki:
It looks good to me.

qiskit_machine_learning/neural_networks/two_layer_qnn.py
adekusar-drl: @woodsp-ibm Can you please point me out to the code in VQE you are referring to? I don't fully understand where the problem is. Otherwise, I fine with the PR.

woodsp-ibm: @adekusar-drl  There is not a "problem" per-se more I suggesting, like we do in VQE, that potentially here we could adjust the circuit to the number of qubits where possible rather than simply throwing an exception. If you know the number of qubits then one could argue, that if you build out a circuit, which is a blueprint circuit, that you as a caller can also set its number of qubits so it's matched. Since this was often not known in VQE, when the constructor was given an ansatz circuit, we set it later based on the operator, where possible, to match. https://github.com/Qiskit/qiskit-terra/blob/573198bdc55eaf4ad0691d441ae41c874f46313d/qiskit/algorithms/minimum_eigen_solvers/vqe.py#L308-L320

adekusar-drl: Oh I see now. Yes, this can be done as in VQE, but I suggest to move to a separate PR and merge this one. In VQC we can introduce the same improvement. I can take care of your proposal.

PR #419, opened 2022-06-18
Grammar mistake in 07_pegasos_qsvc.ipynb

sitong1011:
### Summary
A minor grammar mistake, but may lead to misunderstanding.

### Details and comments
In 07_pegasos_qsvc.ipynb, there's a sentence "This algorithm is ... beneficial of the kernel trick, and yields a training complexity...", but I didn't find any sentence examples using "beneficial of", I suppose it means "benefit from" here. It's a minor grammar mistake, but since it causes some misunderstandings, I think it would be better to fix it.


adekusar-drl:
@woodsp-ibm I'd rely on your knowledge here. Thanks.

woodsp-ibm:
Usually its something is `beneficial for` or `beneficial to` when doing something. I get what was intended, the proposed text does seem clearer though as in "A benefits from B".

As an FYI for the author CI it appears is failing at present since an new pylint was recently released and its finding things that were not flagged before. This will be fixed and once things are corrected the this should pass fine I think.

manoelmarques:
Doesn't this grammar mistake happen on `main` ? Why is it being fixed on stable ? It should be fixed on a PR in `main` first and then the fix would be backported to 'stable` otherwise, next release it will show up again.

sitong1011:
@manoelmarques Thank you! I think I haven't noticed that when I created the pull request. I have changed the base branch of this pull request to `main`, I don't know if it will work.
Besides this, I noticed that there are still some checks that failed, I wonder is there anything I should do about that?

woodsp-ibm (approved):

PR #424, opened 2022-06-24
Fix FakeToronto import and Fix Aer Deprecate Messages

manoelmarques:
### Summary
Change FakeToronto import, fix Aer deprecate messages, all due to recent Terra changes.

Related to QIskit/qiskit-terra#8121 and Qiskit/qiskit-terra#5089


adekusar-drl (approved):
Would be nice to have references to the changes in Terra.

PR #432, opened 2022-07-06
Evaluate quantum kernel matrix elements for identical samples

### LINKED ISSUE #425: Quantum kernel - missing indices in regression tasks
### Environment

- **Qiskit Machine Learning version**:  0.4.0
- **Python version**:  3.9.7 64-bit
- **Operating system**:  Windows 10 Enterprise Version 21H2

### What is happening?

The bug arises in the Qiskit QuantumKernel class (more precisely in the evaluation routine) when performing a regression task with the quantum kernel (SVR,...) and **not** using the statevector simulator.
It just happens if a non-symmetric kernel matrix is computed. 
When some training- and testing points of the regression happen to be the same point the bug prevents the inclusion of the entries in the list of the "to_be_computed_data_pair"'s - which results in a wrong kernel entry of 0., and subsequently leading to the wrong regression result - since the kernel entry of two identical points should be 1.

Create a simple test function for the regression, where a kernel evaluated on a quantum computer is used.
```python
# General Imports
import numpy as np

# Scikit Imports
from sklearn.preprocessing import StandardScaler

# Qiskit Imports
from qiskit import Aer
from qiskit.circuit import QuantumCircuit, ParameterVector
from qiskit_machine_learning.kernels import QuantumKernel

# backends
statevec = Aer.get_backend('statevector_simulator')
qasm = Aer.get_backend('qasm_simulator')

# Function to regress: in this example x*sin(x)
X = np.linspace(start=0, stop=10, num=1_00).reshape(-1, 1)
y = np.squeeze( X*np.sin(X) )

rng = np.random.RandomState(10)
training_indices = rng.choice(np.arange(y.size), size=20, replace=False)
training_indices = np.sort(training_indices)
X_train, y_train = X[training_indices], y[training_indices]
print(training_indices)
# in this case 20 of our training points gonna be similar to testing points
# Scaling of input feature
scaler = StandardScaler()
X_scaled = scaler.fit_transform(X)
y_scaled = scaler.fit_transform(y.reshape(-1, 1))
X_train = X_scaled[training_indices]
y_train = y_scaled[training_indices]
# artificial feature expansion
# stack the training and test points to the number of qubits used in the encoding
XX = np.column_stack((X_scaled,X_scaled))
X4_test = np.column_stack((XX,XX))
XX_train = np.column_stack((X_train,X_train))
X4_train = np.column_stack((XX_train,XX_train))

# define data encoding (a custom one is used in this case):
#custom parameterized quantum circuit
def PQC(qubits,layers,c):
    
    x = ParameterVector('x', length=qubits)
    theta = ParameterVector('θ',length=2*qubits)
    
    theta_range = np.linspace(start=0, stop=2*np.pi, num=2*qubits)
    rand_idx = np.random.choice(np.arange(theta_range.size), size=2*qubits, replace=False)
    theta_sample = theta_range[rand_idx]
    var_custom = QuantumCircuit(qubits)
    counter = 0

    for j in range(layers):
        for i in range(qubits):
            if i != 0:
                counter += 1
            var_custom.ry(c*x[i] + theta[i+counter],i)
            var_custom.rz(c*x[i] + theta[i+1+counter],i)
            if i == qubits-1:
                counter = 0
        
        if (j % 2) == 0:
            for i in range(qubits-1):
                if(i % 2) == 0:
                    var_custom.cx(i, i+1)
        else:
            for i in range(qubits-1):
                if(i % 2) == 1:
                    var_custom.cx(i, i+1)
    var_custom = var_custom.bind_parameters({theta: theta_sample})
    return var_custom

# use a feature map with 4 qubits to built up the quantum kernel (build two kernels - one for the statevector one with the qasm to compare the kernel entries and show the error):
pqc_map = PQC(qubits=4,layers=2,c=1.0)
quantum_kernel_statevec = QuantumKernel(feature_map=pqc_map, quantum_instance=statevec)
quantum_kernel_qasm = QuantumKernel(feature_map=pqc_map, quantum_instance=qasm)

# evaluate the non-symmetric test-training kernel matrix for both backends and compare the entries to see the error with the 0. entries (where there should be a 1.) in the kernel evaluated with the qasm simulator:
K_test_train_statevec = quantum_kernel_statevec.evaluate(x_vec=X4_test, y_vec=X4_train)
K_test_train_qasm = quantum_kernel_qasm.evaluate(x_vec=X4_test, y_vec=X4_train)

print(K_test_train_statevec)
print(K_test_train_qasm)
```

### What should happen?

In the Qiskit QuantumKernel class there is a line in the code which prevents the inclusion of the data pair's where the data points have the same value (for noin-symmetric kernel matrices). 
These have to be 1. (or with slightly tolerance around 1. if a noisy backend is used). 
The following code is copied out of the QuantumKernel class (evaluate function) with a comment over the line where the bug lays (in my opinion):
```python
else:  # not using state vector simulator
            feature_map_params_x = ParameterVector("par_x", self._feature_map.num_parameters)
            feature_map_params_y = ParameterVector("par_y", self._feature_map.num_parameters)
            parameterized_circuit = self.construct_circuit(
                feature_map_params_x,
                feature_map_params_y,
                measurement=measurement,
                is_statevector_sim=is_statevector_sim,
            )
            parameterized_circuit = self._quantum_instance.transpile(
                parameterized_circuit, pass_manager=self._quantum_instance.unbound_pass_manager
            )[0]

            for idx in range(0, len(mus), self._batch_size):
                to_be_computed_data_pair = []
                to_be_computed_index = []
                for sub_idx in range(idx, min(idx + self._batch_size, len(mus))):
                    i = mus[sub_idx]
                    j = nus[sub_idx]
                    x_i = x_vec[i]
                    y_j = y_vec[j]
                    #########  here occurs the little bug for non-symmetric matrices  #####
                    if not np.all(x_i == y_j):
                        to_be_computed_data_pair.append((x_i, y_j))
                        to_be_computed_index.append((i, j))

                circuits = [
                    parameterized_circuit.assign_parameters(
                        {feature_map_params_x: x, feature_map_params_y: y}
                    )
                    for x, y in to_be_computed_data_pair
                ]]
```

### Any suggestions?

Simply comment out the line `if not np.all(x_i == y_j)` in the evaluate function attached above.
Then all indices of non-symmetric matrices are included in the inner product computation and the regression method of choice is going to produce valid results with a quantum kernel.

adekusar-drl:
### Summary
Introduced a new parameter `evaluate_duplicates` in `QuantumKernel`. This parameter defines a strategy how kernel matrix elements are evaluated if identical samples are found. Possible values are:
- ``all`` means that all kernel matrix elements are evaluated, even the diagonal ones when training. This may introduce additional noise in the matrix.
- ``non_diagonal`` when training the matrix diagonal is set to `1`, the rest elements are fully evaluated, e.g., for two identical samples in the dataset. When inferring, all elements are evaluated. This is the default value.
 - ``none`` when training the diagonal is set to `1` and if two identical samples are found in the dataset the corresponding matrix element is set to `1`. When inferring, matrix elements for identical samples are set to `1`.

This PR fixes #425, #336.


qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -60,6 +61,7 @@ def __init__(
         batch_size: int = 900,
         quantum_instance: Optional[Union[QuantumInstance, Backend]] = None,
         training_parameters: Optional[Union[ParameterVector, Sequence[Parameter]]] = None,
+        evaluate_duplicates: str | None = "non_diagonal",
```
woodsp-ibm, line 64: Should this just be of type `str`? If I pass in `None` it looks like it will work, since it does a `str(None)` which ends up as `"None"` and then lowercase makes it the same as passing `"none"` in. I guess having both I find confusing - I mean the None/Optional aspect here and the "none" documented string value. (In the unit tests you always pass in "none" as a string.)

On a side note, would we want to mix/match in the same function both the Union the new | way of doing things. I think it would be better to be consistent at least at this level, though I realize we are moving towards the newer way of doing things where some modules may now be in new style.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -72,6 +74,21 @@ def __init__(
             training_parameters: Iterable containing ``Parameter`` objects which correspond to
                  quantum gates on the feature map circuit which may be tuned. If users intend to
                  tune feature map parameters to find optimal values, this field should be set.
+            evaluate_duplicates: Defines a strategy how kernel matrix elements are evaluated if
+                identical samples are found. Possible values are:
+
+                    - ``all`` means that all kernel matrix elements are evaluated, even the diagonal
+                        ones when training. This may introduce additional noise in the matrix.
+                    - ``non_diagonal`` when training the matrix diagonal is set to `1`, the rest
```
woodsp-ibm, line 82: I was wondering whether you had considered `off_diagonal` - or maybe thats not quite what you wanted. Since it evaluates everything off the diagonal it seemed to fit for me.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -559,15 +583,24 @@ def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
...
 
             offset = 0 if is_symmetric else len(x_vec)
-            matrix_elements = [
-                self._compute_overlap(idx, statevectors, is_statevector_sim, measurement_basis)
-                for idx in list(zip(mus, nus + offset))
-            ]
+            for (
+                i,
+                j,
+            ) in zip(mus, nus):
```
woodsp-ibm, line 589: This is really how black wants this or is it an aspect of the trailing , after j that is causing this formatting. If I compare to something below (now deleted), which has more in it, that went on one line ok i.e. (line 567 below deleted)
>for i, j, value in zip(mus, nus, matrix_elements):

releasenotes/notes/fix-quantum-kernel-duplicates-b75d6ed8a0f37f60.yaml
```diff
@@ -0,0 +1,23 @@
...
+          training. This may introduce additional noise in the matrix.
+      - ``non_diagonal`` when training the matrix diagonal is set to `1`, the rest elements are
+          fully evaluated, e.g., for two identical samples in the dataset. When inferring, all
+          elements are evaluated. This is the default value.
+      - ``none`` when training the diagonal is set to `1` and if two identical samples are found
+          in the dataset the corresponding matrix element is set to `1`. When inferring, matrix
+          elements for identical samples are set to `1`.
+fixes:
+  - |
+    Fixed quantum kernel evaluation when duplicate samples are found in the dataset. Originally,
```
woodsp-ibm, line 19: `when duplicate samples` above, apart from the parameter `evaluate_duplicates` it uses `identical` to describe samples, Here it uses `duplicate` like the parameter does, but this is the only place. I guess I should read `duplicate` and `identical` as meaning the same thing - i.e. two or more samples having an identical value and thus being in essence duplicates of one another.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: 1. I wanted to show that `None` is a valid value as well. But I fine to keep just `str` as the type hint as well. Should I keep only `str`?
2. I changed the type hints across the class to the new way.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: Ah, you are totally right! Replaced with `off_diagonal`

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: This was a `black` artifact from the previous version. Now it is just one line `for (i, j) in zip(mus, nus):`. Thank you.

releasenotes/notes/fix-quantum-kernel-duplicates-b75d6ed8a0f37f60.yaml
adekusar-drl: That's right. I used both terms in the same meaning. I agree, I was not consistent. In terms of language and docstrings/parameters what is better?

adekusar-drl:
@ElePT while the PR is almost done, you review might be helpful.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -72,6 +74,21 @@ def __init__(
             training_parameters: Iterable containing ``Parameter`` objects which correspond to
                  quantum gates on the feature map circuit which may be tuned. If users intend to
                  tune feature map parameters to find optimal values, this field should be set.
+            evaluate_duplicates: Defines a strategy how kernel matrix elements are evaluated if
+                identical samples are found. Possible values are:
```
ElePT, line 78:
```suggestion
               duplicate samples are found. Possible values are:
```

releasenotes/notes/fix-quantum-kernel-duplicates-b75d6ed8a0f37f60.yaml
```diff
@@ -0,0 +1,23 @@
+---
+features:
+  - |
+    Introduced a new parameter `evaluate_duplicates` in
+    :class:`~qiskit_machine_learning.kernels.QuantumKernel`. This parameter defines a strategy how
+    kernel matrix elements are evaluated if identical samples are found.
```
ElePT, line 6:
```suggestion
    kernel matrix elements are evaluated if duplicate samples are found.
```

ElePT (commented):
The code looks good to me, except for the minor comments I added above.

releasenotes/notes/fix-quantum-kernel-duplicates-b75d6ed8a0f37f60.yaml
ElePT: I didn't find it confusing to have 2 terms, I think it's pretty clear that they refer to the same concept. I believe that maybe balancing them out and using duplicate a bit more should be enough (see suggestions below).

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -81,6 +98,12 @@ def __init__(
         self._enforce_psd = enforce_psd
         self._batch_size = batch_size
         self._quantum_instance = quantum_instance
+        eval_duplicates = str(evaluate_duplicates).lower()
```
ElePT, line 101:
```suggestion
        eval_duplicates = evaluate_duplicates.lower()
```

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -55,11 +56,12 @@ class QuantumKernel:
...
         self,
-        feature_map: Optional[QuantumCircuit] = None,
+        feature_map: QuantumCircuit | None = None,
         enforce_psd: bool = True,
         batch_size: int = 900,
-        quantum_instance: Optional[Union[QuantumInstance, Backend]] = None,
-        training_parameters: Optional[Union[ParameterVector, Sequence[Parameter]]] = None,
+        quantum_instance: QuantumInstance | Backend | None = None,
+        training_parameters: ParameterVector | Sequence[Parameter] | None = None,
+        evaluate_duplicates: str | None = "off_diagonal",
```
ElePT, line 64:
```suggestion
        evaluate_duplicates: str = "off_diagonal",
```

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: OK, let me summarize:

QuantumKernel() -> evaluate_duplicates = off_diagonal
QuantumKernel(evaluate_duplicates=None) -> evaluate_duplicates = none
QuantumKernel(evaluate_duplicates="none") -> evaluate_duplicates = none

This is how it works now and this is exactly what I wanted to achieve. If you think is a wrong way, let me know why.

woodsp-ibm: > I wanted to show that None is a valid value as well. But I fine to keep just str as the type hint as well. Should I keep only str?

My preference would be to get rid of None and have it just as `str` type taking "all", "off_diagonal" or "none". None would feel to me it ought to be something different, but it ends up as just a synonym for "none" if you will so to me dropping it seems the thing to do.

releasenotes/notes/fix-quantum-kernel-duplicates-b75d6ed8a0f37f60.yaml
woodsp-ibm: Yes, I would prefer to see duplicate used more in the descriptions when referring to elements managed by the evaluate_duplicates. 

I can see a duplicate as being another sample with an identical value as I mentioned above, but that begs another question I guess....
....when something is a duplicate is this an exact value match? - we have no tolerance here that can be set for when things are very very close that we consider them the same (a duplicate)?

adekusar-drl: Replaced with `duplicated`, thanks to Elena

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: OK, removed `None`.

woodsp-ibm (approved):

PR #433, opened 2022-07-06
QuantumKernel refactor

adekusar-drl:
### Summary
No new features, the `QuantumKernel` code is just split into smaller methods.


qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -355,22 +357,27 @@ def construct_circuit(
...
-            )
+        qc = self._construct_circuit_with_feature_map(x, y)
 
-        if len(x) != self._feature_map.num_parameters:
-            raise ValueError(
-                "x and class feature map incompatible dimensions.\n"
-                f"x has {len(x)} dimensions, but feature map has {self._feature_map.num_parameters}."
-            )
+        if y is None:
+            y = x
```
dlasecki, line 363:
```suggestion
        y = x if y is None
```
For compactness.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -380,38 +387,34 @@ def construct_circuit(
...
-                    f"y has {len(y)} dimensions, but feature map has {self._feature_map.num_parameters}."
-                )
+        return qc
 
-            if y is None:
-                y = x
+    # just a synonym
+    def _construct_circuit_statevector(self, x: ParameterVector, y: ParameterVector = None):
+        """This is just a synonym for ``_construct_circuit_with_feature_map``"""
+        return self._construct_circuit_with_feature_map(x, y)
```
dlasecki, line 395: Why do we need a synonym?

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -380,38 +387,34 @@ def construct_circuit(
...
 
-            y_dict = dict(zip(self._feature_map.parameters, y))
-            psi_y_dag = self._feature_map.assign_parameters(y_dict)
-            qc.append(psi_y_dag.to_instruction().inverse(), qc.qubits)
+    def _compute_overlap_statevector(self, idx: Tuple[int, int], results: List) -> float:
+        """
+        Helper function to compute overlap for given input if a statevector simulator is used.
+        """
+        # |<0|Psi^dagger(y) x Psi(x)|0>|^2, take the amplitude
+        v_a, v_b = [results[int(i)] for i in idx]
```
dlasecki, line 402: Is the typhint for idx correct? Casting a tuple onto int would not work...

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -441,188 +444,224 @@ def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
...
+            is_symmetric = False
+
+        # initialize kernel matrix
+        kernel = np.zeros((x_vec.shape[0], y_vec.shape[0]))
+
+        # set diagonal to 1 if symmetric
+        if is_symmetric:
+            np.fill_diagonal(kernel, 1)
+
+        is_statevector_sim = self._quantum_instance.is_statevector
```
dlasecki, line 470: How about we skip this assignment? The assigned variable is used only once.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -441,188 +444,224 @@ def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
...
+        # calculate kernel
+        if is_statevector_sim:
+            kernel = self._calculate_kernel_statevector(x_vec, y_vec, is_symmetric, kernel)
+        else:
+            kernel = self._calculate_kernel(x_vec, y_vec, is_symmetric, kernel)
+
+        return kernel
+
+    def _validate_input(self, x_vec: np.ndarray, y_vec: np.ndarray):
```
dlasecki, line 479: Return type hint missing.

dlasecki, line 479: Validate might suggest that vectors are not modified in the method but they might be.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -441,188 +444,224 @@ def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
...
+        Validates the size (length) of the vector against the number of parameters of the feature
+        map. Updates the feature map if required to the size of the vector.
 
-        # set diagonal to 1 if symmetric
-        if is_symmetric:
-            np.fill_diagonal(kernel, 1)
+        Args:
+            vec: A vector to validate.
+            vec_name: A vector name, if an error is raised it ise used in the error message.
+            adjust_feature_map: If ``True`` tried to adjust the number of qubits of the feature map.
```
dlasecki, line 519:
```suggestion
            adjust_feature_map: If ``True`` tries to adjust the number of qubits of the feature map.
```

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -441,188 +444,224 @@ def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
...
+        Raises:
+            ValueError: If the size of the vector is not the same as the number of parameters of the
+                feature map or when the feature map does not allow to change the number of qubits.
+        """
+        if vec is not None and len(vec) != self._feature_map.num_parameters:
+            raise_error = True
+            if adjust_feature_map:
+                try:
+                    # an attempt to update the feature map with the required number of qubits
+                    self._feature_map.num_qubits = vec
```
dlasecki, line 530: Why do we assign a vector to a field holding an int?

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -441,188 +444,224 @@ def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
...
+            if raise_error:
+                raise ValueError(
+                    f"{vec_name} and class feature map have incompatible dimensions.\n"
+                    f"{vec_name} has {len(vec)} dimensions, "
+                    f"but feature map has {self._feature_map.num_parameters}."
+                )
+
+    def _calculate_kernel_statevector(
+        self, x_vec: np.ndarray, y_vec: np.ndarray, is_symmetric: bool, kernel: np.ndarray
+    ):
```
dlasecki, line 544: Return type hint missing. Please check other places.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -441,188 +444,224 @@ def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
...
+                parameterized_circuit.assign_parameters({feature_map_params: x})
+                for x in to_be_computed_data[min_idx:max_idx]
+            ]
+            if self._quantum_instance.bound_pass_manager is not None:
+                circuits = self._quantum_instance.transpile(
+                    circuits, pass_manager=self._quantum_instance.bound_pass_manager
+                )
+            results = self._quantum_instance.execute(circuits, had_transpiled=True)
+            for j in range(max_idx - min_idx):
+                statevectors.append(results.get_statevector(j))
```
dlasecki, line 575: I would extract it to a method.

dlasecki (changes requested):
Performed the first scan and left some comments.
Overall, the class is very lengthy. Is there a way to break it down into smaller classes with clear and focused responsibilities?

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: `else` is required :)
could be `y = x if y is None else y` but I think it is better.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: I'd prefer to keep as it's now for consistency with `_calculate_kernel()`.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: Reviewed type hints here and there.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: Yeah, that was wrong. Anyway, the method is removed.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: This method is removed now.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: Added type hints.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: Makes sense. Removed the variable.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: The type hint was correct! I understand the code was confusing, I rewrote it without tuples, I hope now it is easier to understand.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: This is a good question, there's a bunch of methods like `whatever()` and `whatever_statevector()` now. But `_construct_circuit_with_feature_map()` is used to construct circuits in both cases: statevector and non-statevector. That's why it is named like now. I can rename it to `_construct_circuit_statevector()`, but then it would be used in `construct_circuit()` that is used in the non-statevector case and such code might be confusing. I hope I did not bring more confusion.

qiskit_machine_learning/kernels/quantum_kernel.py
dlasecki: You are right, my bad.

qiskit_machine_learning/kernels/quantum_kernel.py
gentinettagian: I think in this case it would be less confusing to just have one `_construct_circuit_with_feature_map()` method that is used for both statevector and hardware/qasm-simulator. At least when glancing over the code this synonym method confused me at first. At the places where the logic differs it is definitely nice to separate the code into separate methods.

adekusar-drl: @gentinettagian I'm not saying I'm against the change, I wonder what is exactly confusing as method name consistency is also important.

adekusar-drl:
@gentinettagian Thanks for looking into this and your comment are very welcome! I'll try to think what can be improved further, but considering a new incoming primitive based version, I don't want to spend much time on this one. Perhaps, splitting into methods should be enough :)

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -441,188 +453,208 @@ def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
...
+            raise ValueError(
+                f"""
+                The feature map circuit contains unbound training parameters ({unbound_params}).
+                All training parameters must be bound to numerical values before evaluating
+                the kernel matrix.
+                """
+            )
 
-        return kernel
+    def _raise_incompatible_feature_map(self, vec_name: str, vec_len: int) -> None:
```
woodsp-ibm, line 655: The return type here ought to be `NoReturn` I think since all this function never ever returns

woodsp-ibm:
Just a comment. I will note that overall we could do better with testing error condition paths. Here with it refactored as a method that ends up not being tested at all - not that its really any worse than before. Something to improve upon with the unit testing perhaps when we migrate these over to using primitives.

adekusar-drl:
> Just a comment. I will note that overall we could do better with testing error condition paths. Here with it refactored as a method that ends up not being tested at all - not that its really any worse than before. Something to improve upon with the unit testing perhaps when we migrate these over to using primitives.

Added some more tests. Removed one of the `if` statements that always evaluates to true.

woodsp-ibm (approved):

adekusar-drl:
@woodsp-ibm please review, last updates were just merges.

woodsp-ibm (approved):

PR #436, opened 2022-07-11
Add `SamplerQNN` class using terra primitives

### LINKED ISSUE #400: Implement a QNN based on the Sampler primitive
### What should we add?

Terra has implemented new ways to run circuits that may potentially speed up QNN training in the cloud. In this issue we implement a new type of QNN, a `SamplerQNN` that will be based on the `Sampler` primitive.

The implementation should be given an instance of the Sampler instead of a quantum circuit and should re-use the sampler for every evaluation of the circuit.

The implementation should be tested again the reference implementation in Terra and cloud versions of the primitives. The latter to be decided how to implement.

ElePT:
### Summary
This PR presents a reference implementation of a Qiskit `NeuralNetwork` that leverages the newly introduced primitives (so far, this class is based on the terra interface). Closes #400.

### Details and comments
This PR is blocked by the release of Qiskit-terra that will include gradients and changes in the primitives interface.

### Open Points of Discussion
- [x] Where to set keyword args -> all keyword?
- [x] How to define typehints: list vs sequence vs numpy array...


woodsp-ibm:
Why would this not be in `neural networks` (same comment for the kernel PR but in kernels of course) and subclass NeuralNetwork. I would think we would place this for what it is not what its built from. I realize its all draft and work in progress, so I am not going to comment beyond this aspect for now. Maybe because its all work in progress it just did not get that far yet.

ElePT:
> Why would this not be in `neural networks` (same comment for the kernel PR but in kernels of course) and subclass NeuralNetwork. I would think we would place this for what it is not what its built from. I realize its all draft and work in progress, so I am not going to comment beyond this aspect for now. Maybe because its all work in progress it just did not get that far yet.

Yes, I totally agree, I just moved my `primitives` PoC directory as-is from a different repo, but the class definitely fits better into `neural_networks`.

qiskit_machine_learning/neural_networks/__init__.py
```diff
@@ -57,6 +57,15 @@
    EffectiveDimension
    LocalEffectiveDimension
 
+Neural Networks using Primitives
```
woodsp-ibm, line 60: Do we want to separate them out like this? The existing networks will be marked Pending Deprecation and we normally change the first sentence that comes out in the resultant html table here to prefix it with Pending Deprecation so you would see at a glance here some are pending deprecation.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+class SamplerQNN(NeuralNetwork):
+    """A Neural Network implementation based on the Sampler primitive."""
+
+    def __init__(
+        self,
+        sampler: BaseSampler,
+        circuit: QuantumCircuit,
+        input_params: List[Parameter] | None = None,
```
woodsp-ibm, line 39: Do we need List  (list) or could we go with Sequence as bit more generic.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,296 @@
...
+        output_shape: int | Tuple[int, ...] | None = None,
+        gradient: BaseSamplerGradient | None = None,
+        input_gradients: bool = False
+    )-> None:
+        """
+
+        Args:
+            circuit: The parametrized quantum circuit that generates the samples of this network.
+            input_params: The parameters of the circuit corresponding to the input.
+            weight_params: The parameters of the circuit corresponding to the trainable weights.
```
woodsp-ibm, line 51: I think the descriptions here could include what happens for None (default) case.

test/neural_networks/test_sampler_qnn.py
```diff
@@ -0,0 +1,182 @@
...
+        self.qc = RealAmplitudes(num_qubits, entanglement="linear", reps=1)
+        self.qi_qasm = QuantumInstance(Aer.get_backend("aer_simulator"))
+        self.sampler = Sampler()
+
+    def test_forward_pass(self):
+
+        parity = lambda x: "{:b}".format(x).count("1") % 2
+        output_shape = 2  # this is required in case of a callable with dense output
+
+        circuit_qnn = CircuitQNN(
```
woodsp-ibm, line 51: Will this be reduced to just the Sampler QNN testing - it looks like you were comparing new/old here but in the end we already have the separate test for the CircuitQNN so this will become independent of that right - it needs to be as CircuitQNN will be removed in the future.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,296 @@
...
+            output_shape: The output shape of the custom interpretation
+            sampler_factory: Factory for sampler primitive
+            gradient_factory: String indicating pre-implemented gradient method or factory for
+                gradient class
+            input_gradients: to be added
+        Raises:
+            QiskitMachineLearningError: Invalid parameter values.
+        """
+        # set sampler --> make property?
+        self.sampler = sampler
```
woodsp-ibm, line 65: Public instance vars are what we have been moving towards - they do need an Attributes: docstring part in the class description though to make them visible in the API ref,

woodsp-ibm (commented):
I added a few comments as I skimmed over this noting it's still Draft

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -490,7 +490,6 @@ def _probability_gradients(
             num_grad_vars = self._num_inputs + self._num_weights
         else:
             num_grad_vars = self._num_weights
-
         # construct gradients
```
ElePT, line 493:
```suggestion

        # construct gradients
```

qiskit_machine_learning/utils/utils.py
```diff
@@ -0,0 +1,20 @@
...
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+import numpy as np
+
+
+def make_2d(array: np.ndarray, n_copies: int):
```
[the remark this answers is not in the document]
ElePT: I forgot to remove this method from this PR (you're right, not very useful).

qiskit_machine_learning/neural_networks/__init__.py
ElePT: Removed. I think that we will leave the deprecation warnings for a follow-up PR

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,296 @@
...
+class SamplerQNN(NeuralNetwork):
+    """A Neural Network implementation based on the Sampler primitive."""
+
+    def __init__(
+        self,
+        sampler: BaseSampler,
+        circuit: QuantumCircuit,
+        input_params: List[Parameter] | None = None,
+        weight_params: List[Parameter] | None = None,
+        interpret: Callable[[int], int | Tuple[int, ...]] | None = None,
```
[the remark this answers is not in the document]
ElePT: yep

test/neural_networks/test_sampler_qnn.py
ElePT: I kept the test for the time being but added that it should be removed once `CircuitQNN` is fully deprecated

qiskit_machine_learning/neural_networks/sampler_qnn.py
ElePT: I wonder if I should migrate all the existing properties to public instances then, I guess this would also affect the `NeuralNetwork` base class, because some properties like `input_gradients` are inherited from it.

qiskit_machine_learning/neural_networks/sampler_qnn.py
ElePT: I have changed to Sequence, but I am wondering about the discussion about the `Sequence` type not including `np.ndarray` from here: https://github.com/Qiskit/qiskit-terra/pull/8812#discussion_r985427636

pedrorrivero: Even though it says `Sequence` the primitives will keep supporting `np.ndarray`. Some people suggest moving back to `Iterable`, however these can actually be infinite (e.g. `itertools.count`), which raises concerns of secutiry, memory usage, and potential DOS attacks among others.

An alternative that I have proposed in the past is to move to `Collection` which does include `np.ndarray` but without being infinite. I personally like this approach, even though `Mapping`s are included so a dictionary could be passed.

For now, we are sticking to `Sequence` in the primitives since it is the least general. That way, generalizing it later on in the future will not be a breaking a change.

qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -117,7 +117,7 @@ def forward(  # type: ignore
             return result_tensor
 
         @staticmethod
-        def backward(ctx: Any, grad_output: Tensor) -> Tuple:  # type: ignore
+        def backward(ctx: Any, grad_output: Tensor) -> Tuple:
```
ElePT, line 120:
```suggestion
        def backward(ctx: Any, grad_output: Tensor) -> Tuple:  # type: ignore
```

qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -55,7 +55,7 @@ class TorchConnector(Module):
     class _TorchNNFunction(Function):
         # pylint: disable=arguments-differ
         @staticmethod
-        def forward(  # type: ignore
+        def forward(
```
ElePT, line 58:
```suggestion
        def forward(  # type: ignore
```

test/connectors/test_torch_connector.py
```diff
@@ -107,7 +107,7 @@ def _validate_backward_pass(self, model: TorchConnector) -> None:
             model.neural_network,
             False,
         )
-        test = torch.autograd.gradcheck(func, input_data, eps=1e-4, atol=1e-3)  # type: ignore
+        test = torch.autograd.gradcheck(func, input_data, eps=1e-4, atol=1e-3)
```
ElePT, line 110:
```suggestion
        test = torch.autograd.gradcheck(func, input_data, eps=1e-4, atol=1e-3)  # type: ignore
```

.github/workflows/main.yml
```diff
@@ -22,9 +22,8 @@ on:
       - main
       - 'stable/**'
   schedule:
-    # run every day at 1AM
-    - cron: '0 1 * * *'
-
+     # run every day at 1AM
+     - cron: '0 1 * * *'
```
ElePT, line 26:
```suggestion
     # run every day at 1AM
     - cron: '0 1 * * *'
     
```

.github/workflows/main.yml
```diff
@@ -23,7 +23,7 @@ on:
       - 'stable/**'
   schedule:
     # run every day at 1AM
-    - cron: '0 1 * * *'
+     - cron: '0 1 * * *'
```
ElePT, line 26:
```suggestion
    - cron: '0 1 * * *'
```

.github/workflows/main.yml
```diff
@@ -22,8 +22,8 @@ on:
       - main
       - 'stable/**'
   schedule:
-    # run every day at 1AM
-    - cron: '0 1 * * *'
+     # run every day at 1AM
+     - cron: '0 1 * * *'
```
ElePT, line 26:
```suggestion
    # run every day at 1AM
     - cron: '0 1 * * *'
```

.github/workflows/main.yml
ElePT, line 26:
```suggestion
    - cron: '0 1 * * *'
```

test/neural_networks/test_sampler_qnn.py
```diff
@@ -0,0 +1,279 @@
...
+
+        parity = lambda x: f"{x:b}".count("1") % 2
+        output_shape = 2  # this is required in case of a callable with dense output
+
+        qi_qasm = QuantumInstance(
+            aer.AerSimulator(),
+            shots=100,
+            seed_simulator=algorithm_globals.random_seed,
+            seed_transpiler=algorithm_globals.random_seed,
+        )
```
adekusar-drl, line 246: Statevector simulator should be used, in this case you should get exact match and the tolerance (now 0.1) could be reduced.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -47,6 +48,8 @@ class VQC(NeuralNetworkClassifier):
 
     def __init__(
         self,
+        *,
+        sampler: BaseSampler = None,
```
adekusar-drl, line 52: I don't think we can easily convert VQC to keyword only parameters. It should be a transition period when positional are accepted and generate warnings, while keywords are accepted without warnings.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -100,16 +103,30 @@ def __init__(
...
-            interpret=self._get_interpret(2),
-            output_shape=2,
-            quantum_instance=quantum_instance,
-            input_gradients=False,
-        )
+        # needed for mypy
+        neural_network: SamplerQNN | CircuitQNN = None
+        if sampler is None:
+            # construct circuit QNN
+            neural_network = CircuitQNN(
```
adekusar-drl, line 110: I'd construct `SamplerQNN` if sampler is none with the defaults. And only if `quantum_instance` is not none fallback to `CircuitQNN`. Basically, it should be preferable to create `SamplerQNN` unless we can't.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -154,9 +171,13 @@ def _fit_internal(self, X: np.ndarray, y: np.ndarray) -> OptimizerResult:
         """
         X, y = self._validate_input(X, y)
         num_classes = self._num_classes
-        cast(CircuitQNN, self._neural_network).set_interpret(
-            self._get_interpret(num_classes), num_classes
-        )
+
+        if isinstance(self._neural_network, SamplerQNN):
+            self._neural_network.set_interpret(self._get_interpret(num_classes), num_classes)
+        else:
```
adekusar-drl, line 177: Perhaps, instead of `else` should be `elif`.

qiskit_machine_learning/algorithms/classifiers/vqc.py
woodsp-ibm: As such `sampler` needs to go last in the list here, as we needed to do when we have altered other algorithms in place elsewhere

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,343 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2020, 2022.
```
woodsp-ibm, line 3:
```suggestion
# (C) Copyright IBM 2022.
```

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,343 @@
...
+logger = logging.getLogger(__name__)
+
+
+class SamplerQNN(NeuralNetwork):
+    """A Neural Network implementation based on the Sampler primitive."""
+
+    def __init__(
+        self,
+        *,
+        sampler: BaseSampler,
```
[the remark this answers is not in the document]
woodsp-ibm: FYI In algorithms in Terra we are not defaulting the primitives as required - so a user needs to provided it whether it be the reference one or not.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,343 @@
...
+            output_shape: The output shape of the custom interpretation
+            gradient: An optional sampler gradient to be used for the backward pass.
+            input_gradients: Determines whether to compute gradients with respect to input data.
+                 Note that this parameter is ``False`` by default, and must be explicitly set to
+                 ``True`` for a proper gradient computation when using ``TorchConnector``.
+        Raises:
+            QiskitMachineLearningError: Invalid parameter values.
+        """
+        # set primitive
+        self.sampler = sampler
```
woodsp-ibm, line 86: For public instance var that we are allowing a user to directly read/update these should be documented in an Attributes: section in the class docstring - here's an example if its needed https://github.com/Qiskit/qiskit-terra/blob/35c4b00c7e9bc142137ee8671be945b910198f96/qiskit/algorithms/time_evolvers/pvqd/pvqd.py#L47-L54

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,343 @@
...
+
+        # set gradient
+        # TODO: provide default gradient?
+        self.gradient = gradient or ParamShiftSamplerGradient(self.sampler)
+
+        self._circuit = circuit.copy()
+        if len(self._circuit.clbits) == 0:
+            self._circuit.measure_all()
+
+        self._input_params = list(input_params or [])
```
woodsp-ibm, line 96: `(input_params or []` when dealing with None i.e. for input_params its better to check for None explicitly.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,343 @@
...
+            self._input_gradients,
+        )
+
+    @property
+    def circuit(self) -> QuantumCircuit:
+        """Returns the underlying quantum circuit."""
+        return self._circuit
+
+    @property
+    def input_params(self) -> Sequence:
```
woodsp-ibm, line 122: Sequence here is lacking types - Sequence{Parameter] right? Same for weight_params below

test/neural_networks/test_sampler_qnn.py
```diff
@@ -0,0 +1,279 @@
...
+
+        self.interpret_2d = interpret_2d
+        self.output_shape_2d = (
+            2,
+            3,
+        )  # 1st dim. takes values in {0, 1} 2nd dim in {0, 1, 2}
+
+        # define sampler primitives
+        self.sampler = Sampler()
+        self.sampler_shots = Sampler(options={"shots": 100})
```
woodsp-ibm, line 87: I think adding `seed` to the options when using shots will allow its outcome to be reproducible  which we need for unit tests.

test/neural_networks/test_sampler_qnn.py
```diff
@@ -0,0 +1,279 @@
...
+            weight_params=self.weight_params,
+            sparse=sparse,
+            interpret=interpret,
+            output_shape=output_shape,
+        )
+        return qnn
+
+    def _verify_qnn(
+        self,
+        qnn: CircuitQNN,
```
woodsp-ibm, line 124: I don;t think this should be checking using an entity which will go away. Ideally it should check against some known reference value(s) we expect the outcome to be, Also if doing multiple tests are these dependent on one another in sequence where the first one failing we expect the rest to fail. Or is there independence whereby these can be done using sub-test which is so can give a clearer picture of what might be happening if something fails since more tests may be run.

qiskit_machine_learning/neural_networks/sampler_qnn.py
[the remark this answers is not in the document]
ElePT: Well, given that we are making it keyword, I agree with adding a default here, even if it slightly differs from the algorithms.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,343 @@
...
+        # set primitive
+        self.sampler = sampler
+
+        # set gradient
+        # TODO: provide default gradient?
+        self.gradient = gradient or ParamShiftSamplerGradient(self.sampler)
+
+        self._circuit = circuit.copy()
+        if len(self._circuit.clbits) == 0:
+            self._circuit.measure_all()
```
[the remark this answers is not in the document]
ElePT: As far as I know, yes, at least the last version I checked failed if the circuit didn't have measurements.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,343 @@
...
+        """
+        Pre-processing during forward pass of the network.
+        """
+
+        np_input_data: np.ndarray = np.asarray(input_data)
+        np_weights: np.ndarray = np.asarray(weights)
+        if len(np_input_data.shape) == 1:
+            np_input_data = np.expand_dims(np_input_data, 0)
+
+        num_samples = max(np_input_data.shape[0], 1)
```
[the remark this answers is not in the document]
ElePT: this max is there to deal with the case where `input_data=[]`, but `weights != []`, because `num_samples` is used later to broadcast the weights in the sampler call. I could also remove the max here and check for `num_samples==0` in the forward pass, but I thought this would be more compact.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,343 @@
...
+        self,
+        input_data: list[float] | np.ndarray | float,
+        weights: list[float] | np.ndarray | float,
+    ) -> tuple[np.ndarray, int]:
+        """
+        Pre-processing during forward pass of the network.
+        """
+
+        np_input_data: np.ndarray = np.asarray(input_data)
+        np_weights: np.ndarray = np.asarray(weights)
```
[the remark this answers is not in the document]
ElePT: Well, the `forward` signature accepts `list` and `float` too, so doing `np.asarray()` was necessary in some cases. The reason why the variable name changes is mypy...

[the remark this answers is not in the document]
ElePT: ok, I see, this code is from a time when this class didn't inherit from `NeuralNetwork`. Will update.

test/neural_networks/test_sampler_qnn.py
```diff
@@ -0,0 +1,279 @@
...
+            output_shape=output_shape,
+            gradient=ParamShiftSamplerGradient(self.sampler),
+            input_gradients=True,
+        )
+
+        inputs = np.asarray(algorithm_globals.random.random(size=(1, circuit_qnn._num_inputs)))
+        weights = algorithm_globals.random.random(circuit_qnn._num_weights)
+
+        circuit_qnn_fwd = circuit_qnn.backward(inputs, weights)
+        sampler_qnn_fwd = sampler_qnn.backward(inputs, weights)
```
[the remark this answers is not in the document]
ElePT: Yes, I somehow deleted those. They are back.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -154,9 +171,13 @@ def _fit_internal(self, X: np.ndarray, y: np.ndarray) -> OptimizerResult:
...
+            cast(CircuitQNN, self._neural_network).set_interpret(
```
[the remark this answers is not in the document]
ElePT: Yes, I get:
`error: "NeuralNetwork" has no attribute "set_interpret"`

qiskit_machine_learning/algorithms/classifiers/vqc.py
ElePT: sure

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,343 @@
...
+        self._weight_params = list(weight_params or [])
+
+        if sparse:
+            _optionals.HAS_SPARSE.require_now("DOK")
+
+        self.set_interpret(interpret, output_shape)
+
+        self._input_gradients = input_gradients
+
+        # TODO: look into custom transpilation
```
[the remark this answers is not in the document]
ElePT: True, the `CircuitQNN` code allowed to set a flag for custom transpilation, but this is now the primitives' responsibility. I just left that TODO back in the day to make sure to double check this was the case.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -48,13 +48,26 @@ class SparseArray:  # type: ignore
...
+        gradient (BaseSamplerGradient): An optional sampler gradient to be used for the backward pass.
+    """
 
     def __init__(
         self,
         *,
-        sampler: BaseSampler,
-        circuit: QuantumCircuit,
+        sampler: BaseSampler | None = None,
+        circuit: QuantumCircuit | None = None,
```
woodsp-ibm, line 70: Docstrings ought to state  what happens with None ie. like for sampler a reference Sampler() instance is used.

From what I can tell having None for a circuit though will fail below so to me None is invalid.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -82,35 +95,40 @@ def __init__(
...
+        # set primitive, provide default
+        if sampler is None:
+            sampler = Sampler()
         self.sampler = sampler
 
         # set gradient
-        # TODO: provide default gradient?
-        self.gradient = gradient or ParamShiftSamplerGradient(self.sampler)
+        if gradient is None:
+            gradient = ParamShiftSamplerGradient(self.sampler)
```
woodsp-ibm, line 105: The default ought to be to noted in the arg text above where its states it as optional sampler gradient...

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -186,28 +204,39 @@ def _compute_output_shape(
...
+                num_samples = 1
+                parameters = np.broadcast_to(weights, (num_samples, len(weights)))
+            else:
+                return None, None
 
         return parameters, num_samples
 
-    def _postprocess(self, num_samples: int, result: SamplerResult) -> np.ndarray:
+    def _postprocess(
+        self, num_samples: int | None, result: SamplerResult | None
```
woodsp-ibm, line 231: Won't `num_samples` being None fail below e.g in np.zeros and range.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -317,26 +331,53 @@ def _postprocess_gradient(
...
+        if num_samples is not None and np.prod(parameter_values.shape) > 0:
+            # sampler allows batching
+            job = self.sampler.run([self._circuit] * num_samples, parameter_values)
+            results = job.result()
+        else:
+            results = None
+
+        result = self._postprocess(num_samples, results)
+
+        return result
```
woodsp-ibm, line 353: I guess I find the above "awkward" somehow. To me it would be more transparent what it was doing overall were it more like
```
            result = self._postprocess(num_samples, job.result())
        else:
            result = None

        return result
```
Explicitly passing None into postprocess as results and having it internally shortcircuit its processing and return None just seems a bit convoluted and took more to follow. Maybe there is some rationale that its easier to have it check in one place in postprocess but for me it made things harder to follow here.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -85,6 +85,7 @@ def __init__(
                 On each iteration an optimizer invokes the callback and passes current weights
                 as an array and a computed value as a float of the objective function being
                 optimized. This allows to track how well optimization / training process is going on.
+            sampler: The sampler primitive used to compute neural network's results.
```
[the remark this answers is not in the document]
ElePT: Good point, it's just a matter of choice. I will go for a more "conservative" approach and accept either `sampler` or `quantum_instance` and update the docs.

qiskit_machine_learning/neural_networks/sampler_qnn.py
ElePT: Fixed it, together with comment below.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -172,12 +173,11 @@ def _fit_internal(self, X: np.ndarray, y: np.ndarray) -> OptimizerResult:
         X, y = self._validate_input(X, y)
         num_classes = self._num_classes
 
+        # instance check required by mypy (alternative to cast)
```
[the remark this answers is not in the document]
ElePT: I have reduced it to a single instance check for simplicity.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,379 @@
...
+
+        sampler (BaseSampler): The sampler primitive used to compute the neural network's results.
+        gradient (BaseSamplerGradient): An optional sampler gradient to be used for the backward pass.
+    """
+
+    def __init__(
+        self,
+        *,
+        sampler: BaseSampler | None = None,
+        circuit: QuantumCircuit,
```
adekusar-drl, line 70: I'd swap these two. First circuit, then sampler since sampler can be none.

ElePT, line 70:
```suggestion
        circuit: QuantumCircuit,
        sampler: BaseSampler | None = None,
```

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,402 @@
...
+            output_shape: The output shape of the custom interpretation
+            gradient: An optional sampler gradient to be used for the backward pass.
+                If ``None`` is given, a default instance of
+                :class:`~qiskit.algorithms.gradients.ParamShiftSamplerGradient` will be used.
+            input_gradients: Determines whether to compute gradients with respect to input data.
+                 Note that this parameter is ``False`` by default, and must be explicitly set to
+                 ``True`` for a proper gradient computation when using ``TorchConnector``.
+        Raises:
+            QiskitMachineLearningError: Invalid parameter values.
+        """
```
woodsp-ibm, line 140: I will note that many of the types in the arg list are not clickable links here - just the purple colored ones. I am not sure whether its an aspect of the new syntax or not. If I look at CircuitQNN QuantumCircuit is link there whereas not here. I'll check with something newer from Nature, though one from 10 days ago has a similar issue in terms of some things are links others not.

releasenotes/notes/add-sampler-qnn-a093431afc1c5441.yaml
```diff
@@ -0,0 +1,54 @@
...
+features:
+  - |
+    Introduced Sampler Quantum Neural Network
+    (:class:`~qiskit_machine_learning.neural_networks.SamplerQNN`) based on (runtime) primitives.
+    This implementation leverages the sampler primitive
+    (see :class:`~qiskit.primitives.BaseSampler`) and the sampler gradients
+    (see :class:`~qiskit.algorithms.gradients.BaseSamplerGradient`) to enable runtime access and
+    more efficient computation of forward and backward passes more efficiently.
+
+    The new Sampler QNN exposes a similar interface to the Circuit QNN, with a few differences.
```
woodsp-ibm, line 11: Is there a reason not to have SamplerQNN and CircuitQNN  instead of having spaces in there. The former will match a search the other not so much. I see it was done as Oflow QNN with a space in the Estimator release note though EstimatorQNN I see just as the one word.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,402 @@
...
+    ) -> None:
+        """Change 'interpret' and corresponding 'output_shape'.
+
+        Args:
+            interpret: A callable that maps the measured integer to another unsigned integer or
+                tuple of unsigned integers. See constructor for more details.
+            output_shape: The output shape of the custom interpretation, only used in the case
+                where an interpret function is provided. See constructor
+                for more details.
+        """
```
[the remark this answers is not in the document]
adekusar-drl: Updated in the constructor and here in the property as you suggested.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,402 @@
...
+
+        pass
+
+
+logger = logging.getLogger(__name__)
+
+
+class SamplerQNN(NeuralNetwork):
+    """A Neural Network implementation based on the Sampler primitive.
```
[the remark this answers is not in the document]
adekusar-drl: Updated to make lower case in the QNNs. I don't know why caps.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,402 @@
...
+
+    The ``Sampler QNN`` is a neural network that takes in a parametrized quantum circuit
```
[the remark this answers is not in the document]
adekusar-drl: Removed the space.

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,402 @@
...
+                tuple of unsigned integers. These are used as new indices for the (potentially
+                sparse) output array. If no interpret function is
+                passed, then an identity function will be used by this neural network.
+            output_shape: The output shape of the custom interpretation
+            gradient: An optional sampler gradient to be used for the backward pass.
+                If ``None`` is given, a default instance of
+                :class:`~qiskit.algorithms.gradients.ParamShiftSamplerGradient` will be used.
+            input_gradients: Determines whether to compute gradients with respect to input data.
+                 Note that this parameter is ``False`` by default, and must be explicitly set to
+                 ``True`` for a proper gradient computation when using ``TorchConnector``.
```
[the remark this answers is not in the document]
adekusar-drl: Updated the link to `TorchConnector`

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -0,0 +1,402 @@
...
+        super().__init__(
+            num_inputs=len(self._input_params),
+            num_weights=len(self._weight_params),
+            sparse=sparse,
+            output_shape=self._output_shape,
+            input_gradients=self._input_gradients,
+        )
+
+    @property
+    def circuit(self) -> QuantumCircuit:
```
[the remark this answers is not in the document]
adekusar-drl: I'd avoid copying objects until it is required. There are many other ways how to shoot in the foot.

qiskit_machine_learning/neural_networks/sampler_qnn.py
adekusar-drl: I don't know why as well. Perhaps due to the new syntax.

releasenotes/notes/add-sampler-qnn-a093431afc1c5441.yaml
adekusar-drl: I don't see a reason. Updated with the links.
I can separately update the EstimatorQNN reno if required. Let me know.

woodsp-ibm (approved):

PR #437, opened 2022-07-11
Add `QuantumKernel` class using terra primitives

ElePT:
### Summary
This PR presents a reference implementation of a `QuantumKernel` (original code by @gentinettagian) that leverages the newly introduced primitives (so far, this class is based on the terra interface).

### Details and comments

The current design is pending on the approval of the fidelity primitive (see [this PR](https://github.com/Qiskit/qiskit-terra/pull/8303)), which is currently blocked by the release of the new primitive sessions.

There are a series of items in the to-do list:

- [x] Dealing with identical samples in the kernel matrix (see #432)
- [x] **[Update: final names settled]** Deciding on class names -> once the new names are set, we can move the code to the `kernels` directory instead of creating a new `primitives` submodule
- [x] **[Update: there will be an abstract class]** Assessing whether the `TrainableQuantumKernel` should have a mixin/additional base class (as proposed in the latest version of the code). It makes sense for type-hinting purposes (so that we can check in algorithms whether Kernel is trainable), but in terms of functionality I don't see much advantage.
- [x] Assessing whether additional batching functionality is needed


qiskit_machine_learning/primitives/kernels/base_kernel.py
```diff
@@ -0,0 +1,117 @@
...
+                                                            D is the feature dimension
+            y_vec: 1D or 2D array of datapoints, MxD, where M is the number of datapoints,
+                                                            D is the feature dimension
+
+        Returns:
+            2D matrix, NxM
+
+        Raises:
+            QiskitMachineLearningError:
+                A quantum instance or backend has not been provided
```
dlasecki, line 55:
```suggestion
                A quantum instance or backend has not been provided.
```
Please check other places.

qiskit_machine_learning/primitives/kernels/base_kernel.py
```diff
@@ -0,0 +1,117 @@
...
+        r"""
+        Construct kernel matrix for given data
+
+        If y_vec is None, self inner product is calculated.
+
+        Args:
+            x_vec: 1D or 2D array of datapoints, NxD, where N is the number of datapoints,
+                                                            D is the feature dimension
+            y_vec: 1D or 2D array of datapoints, MxD, where M is the number of datapoints,
+                                                            D is the feature dimension
```
dlasecki, line 48:
```suggestion
            x_vec: 1D or 2D array of datapoints, NxD, where N is the number of datapoints,
                D is the feature dimension
            y_vec: 1D or 2D array of datapoints, MxD, where M is the number of datapoints,
                D is the feature dimension
```

qiskit_machine_learning/primitives/kernels/base_kernel.py
```diff
@@ -0,0 +1,117 @@
...
+            y_vec = np.asarray(y_vec)
+
+        if x_vec.ndim > 2:
+            raise ValueError("x_vec must be a 1D or 2D array")
+
+        if x_vec.ndim == 1:
+            x_vec = x_vec.reshape(1, -1)
+
+        if y_vec is not None and y_vec.ndim > 2:
+            raise ValueError("y_vec must be a 1D or 2D array")
```
dlasecki, line 82: One could avoid some duplication if checks for `x_vec` were extracted as a method. The same method would be calld on `y_vec` after checking it is not `None`.

qiskit_machine_learning/primitives/kernels/pseudo_kernel.py
```diff
@@ -0,0 +1,136 @@
...
+
+from qiskit import QuantumCircuit
+from qiskit.primitives import Sampler
+from qiskit.primitives.fidelity import BaseFidelity
+
+from qiskit_machine_learning.utils import make_2d
+from .quantum_kernel import QuantumKernel
+
+SamplerFactory = Callable[[List[QuantumCircuit]], Sampler]
+FidelityFactory = Callable[[List[QuantumCircuit], SamplerFactory], BaseFidelity]
```
dlasecki, line 26: Is it possible to define these in one place, e.g. `BaseKernel`? They seem to be used in all child classes.

qiskit_machine_learning/primitives/kernels/pseudo_kernel.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class PseudoKernel(QuantumKernel):
+    """
+    Pseudo kernel
```
dlasecki, line 31: More info recommended.

qiskit_machine_learning/primitives/kernels/pseudo_kernel.py
```diff
@@ -0,0 +1,136 @@
...
+        then perform inner product classically.
+
+        Args:
+            x_vec: 1D or 2D array of datapoints, NxD, where N is the number of datapoints,
+                                                            D is the feature dimension
+            y_vec: 1D or 2D array of datapoints, MxD, where M is the number of datapoints,
+                                                            D is the feature dimension
+            x_parameters: 1D or 2D array of parameters, NxP, where N is the number of datapoints,
+                                                        P is the number of trainable parameters
+            y_parameters: 1D or 2D array of parameters, MxP
```
dlasecki, line 74: Indentation too large.

qiskit_machine_learning/primitives/kernels/pseudo_kernel.py
```diff
@@ -0,0 +1,136 @@
...
+
```
dlasecki, line 75:
```suggestion
```

qiskit_machine_learning/primitives/kernels/pseudo_kernel.py
```diff
@@ -0,0 +1,136 @@
...
+            x_vec = x_vec.reshape(1, -1)
+
+        if len(y_vec.shape) == 1:
+            y_vec = y_vec.reshape(1, -1)
+
+        if len(x_parameters.shape) == 1:
+            x_parameters = make_2d(x_parameters, x_vec.shape[0])
+
+        if len(y_parameters.shape) == 1:
+            y_parameters = make_2d(y_parameters, y_vec.shape[0])
```
dlasecki, line 105: Code duplication.

qiskit_machine_learning/primitives/kernels/pseudo_kernel.py
```diff
@@ -0,0 +1,136 @@
...
+        if x_parameters.shape[1] != self.num_parameters:
+            raise ValueError(
+                f"Number of parameters provided ({x_parameters.shape[0]}) does not"
+                f"coincide with the number provided in the feature map ({self.num_parameters})."
+            )
+
+        if y_parameters.shape[1] != self.num_parameters:
+            raise ValueError(
+                f"Number of parameters provided ({y_parameters.shape[0]}) does not coincide"
+                f"with the number provided in the feature map ({self.num_parameters})."
```
dlasecki, line 133: Code duplication.

qiskit_machine_learning/primitives/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,162 @@
...
+from .base_kernel import BaseKernel
+
+
+SamplerFactory = Callable[[List[QuantumCircuit]], Sampler]
+FidelityFactory = Callable[[List[QuantumCircuit], SamplerFactory], BaseFidelity]
+
+
+class QuantumKernel(BaseKernel):
+    """
+    Overlap Kernel
```
dlasecki, line 31: More info needed.

qiskit_machine_learning/primitives/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+
+    def __init__(
+        self,
+        sampler_factory: SamplerFactory,
+        feature_map: Optional[QuantumCircuit] = None,
+        *,
+        fidelity: Union[str, FidelityFactory] = "zero_prob",
```
dlasecki, line 39: An Enum might be better than a string.

qiskit_machine_learning/primitives/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,162 @@
...
+                right_circuit=feature_map,
+                sampler_factory=sampler_factory,
+            )
+        else:
+            self._fidelity = fidelity(
+                left_circuit=feature_map,
+                right_circuit=feature_map,
+                sampler_factory=sampler_factory,
+            )
+        self._shots = 10000
```
dlasecki, line 62: Should be passed as an argument.

qiskit_machine_learning/primitives/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,162 @@
...
+
+        else:
+            left_parameters, right_parameters = self._get_parametrization(x_vec, y_vec)
+            kernel_matrix = self._get_kernel_matrix(left_parameters, right_parameters, shape)
+
+        if is_symmetric and self._enforce_psd:
+            kernel_matrix = self._make_psd(kernel_matrix)
+        return kernel_matrix
+
+    def _get_parametrization(self, x_vec: np.ndarray, y_vec: np.ndarray) -> Tuple[np.ndarray]:
```
dlasecki, line 83:
```suggestion
    def _get_parametrization(self, x_vec: np.ndarray, y_vec: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
```

qiskit_machine_learning/primitives/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,162 @@
...
+
+        index = 0
+        for i in range(shape[0]):
+            for j in range(shape[1]):
+                kernel_matrix[i, j] = kernel_entries[index]
+                index += 1
+        return kernel_matrix
+
+    def _get_symmetric_kernel_matrix(
+        self, left_parameters: np.ndarray, right_parameters: np.ndarray, shape: Tuple
```
dlasecki, line 136: Tuple type missing.

qiskit_machine_learning/primitives/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        Given a set of parametrization, this computes the kernel matrix.
+        """
+        kernel_entries = self._fidelity.compute(left_parameters, right_parameters)
+        kernel_matrix = np.zeros(shape)
+        index = 0
+        for i in range(shape[0]):
+            for j in range(i, shape[1]):
+                kernel_matrix[i, j] = kernel_entries[index]
+                index += 1
```
dlasecki, line 147: Code duplication with the previous method. Could parametrize the starting point of the inner `for` loop - `0` vs `i`.

qiskit_machine_learning/primitives/kernels/trainable_quantum_kernel.py
```diff
@@ -0,0 +1,70 @@
...
+from qiskit_machine_learning.utils import make_2d
+from .quantum_kernel import QuantumKernel
+
+SamplerFactory = Callable[[List[QuantumCircuit]], Sampler]
+FidelityFactory = Callable[[List[QuantumCircuit], SamplerFactory], BaseFidelity]
+
+
+class TrainableQuantumKernel(QuantumKernel):
+    """
+    Trainable overlap kernel.
```
dlasecki, line 31: More info needed.

qiskit_machine_learning/utils/utils.py
```diff
@@ -0,0 +1,22 @@
...
+# that they have been altered from the originals.
+
+"""Qiskit machine learning utils"""
+
+import numpy as np
+
+
+def make_2d(array: np.ndarray, n_copies: int):
+    """
+    Takes a 1D numpy array and copies n times it along a second axis.
```
dlasecki, line 20:
```suggestion
    Takes a 1D numpy array and copies it n times along a second axis.
```

test/primitives/kernels/test_qkernel_primitives.py
```diff
@@ -0,0 +1,508 @@
...
+
+from qiskit_machine_learning.primitives.kernels import (
+    QuantumKernel,
+    PseudoKernel,
+    TrainableQuantumKernel,
+)
+
+
+@ddt
+class TestQuantumKernelClassify(QiskitMachineLearningTestCase):
```
dlasecki, line 41: Usually, a test class should correspond to a class being tested, not a behaviour being tested.

dlasecki (changes requested):

qiskit_machine_learning/primitives/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,225 @@
...
+                raise ValueError(
+                    f"{fidelity} is not a valid string for a fidelity. Currently supported: 'zero_prob'."
+                )
+        else:
+            if sampler is not None:
+                warnings.warn(
+                    "Passed both a sampler and a fidelity instance. If passing a fidelity instance"
+                    " for `fidelity`, the sampler will not be used.",
+                )
+            self._fidelity = fidelity
```
woodsp-ibm, line 95: Since it seems only a Sampler or Fidelity is allowed really despite being two parameters might it be worth considering to merge to a single parameter - Sampler | Fidelity - not sure what the name would be. Given a sampler it will always use the standard/default fidelity we create internally, if you want to use a different one then you pass a Fidelity instance containing the Sampler. This gets rid of str values too which is not great when it comes to IDE autocomplete

qiskit_machine_learning/primitives/kernels/base_kernel.py
```diff
@@ -0,0 +1,141 @@
...
+
+    The quantum kernel algorithm calculates a kernel matrix, given datapoints x and y and feature
+    map f, all of n dimension. This kernel matrix can then be used in classical machine learning
+    algorithms such as support vector classification, spectral clustering or ridge regression.
+    """
+
+    def __init__(self, feature_map: QuantumCircuit = None, enforce_psd: bool = True) -> None:
+        """
+        Args:
+            enforce_psd: Project to closest positive semidefinite matrix if x = y.
```
ElePT, line 48: I know it's not shown in the docs later, but FYI, missing feature map argument.

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
```diff
@@ -11,6 +11,7 @@
 # that they have been altered from the originals.
 
 """Quantum Kernel Trainer"""
+
 import copy
 from functools import partial
 from typing import Union, Optional, Sequence
```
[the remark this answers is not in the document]
ElePT: - line 41: `def quantum_kernel(self) -> QuantumKernel | None:`
- line 94+ :     
```
def __init__(
        self,
        quantum_kernel: QuantumKernel,
        loss: str | KernelLoss | None = None,
        optimizer: Optimizer | None = None,
        initial_point: Sequence[float] | None = None,
    ):
```
- line 141: `def loss(self, loss: str | KernelLoss | None) -> None:`
- line 164:  `def initial_point(self) -> Sequence[float] | None:`
- line 169: ` def initial_point(self, initial_point: Sequence[float] | None) -> None:`
- line 228:  `def _set_loss(self, loss: str | KernelLoss | None) -> None:`

test/primitives/kernels/test_new_vs_old_kernel.py
```diff
@@ -0,0 +1,90 @@
...
+from qiskit.circuit.library import ZFeatureMap, ZZFeatureMap
+from qiskit.primitives import Sampler
+from qiskit.utils import algorithm_globals, QuantumInstance
+
+from qiskit_machine_learning.kernels import QuantumKernel as QuantumKernelOld
+from qiskit_machine_learning.primitives.kernels import QuantumKernel
+
+
+@ddt
+class TestNewVsOldQuantumKernel(QiskitMachineLearningTestCase):
```
ElePT, line 32: We already discussed this, but I really like this idea of unit test, and I plan to formally add it to the QNNs too :)

ElePT (commented):
Hello! I have gone through the PR and it looks good to me. Thanks @adekusar-drl for taking over and bringing the PR to "final" shape. I have only left a few minor comments regarding mainly typehints, but I think it's pretty much ready to be merged :)

qiskit_machine_learning/primitives/kernels/base_kernel.py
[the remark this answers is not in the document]
adekusar-drl: Added `feature_map` to the documentation

qiskit_machine_learning/primitives/kernels/base_kernel.py
```diff
@@ -0,0 +1,141 @@
...
+            2D matrix, NxM
+
+        Raises:
+            QiskitMachineLearningError:
+                A quantum instance or backend has not been provided.
+            ValueError:
+                x_vec and/or y_vec are not one or two dimensional arrays
+                x_vec and y_vec have have incompatible dimensions
+                x_vec and/or y_vec have incompatible dimension with feature map and
+                    and feature map can not be modified to match.
```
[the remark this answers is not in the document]
adekusar-drl: Removed the `Raises` section completely.

qiskit_machine_learning/primitives/kernels/base_kernel.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        raise NotImplementedError()
+
+    def _validate_input(
+        self, x_vec: np.ndarray, y_vec: np.ndarray
```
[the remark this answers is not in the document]
adekusar-drl: Added.

qiskit_machine_learning/primitives/kernels/base_kernel.py
```diff
@@ -0,0 +1,141 @@
...
+    ) -> tuple[np.ndarray, np.ndarray]:
```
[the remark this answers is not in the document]
adekusar-drl: Added.

qiskit_machine_learning/primitives/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,258 @@
...
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+"""Overlap Quantum Kernel"""
+
+from __future__ import annotations
+
+import warnings
+from typing import List, Tuple
```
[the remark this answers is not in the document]
adekusar-drl: I'm going to keep the original `QuantumKernel` as it now. It can be updated separately.

qiskit_machine_learning/primitives/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,258 @@
...
+
+import numpy as np
+from qiskit import QuantumCircuit
+from qiskit.algorithms.state_fidelities import BaseStateFidelity, ComputeUncompute
+from qiskit.primitives import Sampler
+
+from .base_kernel import BaseKernel
+
+KernelIndices = List[Tuple[int, int]]
```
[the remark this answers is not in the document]
adekusar-drl: This does not work, as these are types, not type hints.

qiskit_machine_learning/primitives/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,258 @@
...
+        evaluate_duplicates: str = "off_diagonal",
+    ) -> None:
+        super().__init__(feature_map=feature_map, enforce_psd=enforce_psd)
+
+        # if feature_map is None:
+        #     feature_map = ZZFeatureMap(2)
+        # self._feature_map = feature_map
+        #
+        # self._num_features = self._feature_map.num_parameters
+
```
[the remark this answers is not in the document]
adekusar-drl: Yeah, forgot to remove. Done.

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
```diff
@@ -22,7 +23,11 @@
 from qiskit.algorithms.variational_algorithm import VariationalResult
 from qiskit_machine_learning.utils.loss_functions import KernelLoss, SVCLoss
 
-from qiskit_machine_learning.kernels import QuantumKernel
+from qiskit_machine_learning.kernels import QuantumKernel as QuantumKernelOld
+from qiskit_machine_learning.primitives.kernels import TrainableQuantumKernel as QuantumKernelNew
+
+
+QuantumKernel = Union[QuantumKernelOld, QuantumKernelNew]
```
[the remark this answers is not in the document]
adekusar-drl: Unfortunately, this suggestion does not work.

gentinettagian (commented):
I really like what you did with the code, it seems to be in a very good shape. Not a huge fan of names like `TrainableFidelityQuantumKernel` but I'm sure you already discussed this in detail. Other then that I just added some minor comments.

qiskit_machine_learning/primitives/kernels/quantum_kernel.py
```diff
@@ -0,0 +1,258 @@
...
+                )
+        else:
+            if sampler is not None:
+                warnings.warn(
+                    "Passed both a sampler and a fidelity instance. If passing a fidelity instance"
+                    " for 'fidelity', the sampler instance will not be used.",
+                )
+            self._fidelity = fidelity
+
+    def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
```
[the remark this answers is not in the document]
adekusar-drl: Added.

qiskit_machine_learning/primitives/kernels/trainable_quantum_kernel.py
```diff
@@ -0,0 +1,101 @@
...
+
+        self._num_features = feature_map.num_parameters - self._num_trainable_parameters
+        self._feature_parameters = [
+            parameter
+            for parameter in feature_map.parameters
+            if parameter not in training_parameters
+        ]
+        self._parameter_dict = {parameter: None for parameter in feature_map.parameters}
+
+    def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray = None) -> np.ndarray:
```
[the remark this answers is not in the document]
adekusar-drl: Added

adekusar-drl:
> I really like what you did with the code, it seems to be in a very good shape. Not a huge fan of names like TrainableFidelityQuantumKernel but I'm sure you already discussed this in detail. Other than that I just added some minor comments.

Thanks for the good start! I'm not a fan of the `TrainableFidelityQuantumKernel` name either, but this is the best one we came up. The `QuantumKernel` is reserved. So, that's why we have `FidelityQuantumKernel` and then `Trainable` :)
Otherwise, it could be something like `TrainableQK` or similar which even worse.

And thanks for the comments!

qiskit_machine_learning/kernels/trainable_fidelity_quantum_kernel.py
```diff
@@ -0,0 +1,101 @@
...
+
+    This kernel has trainable parameters :math:`\theta` that can be bound using training algorithms.
+    The kernel entries are given as
+
+    .. math::
+
+        K_{\theta}(x,y) = |\langle \phi_{\theta}(x) | \phi_{\theta}(y) \rangle|^2
+    """
+
+    def __init__(
```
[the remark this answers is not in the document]
adekusar-drl: No reasons, just missed. Thanks for spotting!

qiskit_machine_learning/kernels/base_kernel.py
```diff
@@ -0,0 +1,133 @@
...
+            raise ValueError("x_vec must be a 1D or 2D array")
+
+        if x_vec.ndim == 1:
+            x_vec = np.reshape(x_vec, (-1, len(x_vec)))
+
+        if x_vec.shape[1] != self._feature_map.num_parameters:
+            # before raising an error we try to adjust the feature map
+            # to the required number of qubit.
+            try:
+                self._feature_map.num_qubits = x_vec.shape[1]
```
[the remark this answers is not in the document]
adekusar-drl: This logic was adapted from another algorithm, perhaps VQE, where the same approach without warnings is implemented. I'd keep as is for consistency reasons.

test/kernels/test_fidelity_qkernel.py
```diff
@@ -0,0 +1,345 @@
...
+        with self.assertRaises(ValueError, msg="Unsupported value of 'evaluate_duplicates'."):
+            _ = FidelityQuantumKernel(evaluate_duplicates="wrong")
+
+        with self.assertRaises(ValueError, msg="Unsupported value of 'fidelity'."):
+            _ = FidelityQuantumKernel(fidelity="wrong")
+
+        with self.assertWarns(UserWarning, msg="Both fidelity and samples are passed"):
+            _ = FidelityQuantumKernel(sampler=self.sampler, fidelity=self.fidelity)
+
+    @idata(
```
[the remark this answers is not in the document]
adekusar-drl: In general, I can add a test when a kernel is not PSD and `enforce_psd` is `False`. Apparently, the output of `evaluate` will be wrong. Let me add a quick and simple test for this case.

qiskit_machine_learning/utils/utils.py
```diff
@@ -0,0 +1,22 @@
...
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Qiskit machine learning utils"""
+
+import numpy as np
+
+
+def make_2d(array: np.ndarray, n_copies: int):
```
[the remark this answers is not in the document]
adekusar-drl: Thanks! Removed.

test/kernels/test_fidelity_qkernel.py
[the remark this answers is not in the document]
adekusar-drl: Added a test, please take a look if it makes sense.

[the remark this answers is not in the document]
gentinettagian: Looks good to me, thank you!

qiskit_machine_learning/kernels/base_kernel.py
```diff
@@ -0,0 +1,133 @@
...
+                )
+
+        return x_vec, y_vec
+
+    # pylint: disable=invalid-name
+    def _make_psd(self, kernel_matrix: np.ndarray) -> np.ndarray:
+        r"""
+        Find the closest positive semi-definite approximation to symmetric kernel matrix.
+        The (symmetric) matrix should always be positive semi-definite by construction,
+        but this can be violated in case of noise, such as sampling noise, thus the
```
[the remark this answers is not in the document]
gentinettagian: I think removing the last part ("thus adjustment is only done if NOT using the statevector simulation.") should be enough. After all, we call this method whenever `enforce_psd == True` without checking whether the sampler is statevector based beforehand.

qiskit_machine_learning/kernels/base_kernel.py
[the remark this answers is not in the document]
gentinettagian: Seems reasonable. I just thought it would be more transparent and probably help with debugging as I think if someone uses training data that doesn't fit the feature map it's probably more likely to be a mistake. It's nice that the algorithm still runs but might lead to unexpected results..

gentinettagian:
> > I really like what you did with the code, it seems to be in a very good shape. Not a huge fan of names like TrainableFidelityQuantumKernel but I'm sure you already discussed this in detail. Other than that I just added some minor comments.
> 
> Thanks for the good start! I'm not a fan of the `TrainableFidelityQuantumKernel` name either, but this is the best one we came up with. The `QuantumKernel` is reserved. So, that's why we have `FidelityQuantumKernel` and then `Trainable...` :) Otherwise, it could be something like `TrainableQK` or similar which even worse.
> 
> And thanks for the comments!

I agree, I prefer `TrainableFidelityQuantumKernel` over `TFQK` :D

qiskit_machine_learning/kernels/base_kernel.py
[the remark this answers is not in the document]
adekusar-drl: Shame on me! I've just noticed `statevector` in the docstring. You are totally right as now we don't know where circuits are executed. Removed, thanks a lot.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -85,17 +89,20 @@ def __init__(
                 - if ``quantum_kernel`` is passed and ``precomputed`` is set to ``True``. To use
                 a precomputed kernel, ``quantum_kernel`` has to be of the ``None`` type.
             TypeError:
-                - if ``quantum_instance`` neither instance of ``QuantumKernel`` nor ``None``.
+                - if ``quantum_instance`` neither instance of
```
gentinettagian, line 92:
```suggestion
                - if ``quantum_kernel`` neither instance of
```

qiskit_machine_learning/kernels/base_kernel.py
```diff
@@ -0,0 +1,134 @@
...
+                in the dataset, then the kernel will try to adjust the feature map to reflect the
+                number of features.
+            enforce_psd: Project to closest positive semidefinite matrix if ``x = y``.
+                Default ``True``.
+        """
+        if feature_map is None:
+            feature_map = ZZFeatureMap(2)
+
+        self._feature_map = feature_map
+        self._enforce_psd = enforce_psd
```
woodsp-ibm, line 60: Would we want to expose these (feature_map and enforce_psd) as public instance vars that the user can get/set to read/update these?

qiskit_machine_learning/kernels/base_kernel.py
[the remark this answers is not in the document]
woodsp-ibm: The idea was these were more template circuits where you defined what you wanted upfront but maybe at that time do not have detail on the problem. Later when we do have actual problem it can then be updated. In VQE it adjusts the ansatz to match the operator.

qiskit_machine_learning/kernels/fidelity_quantum_kernel.py
```diff
@@ -0,0 +1,289 @@
...
+    .. math::
+
+        K(x,y) = |\langle \phi(x) | \phi(y) \rangle|^2
+    """
+
+    def __init__(
+        self,
+        sampler: Sampler | None = None,
+        feature_map: QuantumCircuit | None = None,
+        fidelity: str | BaseStateFidelity = "zero_prob",
```
woodsp-ibm, line 61: Do you see doing other string values later? If other fidelities have options then strings are not really going to cut it. I have to wonder why not just have the user pass a fidelity instance and drop the sampler. If you want to default the fidelity to ComputeUncompute with the reference Sampler that seems fine. The duality of Sampler and Fidelity here just seems to complicate things for me.

qiskit_machine_learning/kernels/fidelity_quantum_kernel.py
```diff
@@ -0,0 +1,289 @@
...
+                number of features.
+            fidelity: An instance of the
+                :class:`~qiskit.algorithms.state_fidelities.BaseStateFidelity` primitive or a string
+                value defining a fidelity algorithm. Currently, the only supported string value
+                is ``zero_prob`` which corresponds to the
+                :class:`~qiskit.algorithms.state_fidelities.ComputeUncompute` fidelity.
+            enforce_psd: Project to closest positive semidefinite matrix if ``x = y``.
+                Default ``True``.
+            evaluate_duplicates: Defines a strategy how kernel matrix elements are evaluated if
+               duplicate samples are found. Possible values are:
```
woodsp-ibm, line 85: Would it make sense for evaluate_duplicates to be part of base class like enforce psd is? I see a comment about it being missed to start with from the trainable version.

qiskit_machine_learning/kernels/trainable_kernel_mixin.py
```diff
@@ -0,0 +1,78 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Trainable Quantum Kernel"""
```
woodsp-ibm, line 13: Maybe this can be altered to note its a mixin here - otherwise its the exact same as the TrainableFidelityQKernel

releasenotes/notes/add-fidelity-quantum-kernel-d40278abb49e19b5.yaml
```diff
@@ -0,0 +1,22 @@
+---
+features:
+  - |
+    Added a new implementation of quantum kernel. This implementation leverages the fidelity
```
woodsp-ibm, line 4: Maybe we can more say something like... this introduces Quantum Kernels based on (runtime) primitives that will allow better performance, and use of noise mitigation etc provided by the primitives. Something with a bit more of a top level statement rather than the matter of fact "new implementation"

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -14,15 +14,19 @@
...
 from qiskit.utils import algorithm_globals
 from sklearn.base import ClassifierMixin
 
 from ...algorithms.serializable_model import SerializableModelMixin
 from ...exceptions import QiskitMachineLearningError
-from ...kernels.quantum_kernel import QuantumKernel
+from ...kernels import BaseKernel, FidelityQuantumKernel
+from ...kernels.quantum_kernel import QuantumKernel as QuantumKernelOld
+
+QuantumKernel = Union[QuantumKernelOld, BaseKernel]
```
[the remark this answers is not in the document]
adekusar-drl: I'd prefer to keep `QuantumKernel` separated from the new hierarchy. Especially considering it's training capabilities. While inheriting from `BaseKernel` may be not so bad, but adding the mixin will make it over complicated, the ways how kernel elements are computed are too different now. Since we want to deprecate the existing one, I don't really want to change it now.

adekusar-drl:
> The idea was these were more template circuits where you defined what you wanted upfront but maybe at that time do not have detail on the problem. Later when we do have actual problem it can then be updated. In VQE it adjusts the ansatz to match the operator.

For some reasons, I can't reply in the thread, there's no reply box. Honestly, I don't really get the comment. This adjustment was introduced some time ago and here it is just copied. In fact, I don't really like the idea that we change the feature map without asking user. I'd like to take this adjustment down at all and raise an exception.

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
```diff
@@ -22,7 +23,11 @@
 from qiskit.algorithms.variational_algorithm import VariationalResult
 from qiskit_machine_learning.utils.loss_functions import KernelLoss, SVCLoss
 
-from qiskit_machine_learning.kernels import QuantumKernel
+from qiskit_machine_learning.kernels import QuantumKernel as QuantumKernelOld
+from qiskit_machine_learning.kernels import TrainableFidelityQuantumKernel
+
+
+QuantumKernel = Union[QuantumKernelOld, TrainableFidelityQuantumKernel]
```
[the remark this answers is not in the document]
gentinettagian: I highly doubt that this would be straightforward. The whole reason for this PR (before we started adding the primitives) was that it was basically impossible to extend the QuantumKernel class in a clean way.

qiskit_machine_learning/kernels/base_kernel.py
gentinettagian: Resetting the feature map means reinitializing a fidelity instance with the corresponding circuit. If number of parameters change, this also needs to be accounted for which is especially non-trivial in the trainable case. We thought this through at some point and came to the conclusion that it might be easier to just create a new kernel instance when you need to change the feature map.
With `enforce_psd` you're right, I think that could be made public without issues.

qiskit_machine_learning/kernels/fidelity_quantum_kernel.py
gentinettagian: I think the idea was that this way allows users to use the kernel very similarly to how they used the old QuantumKernel. Instead of a quanutm instance, they now just provide a sampler and don't have to care about how fidelities are calculated. 

I agree that this complicates the code a lot and maybe should be discussed again.

qiskit_machine_learning/kernels/fidelity_quantum_kernel.py
gentinettagian: This would restrict the type of kernels to be implemented to those where  `k(x,x) = 1` is normalized, which I don't think always needs to be the case.

woodsp-ibm:
> >  The idea was these were more template circuits where you defined what you wanted upfront but maybe at that time do not have detail on the problem. Later when we do have actual problem it can then be updated. In VQE it adjusts the ansatz to match the operator.

> For some reasons, I can't reply in the thread, there's no reply box. Honestly, I don't really get the comment. This adjustment was introduced some time ago and here it is just copied. In fact, I don't really like the idea that we change the feature map without asking user. I'd like to take this adjustment down at all and raise an exception.

My comment was more to the effect that a number of the library circuits were designed so that the number of qubits could be set later. They are referred to as template circuits since you can state the template you want but do not the size (num qubits) up front. Later when the problem is known it can be updated. This allowed us to build out/configure an algorithm with a circuit without needing to know the size up front. Later when the problem was processed and size was figured it could be set later. Maybe you don;t need to, for example, set up a classifier in a way it could classify different sizes where the the size is not known when you create the classifier. But the ability to have it updated later allows this. It was more originally for VQE where we want to create an ground state apparatus but until the problem is defined/run we do not know the num qubits (operator size) hence the desire to be able to set it later.

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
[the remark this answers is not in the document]
woodsp-ibm: I was not looking to extend the old QuantumKernel. More would it be simple enough to adapt it a little so the old code looked like the new way wrt to the interface e.g. BaseKernel that I commented elsewhere or TrainableKernelMixin here. The intent remains to drop it going forwards I was more wondering if adapting it lightly would mean this code could be what we want going forwards at this point rather than having some Union and having to change this once again when the QuantumKernel gets dropped.

qiskit_machine_learning/kernels/fidelity_quantum_kernel.py
woodsp-ibm: Maybe convenience/factory method(s) could be provided. Have one that takes a Sampler and builds out the fidelity with it and returns an instance. While a user could do that, if its something they often want to do a separate factory method could handle that rather than having a more complicated constructor that incorporates that too.

adekusar-drl:
> 
> My comment was more to the effect that a number of the library circuits were designed so that the number of qubits could be set later. They are referred to as template circuits since you can state the template you want but do not need the size (num qubits) up front. Later when the problem is known it (the size) can be updated. This allowed us to build out/configure an algorithm with a circuit without needing to know the size up front. Later when the problem was processed and size was figured it could be set later. Maybe you don;t need to, for example, set up a classifier in a way it could classify different sizes where the the size is not known when you create the classifier. But the ability to have it updated later allows this. It was more originally for VQE where we want to create an ground state apparatus but until the problem is defined/run we do not know the num qubits (operator size) hence the desire to be able to set it later. It allows reuse too of the same algo+circui on different sized problems.

I recall your similar comment, perhaps in `VQC` or somewhere else in QML. I understand why it is so, but I still don't like that. This adjustment happens under the hood and user don't know about it. Another drawback of this "feature" is that we assume `num_features == num_qubits` which may not true in general. I'm in favor of lightweight classes, when it is easier to create a new one with different parameters, rather than re-using the old one. Each instance is like a new configuration. In the same time, algorithms may be re-usable classes.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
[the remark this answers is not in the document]
adekusar-drl: All right, now we have 
- `BaseKernel`
- `FidelityQuantumKernel(BaseKernel)`
- `TrainableKernel(BaseKernel, ABC)`
- `FidelityTrainableQuantumKernel(TrainableKernel, FidelityQuantumKernel)`
- `QuantumKernel(BaseKernel, TrainableKernel)`
Additional tweaks of `**kwargs` may be required in the future to support more inheritance patterns.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -64,7 +68,7 @@ def __init__(
     ) -> None:
         """
         Args:
-            quantum_kernel: QuantumKernel to be used for classification. Has to be ``None`` when
+            quantum_kernel: a quantum kernel to be used for classification. Has to be ``None`` when
```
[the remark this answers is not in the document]
adekusar-drl: This is just to emphasize this is not a class, but rather a term. The PR is large enough, I don't want to change other things. All type hints require an additional revision of the QML code.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -85,17 +89,20 @@ def __init__(
                 - if ``quantum_kernel`` is passed and ``precomputed`` is set to ``True``. To use
                 a precomputed kernel, ``quantum_kernel`` has to be of the ``None`` type.
             TypeError:
-                - if ``quantum_instance`` neither instance of ``QuantumKernel`` nor ``None``.
+                - if ``quantum_kernel`` neither instance of
+                  :class:`~qiskit_machine_learning.kernels.BaseKernel` nor ``None``.
```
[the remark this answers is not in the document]
adekusar-drl: Now should be fine :)

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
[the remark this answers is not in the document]
adekusar-drl: Yeah, no more unions.

qiskit_machine_learning/kernels/fidelity_quantum_kernel.py
adekusar-drl: I'd agree with @gentinettagian.

qiskit_machine_learning/kernels/trainable_kernel_mixin.py
adekusar-drl: Updated

qiskit_machine_learning/kernels/base_kernel.py
adekusar-drl: Updating the feature map may be non-trivial, that's right. Even for trivial properties, I'm hesitant to provide setters, immutable object safer and easier to maintain. Added some getters.

qiskit_machine_learning/kernels/fidelity_quantum_kernel.py
adekusar-drl: I guess it is fine to have just `fidelity` as a parameter. The code is simpler, but less flexible from the user perspective.

releasenotes/notes/add-fidelity-quantum-kernel-d40278abb49e19b5.yaml
adekusar-drl: Slightly updated. I don't want to write things like "better performance, and use of noise mitigation" as it's unrelated to QML.

woodsp-ibm:
> Another drawback of this "feature" is that we assume num_features == num_qubits which may not true in general. I'm in favor of lightweight classes, when it is easier to create a new one with different parameters, rather than re-using the old one. Each instance is like a new configuration. In the same time, algorithms may be re-usable classes.

I have not looked at the feature maps in some while. The idea of the BlueprintCircuit (what I keep calling template circuit) was that base classes can extend this and invalidate it etc over changes requiring it to be rebuilt. So setting feature_dimension as an indirect to num_qubits may be a better option. Either way it comes down to use cases. When we first did Aqua we did it as you are talking about where a new instance needed to be created for each use. As more application stack logic got built out this linear progression down the stack building stuff out as you go needed to be revised. We wanted higher level iteration over the algorithms, be able to build out the algorithm, ie create/config it, for later use by the stack, but often while a user might be able to select a circuit up front the number of qubits was not known. Hence the need to be able to update in that regard. So for me it comes down to how one might expect these to be used going forwards.

releasenotes/notes/add-fidelity-quantum-kernel-d40278abb49e19b5.yaml
woodsp-ibm: In someways though this is a big piece the rationale around going over to primitives. I imagine there will be some sort of migration section to cover going from the old to the new. That can cover a more general statement I imagine.

adekusar-drl: Updated the reno. I'm not sure if I addressed all you comments there.

adekusar-drl:
> 
> I have not looked at the feature maps in some while. The idea of the BlueprintCircuit (what I keep calling template circuit) was that base classes can extend this and invalidate it etc over changes requiring it to be rebuilt. So setting feature_dimension as an indirect to num_qubits may be a better option, but not sure they handle that or not. Either way it comes down to use cases. When we first did Aqua we did it as you are talking about where a new instance needed to be created for each use. As more application stack logic got built out this linear progression down the stack building stuff out as you go needed to be revised. We wanted higher level iteration over the algorithms, be able to build out the algorithm, ie create/config it, for later use by the stack, but often while a user might be able to select a circuit up front the number of qubits was not known. Hence the need to be able to update in that regard. So for me it comes down to how one might expect these to be used going forwards.

`feature_dimension` does not work with all circuits. I'm going to keep as it is now, but this part to be revised.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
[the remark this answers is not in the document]
woodsp-ibm: Maybe we can also update Union & Optional constructs too, like for the quantum_kernel arg here, to use the | that we are preferring to use now.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -85,17 +86,18 @@ def __init__(
...
             if quantum_kernel is not None:
                 raise ValueError("'quantum_kernel' has to be None to use a precomputed kernel")
         else:
             if quantum_kernel is None:
-                quantum_kernel = QuantumKernel()
-            elif not isinstance(quantum_kernel, QuantumKernel):
-                raise TypeError("'quantum_kernel' has to be of type None or QuantumKernel")
+                quantum_kernel = FidelityQuantumKernel()
+            elif not isinstance(quantum_kernel, BaseKernel):
+                raise TypeError("'quantum_kernel' has to be of type None or BaseKernel")
```
woodsp-ibm, line 100: Typically we do not police types like this especially when only one type is specified by the typehint. If you wish to keep it then it might be nice to indicate the failing type in the error message too.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -340,14 +345,15 @@ def precomputed(self) -> bool:
     @precomputed.setter
     def precomputed(self, precomputed: bool):
         """Sets the pre-computed kernel flag. If ``True`` is passed then the previous kernel is
-        cleared. If ``False`` is passed then a new instance of ``QuantumKernel`` is created."""
+        cleared. If ``False`` is passed then a new instance of
+        :class:`~qiskit_machine_learning.kernels.BaseKernel` is created."""
```
woodsp-ibm, line 349: Should this be FidelityQuantumKernel - that is a BaseKernel, but is that too abstract that a user needs to know the actual type?

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -41,7 +41,7 @@ class QSVC(SVC, SerializableModelMixin):
         qsvc.predict(sample_test)
     """
 
-    def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwargs):
+    def __init__(self, *args, quantum_kernel: Optional[BaseKernel] = None, **kwargs):
         """
         Args:
             quantum_kernel: QuantumKernel to be used for classification.
```
woodsp-ibm, line 47:
```suggestion
            quantum_kernel: Quantum kernel to be used for classification.
```

woodsp-ibm: Maybe you want to create a different issue/PR to address this but a little but further below (cannot directly comment at that point) it has

>f"Positional arguments ({args}) are deprecated as of version 0.3.0 and "

Since that was more 6 months ago and upcoming 0.5.0 will be 2 releases it seems time to do this.

qiskit_machine_learning/algorithms/regressors/qsvr.py
```diff
@@ -40,7 +40,7 @@ class QSVR(SVR, SerializableModelMixin):
         qsvr.predict(sample_test)
     """
 
-    def __init__(self, *args, quantum_kernel: Optional[QuantumKernel] = None, **kwargs):
+    def __init__(self, *args, quantum_kernel: Optional[BaseKernel] = None, **kwargs):
         """
         Args:
             quantum_kernel: QuantumKernel to be used for regression.
```
woodsp-ibm, line 46:
```suggestion
            quantum_kernel: Quantum kernel to be used for regression.
```

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
```diff
@@ -88,7 +89,7 @@ class QuantumKernelTrainer:
 
     def __init__(
         self,
-        quantum_kernel: QuantumKernel,
+        quantum_kernel: TrainableKernel,
```
woodsp-ibm, line 92: The example code above ideally should be updated to use a newer primitive based kernel

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -163,13 +166,14 @@ def training_parameters(self, training_params: ParameterVector | Sequence[Parame
         self._training_parameters = copy.deepcopy(training_params)
 
     def assign_training_parameters(
-        self, values: Mapping[Parameter, ParameterValueType] | Sequence[ParameterValueType]
+        self,
+        parameter_values: Mapping[Parameter, ParameterValueType] | Sequence[ParameterValueType],
```
woodsp-ibm, line 170: If someone called this via keyword arg then they would have an issue with this change. But this is called from the trainer right, which did it positionally anyway from the looks, so maybe any real end user risk of breakage is minimal since they would not (normally) be calling this directly themselves

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -163,13 +166,14 @@ def training_parameters(self, training_params: ParameterVector | Sequence[Parame
...
     ) -> None:
         """
         Assign training parameters in the ``QuantumKernel`` feature map.
```
woodsp-ibm, line 173:
```suggestion
        Assign training parameters in the quantum kernel's feature map.
```

qiskit_machine_learning/kernels/trainable_fidelity_quantum_kernel.py
```diff
@@ -0,0 +1,133 @@
...
+        self,
+        *,
+        feature_map: QuantumCircuit | None = None,
+        fidelity: BaseStateFidelity | None = None,
+        training_parameters: ParameterVector | list[Parameter] | None = None,
```
woodsp-ibm, line 46: Generally we have been trying to be more generic in these hints so as not to tie things down to a list specifically but rather use Sequence so `Sequence[Parameter]`

qiskit_machine_learning/kernels/trainable_kernel.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @property
+    def parameter_values(self) -> np.ndarray:
+        """
+        Returns numerical values assigned to the training parameters as a numpy array.
+        """
+        return np.asarray([self._parameter_dict[param] for param in self._training_parameters])
+
+    @property
+    def training_parameters(self) -> ParameterVector | Sequence[Parameter] | None:
```
woodsp-ibm, line 84: I am not sure the None here is needed. On input None is replaced by an empty list. Indeed if it were possible to be None then the num_training_parameters below would fail ie len(None) which is what made me check things

releasenotes/notes/add-fidelity-quantum-kernel-d40278abb49e19b5.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - A fidelity-based trainable quantum kernel
+        :class:`~qiskit_machine_learning.kernels.TrainableFidelityQuantumKernel` is introduced. This
+        is a replacement of the existing :class:`~qiskit_machine_learning.kernels.QuantumKernel` if
+        a trainable kernel is required. The trainer
+        :class:`~qiskit_machine_learning.kernels.algorithms.QuantumKernelTrainer` now accepts both
+        quantum kernel implementations, the new one and the existing one.
+
+    The existing algorithms such as :class:`~qiskit_machine_learning.algorithms.QSVC`,
+    :class:`~qiskit_machine_learning.algorithms.QSVR` and other kernel-based algorithms are updated
+    to accept both implementation.
```
woodsp-ibm, line 35:
```suggestion
    to accept both implementations.
```

test/kernels/test_fidelity_qkernel.py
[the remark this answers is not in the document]
woodsp-ibm: I think all these tests are used with the reference Sampler() where it defaults to shots=None and gives an ideal result akin to what statevector gave us before. Maybe this is all good for testing and having some tests with some specific shots, which is when we used to need the enforce_psd when using a qasm simulator is not needed?

woodsp-ibm:
Looking at the coverage its very good - BaseKernel is a fair bit lower. Error cases and properties seem to get missed often in testing where they end up not being covered.

qiskit_machine_learning/algorithms/classifiers/qsvc.py
adekusar-drl: #483 is to address this comment, created recently. This PR is getting heavier and heavier.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: This change comes from `mypy` and that comes from the new `TrainableKernel` where `parameter_values` was introduced. I can replace back to `values`. I don't like plain `values`, it's too generic, but it's still okay.

qiskit_machine_learning/algorithms/classifiers/qsvc.py
woodsp-ibm: Ah ok, you already have an issue to address that. (Since the deprecation removals sometimes get overlooked I figured I'd mention it).

qiskit_machine_learning/kernels/quantum_kernel.py
woodsp-ibm: When I looked I could see this being called from the trainer. If we don;t really expect users to call this directly it seems like low risk as I commented. Since its in a method called assign_training_parameters there is already context to the name. If you prefer the new name though as better going forwards we can stay with it.

test/kernels/test_fidelity_qkernel.py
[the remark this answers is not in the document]
adekusar-drl: The code you commented was being updated at that moment and two sub tests were removed as they became irrelevant when `fidelity` and `sampler` properties were merged. There's `test_enforce_psd` to test psd enforcement.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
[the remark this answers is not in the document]
adekusar-drl: I wanted to do that in a separate PR as this one is heavy now, but okay. Replaced with`|`.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: Removed this `elif` clause.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
adekusar-drl: That's right, updated.

qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py
adekusar-drl: The code snippet is updated.

qiskit_machine_learning/kernels/quantum_kernel.py
```diff
@@ -179,6 +183,7 @@ def assign_training_parameters(
             ValueError: Incompatible number of training parameters and values
 
         """
+        values = parameter_values
```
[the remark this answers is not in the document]
adekusar-drl: Replaced with `the quantum kernel`

qiskit_machine_learning/kernels/trainable_fidelity_quantum_kernel.py
adekusar-drl: Thanks for spotting, I'm trying to replace with `Sequence` wherever possible.

qiskit_machine_learning/kernels/trainable_kernel.py
adekusar-drl: `None` is removed.

qiskit_machine_learning/kernels/quantum_kernel.py
adekusar-drl: Then I keep the code.

adekusar-drl:
> Looking at the coverage its very good - BaseKernel is a fair bit lower. Error cases and properties seem to get missed often in testing where they end up not being covered.

Added more tests for the base class and for fidelity kernels.

woodsp-ibm (approved):
LGTM!

PR #443, opened 2022-07-21
Delay gradient operator initialization

adekusar-drl:
### Summary
This PR introduces a delayed construction of the gradient operator in `OpflowQNN`. The problem is that if an operator is complex, has many parameters, then a conversion from the operator to the gradient operator may take quite a long time, dozens of seconds and minutes. When an operator is complex it is easier to use gradient free optimizers and the gradient operator is never used. Now, it is created when `backward` is called for the first time.

### Details and comments
As an example a gradient operator with 42 parameters on 4 qubits takes 141 seconds to be constructed under profiler. Certainly, things to be improved in Terra as well.

 
image


woodsp-ibm:
As this can be seen as a performance improvement for end users, where gradients are not used, maybe we should have reno to note the performance improvement and when it can be expected.

adekusar-drl:
> As this can be seen as a performance improvement for end users, where gradients are not used, maybe we should have reno to note the performance improvement and when it can be expected.

I've added a reno, similar code to `CircuitQNN` and additional tests.

qiskit_machine_learning/neural_networks/circuit_qnn.py
```diff
@@ -141,10 +144,10 @@ def __init__(
         # use given gradient or default
         self._gradient = gradient or Gradient()
 
-        # prepare probability gradient opflow object
-        self._construct_gradient_circuit()
-
     def _construct_gradient_circuit(self):
+        if self._gradient_circuit_constructed:
```
woodsp-ibm, line 148: Normally we define all the instance variables in the init. Its indirectly done via the input_gradients setter but in my mind it would be better to precede that by defining this there, same for gradient_circuit.

adekusar-drl: > Normally we define all the instance variables in the init. Its indirectly done via the input_gradients setter but in my mind it would be better to precede that by defining this there, same for gradient_circuit.

Do you mean that these lines should be in the `init`
```python
        # reset gradient circuit
        self._gradient_circuit = None
        self._gradient_circuit_constructed = False
```
?

adekusar-drl: Idea was to have these lines in the setter to avoid keeping identical lines in both `init` and setter.

adekusar-drl: @woodsp-ibm can you please take a look at the question above?

woodsp-ibm: Its best practice to define/create these instance vars in init. IDEs like PyCharm will show a warning if they are not. Clearly you can do this Python but defining them upfront in init is best practice so for example it lets one easily see all the instance vars of the class for readability/maintenance etc, the vars that exist are not based on whether something is executed or not etc.

woodsp-ibm: Ok, long offline discussion. It seems that since these are defined in a property setter is the reason lint does not flag them. With the setter (property) being initialized in the constructor that seems to be good enough and presumably prevents defining the instance var there and calling the setter which kinda duplicates things and is perhaps a bit ungainly.

woodsp-ibm (approved):

PR #444, opened 2022-07-22
QuantumKernel unit test parameter bind fix

gentinettagian:
### Summary

The unit tests for the `QuantumKernel` and `QuantumKernelTrainer` were changing the names of feature map parameters by manually accessing and changing the private variable `parameter._name`. Although this changes the name of the variable, it does not change the actual instruction in the circuit. In this case, the unit tests still work (which probably means we should add more unit tests), but generally this can lead to unexpected behaviour.

### Details and comments
The code is fixed by changing parameters the intended way, i.e. setting the `parameter_prefix` when initialising the feature map.


Cryoris (approved):
I'm surprised this even worked so far 😂 This update looks good to me!

adekusar-drl (approved):
Thanks a lot for spotting this issue! 
I think the tests were written before `parameter_prefix` was introduced in `ZZFeatureMap` (it was added recently) and this is why the names were set manually.

PR #452, opened 2022-08-11
Enable inter-documentation links to be clickable

manoelmarques:
### Summary
Similar to Qiskit/qiskit-nature#783

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -210,7 +210,11 @@ html_theme_options = {
 }
 
 intersphinx_mapping = {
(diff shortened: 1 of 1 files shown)


adekusar-drl:
@manoelmarques how is it going to work?

woodsp-ibm (approved):

manoelmarques:
> @manoelmarques how is it going to work?

intersphinx connects external documentation so if if there is some reference to Terra, numpy, python etc you will be able to click on it and it will take you to that documentation. In the current docs you can't click on them

PR #470, opened 2022-09-14
Add `EstimatorQNN` class using terra primitives

### LINKED ISSUE #401: Implement a QNN based on the Estimator primitive
### What should we add?

Terra has implemented new ways to run circuits that may potentially speed up QNN training in the cloud. In this issue we implement a new type of QNN, an `EstimatorQNN` that will be based on the `Estimator` primitive.

The implementation should be given an instance of the Sampler instead of a quantum circuit and should re-use the sampler for every evaluation of the circuit.

The implementation should be tested again the reference implementation in Terra and cloud versions of the primitives. The latter to be decided how to implement.

a-matsuo:
### Summary
This PR introduce a new NeuralNetwork class, `EsitmatorQNN`, that uses the `Estimator` primitive inside.

- [x] unittests 

close #401


adekusar-drl (commented):
Thanks, I have run a first pass of the code review.
Can you please:
- Update `VQR` to make use of `EstimatorQNN`.
- Update the tests for `VQR`.
- If you find that `VQR` does not really need `TwoLayerQNN`, then maybe it is easier to deprecate `TwoLayerQNN` instead of updating it.
- Depending on the previous outcome: add tests for `TwoLayerQNN` when both branches (old and new) of the code are tested.

qiskit_machine_learning/neural_networks/two_layer_qnn.py
```diff
@@ -81,22 +105,83 @@ def __init__(
         self._circuit.append(self._ansatz, range(num_qubits))
 
         # construct observable
-        self.observable = (
+        self._observable = (
             observable if observable is not None else PauliSumOp.from_list([("Z" * num_qubits, 1)])
         )
```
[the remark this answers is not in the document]
woodsp-ibm: Do you think we need such a class going forwards? Could instead we just stay with EstimatorQNN and have some convenience/factory function that given a pair of circuits/num qubits etc just builds out an EstimatorQNN and returns it? As far as can tell it just combines the pair into a single circuit and takes the input and weights from each of the circuits. Do we need a class for this? With some factory method the parameters could be ordered more comparably to EstimatorQNN. If a class was really wanted could we come up with some other name so it could be built out to avoid "combining" it with the old

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -0,0 +1,198 @@
...
+            input_gradients: Determines whether to compute gradients with respect to input data.
+                Note that this parameter is ``False`` by default, and must be explicitly set to
+                ``True`` for a proper gradient computation when using ``TorchConnector``.
+        """
+        self._estimator = estimator
+        self._circuit = circuit
+        self._observables = observables
+        self._input_params = list(input_params) or []
+        self._weight_params = list(weight_params) or []
+        self._gradient = gradient
```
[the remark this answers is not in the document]
a-matsuo: I followed the structure of existing `OpflowQNN` and other QNNs, so we can only read `observables`. It seems QNNs are basically immutable. What do you think? @adekusar-drl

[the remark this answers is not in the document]
adekusar-drl: Personally, I'm in favor of immutability. I'm happy with getters, in general, classes should be designed in a way that it is easier to create them with another configuration rather than updating them. Especially if setters are heavy.

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+class EstimatorQNN(NeuralNetwork):
+    """A Neural Network implementation based on the Estimator primitive."""
+
+    def __init__(
+        self,
+        estimator: BaseEstimator,
+        circuit: QuantumCircuit,
+        observables: Sequence[Union[BaseOperator, PauliSumOp]],
```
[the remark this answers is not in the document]
a-matsuo: Ah, this is important thing. Where do you want to put `estimator`? At the end of the arguments like this?
```
def __init__(
    self,
    circuit: QuantumCircuit,
    observables: Sequence[BaseOperator | PauliSumOp] | None = None,
    input_params: Sequence[Parameter] | None = None,
    weight_params: Sequence[Parameter] | None = None,
    gradient: BaseEstimatorGradient | None = None,
    input_gradients: bool = False,
    estimator: BaseEstimator | None = None,
):
```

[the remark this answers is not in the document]
a-matsuo: Oh, I can use `*` like `SamplerQNN` to allow only keyword arguments.

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -131,6 +131,9 @@ def _preprocess(
 
     def _forward_postprocess(self, num_samples: int, result: EstimatorResult) -> np.ndarray:
         """Post-processing during forward pass of the network."""
+        print('====',result.values)
```
woodsp-ibm, line 134: We will need to remove the print right - maybe unintentionally checked in like this

qiskit_machine_learning/algorithms/regressors/vqr.py
```diff
@@ -15,75 +15,127 @@
...
                 returns ``None``. The callback can access intermediate data during training.
                 On each iteration an optimizer invokes the callback and passes current weights
                 as an array and a computed value as a float of the objective function being
                 optimized. This allows to track how well optimization / training process is going on.
+            estimator: If an estimator instance is set, the underlying QNN will be of type
+                :class:`~qiskit_machine_learning.neural_networks.EstimatorQNN`, and the estimator
+                primitive will be used to compute the neural network's results.
         Raises:
             QiskitMachineLearningError: Needs at least one out of ``num_qubits``, ``feature_map`` or
                 ``ansatz`` to be given. Or the number of qubits in the feature map and/or ansatz
```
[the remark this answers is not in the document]
adekusar-drl: Certainly, it will be over complicated!

qiskit_machine_learning/algorithms/regressors/vqr.py
```diff
@@ -15,75 +15,127 @@
...
+                instance is also set, it will override the `quantum_instance` parameter and
+                a :class:`~qiskit_machine_learning.neural_networks.EstimatorQNN`
+                will be used instead.
             initial_point: Initial point for the optimizer to start from.
             callback: a reference to a user's callback function that has two parameters and
                 returns ``None``. The callback can access intermediate data during training.
                 On each iteration an optimizer invokes the callback and passes current weights
                 as an array and a computed value as a float of the objective function being
                 optimized. This allows to track how well optimization / training process is going on.
+            estimator: If an estimator instance is set, the underlying QNN will be of type
```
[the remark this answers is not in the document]
adekusar-drl: Updated the docstring

qiskit_machine_learning/algorithms/regressors/vqr.py
```diff
@@ -15,75 +15,127 @@
...
+            self._feature_map = feature_map
+            self._ansatz = ansatz
+            self._num_qubits = num_qubits
+            circuit = QuantumCircuit(self._num_qubits)
+            circuit.compose(self._feature_map, inplace=True)
+            circuit.compose(self._ansatz, inplace=True)
+
+            # construct observable
+            if observable is None:
+                observable = SparsePauliOp.from_list([("Z" * num_qubits, 1)])
```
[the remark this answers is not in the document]
adekusar-drl: Removed and now VQR relies on the underlying EstimatorQNN

releasenotes/notes/add-estimator-qnn-270b31662988bef9.yaml
```diff
@@ -0,0 +1,56 @@
+---
+features:
+  - |
+    Introduced Estimator Quantum Neural Network
+    (class, `~qiskit_machine_learning.neural_networks.EstimatorQNN`) based on (runtime) primitives.
```
woodsp-ibm, line 5:
```suggestion
    (:class:`~qiskit_machine_learning.neural_networks.EstimatorQNN`) based on (runtime) primitives.
```

releasenotes/notes/add-estimator-qnn-270b31662988bef9.yaml
```diff
@@ -0,0 +1,56 @@
...
+    instead, this parameter expects an (optionally custom) primitive gradient.
+
+    The existing training algorithms such as :class:`~qiskit_machine_learning.algorithms.VQR`,
+    that were based on the Opflow QNN, are updated to accept both implementations. The implementation
+    of :class:`~qiskit_machine_learning.algorithms.NeuralNetworkRegressor` has not changed.
+
+    For example a `VQR` using `EstimatorQNN` can be trained as follows:
+
+
+    Example:
```
woodsp-ibm, line 25: I think these extra blank lines an Example:: can be deleted. With the prior sentence saying "for example ... as follows:" the outcome is not so great with it as you can see from the capture of the docs release note

image

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -0,0 +1,245 @@
...
+        input_params: Sequence[Parameter] | None = None,
+        weight_params: Sequence[Parameter] | None = None,
+        gradient: BaseEstimatorGradient | None = None,
+        input_gradients: bool = False,
+    ):
+        r"""
+        Args:
+            estimator: The estimator used to compute neural network's results.
+                If ``None``, a default instance of the reference estimator,
+                :class:`~qiskit.algorithms.Estimator`, will be used.
```
woodsp-ibm, line 70:
```suggestion
                :class:`~qiskit.primitives.Estimator`, will be used.
```

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -0,0 +1,245 @@
...
+            circuit: The quantum circuit to represent the neural network.
+            observables: The observables for outputs of the neural network. If ``None``,
+                use the default :math:`Z^{\otimes num\_qubits}` observable.
+            input_params: The parameters that correspond to the input data of the network.
+                If ``None``, the input data is not bound to any parameters.
+            weight_params: The parameters that correspond to the trainable weights.
+                If ``None``, the weights are not bound to any parameters.
+            gradient: The estimator gradient to be used for the backward pass.
+                If None, a default instance of the estimator gradient,
+                :class:`~qiskit.algorithms.ParamShiftEstimatorGradient`, will be used.
```
woodsp-ibm, line 80:
```suggestion
                :class:`~qiskit.algorithms.gradients.ParamShiftEstimatorGradient`, will be used.
```

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -0,0 +1,245 @@
...
+            input_gradients: Determines whether to compute gradients with respect to input data.
+                Note that this parameter is ``False`` by default, and must be explicitly set to
+                ``True`` for a proper gradient computation when using ``TorchConnector``.
```
woodsp-ibm, line 83: Do we want to change TorhcConnector here to a class ref. so it becomes a link rather than just a highlighted term

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -0,0 +1,245 @@
...
+    """A Neural Network implementation based on the Estimator primitive.
+
+    The ``EstimatorQNN`` is a neural network that takes in a parametrized quantum circuit
+    with the combined network's feature map (input parameters) and ansatz (weight parameters)
+    and outputs its measurements for the forward and backward passes.
+
+    Attributes:
+
+        estimator (BaseEstimator): The estimator primitive used to compute the neural network's results.
+        gradient (BaseEstimatorGradient): An optional estimator gradient to be used for the backward
```
woodsp-ibm, line 50: Should we remove `optional` from the text here. I do not think its optional - in the constructor it defaults to ParamShiftEstimatorGraident. With the attribute here you will read that or the gradient you supplied. If you supply a gradient here it should not be None - the typehint does not allow that and I imagine the code may fail trying to deference None below.

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+class EstimatorQNN(NeuralNetwork):
+    """A Neural Network implementation based on the Estimator primitive.
+
+    The ``EstimatorQNN`` is a neural network that takes in a parametrized quantum circuit
+    with the combined network's feature map (input parameters) and ansatz (weight parameters)
+    and outputs its measurements for the forward and backward passes.
+
+    Attributes:
```
woodsp-ibm, line 47: I would suggest adding text like this ahead of Attributes otherwise these appear right after the preceding text 


> The following attributes can be set via the initializer but can also be read and
> updated once the EstimatorQNN object has been constructed.

We have done this in algorithms in Terra eg https://github.com/Qiskit/qiskit-terra/blob/ad7e609d10b02fd98090b0fac132113f329825b6/qiskit/algorithms/time_evolvers/pvqd/pvqd.py#L47-L48

a-matsuo:
Thank you @woodsp-ibm! I reflected your comments.

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -0,0 +1,249 @@
...
+        gradient (BaseEstimatorGradient): The estimator gradient to be used for the backward
+            pass.
+
+    A Neural Network implementation based on the Estimator primitive."""
+
+    def __init__(
+        self,
+        *,
+        estimator: BaseEstimator | None = None,
+        circuit: QuantumCircuit,
```
ElePT, line 62: @adekusar-drl suggested putting the `circuit` argument first on the `SamplerQNN`, as it has no default. I am fine with either option because in the end these are keyword-only args, but I think we should be consistent on both QNNs.

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -0,0 +1,249 @@
...
+        return input_grad, weights_grad
+
+    def _backward(
+        self, input_data: np.ndarray | None, weights: np.ndarray | None
+    ) -> Tuple[np.ndarray | None, np.ndarray]:
+        """Backward pass of the network."""
+        # prepare parameters in the required format
+        parameter_values_, num_samples = self._preprocess(input_data, weights)
+
+        if num_samples is None or not (self._input_gradients or self._num_weights):
```
adekusar-drl, line 228: Could this be simpler?

```suggestion
        if num_samples is None or (self._input_gradients is False and self._num_weights == 0):
```

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -0,0 +1,249 @@
...
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Estimator quantum neural network class"""
+
+from __future__ import annotations
+
+from copy import copy
+import logging
+from typing import Sequence, Tuple
```
ElePT, line 19: the `tuple` annotation is included in the `__future__` `annotations`
```suggestion
from typing import Sequence
```

adekusar-drl: And replace `Tuple` to `tuple` down below :)

qiskit_machine_learning/neural_networks/estimator_qnn.py
a-matsuo: I missed it. Ya, let's make `SamplerQNN` and `EstimatorQNN` consistent. Thank you!

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -0,0 +1,249 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class EstimatorQNN(NeuralNetwork):
+    """A Neural Network implementation based on the Estimator primitive.
+
+    The ``EstimatorQNN`` is a neural network that takes in a parametrized quantum circuit
+    with the combined network's feature map (input parameters) and ansatz (weight parameters)
+    and outputs its measurements for the forward and backward passes.
```
ElePT, line 45: I think that this definition could be updated to sth like:

```suggestion
    The ``EstimatorQNN`` is a neural network that takes in and a parametrized quantum circuit
    with the combined network's feature map (input parameters) and ansatz (weight parameters), as well as 
    an observable, and outputs their expectation value computations for the forward and backward passes.
```

qiskit_machine_learning/neural_networks/estimator_qnn.py
woodsp-ibm: `self._input_gradients is False` for None its fine - for False it should be better to do `not self._input_gradients` or ` == False` otherwise `is False` relies on singleton behavior to ensure it `is` the same instance

adekusar-drl: PEP 8: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
So, it is well written. Just double checked, pylint complains on `if xx == False`. And `not` reduces readability in this case. I'm in favor of this style as I suggested this change earlier :)

qiskit_machine_learning/neural_networks/estimator_qnn.py
adekusar-drl: @woodsp-ibm are you fine with this proposal?

qiskit_machine_learning/neural_networks/estimator_qnn.py
woodsp-ibm: https://peps.python.org/pep-0008/#programming-recommendations

```
Don’t compare boolean values to True or False using ==:

# Correct:
if greeting:

# Wrong:
if greeting == True:

Worse:

# Wrong:
if greeting is True:
```
I thought I had read that some Python implementations do not have singletons hence the `is` would not work in that case. I think your ref was to a flake rule right. Yes == is considered incorrect.

qiskit_machine_learning/neural_networks/estimator_qnn.py
woodsp-ibm: It takes a single circuit parameterized circuit right where you designate parameters for input and/or weights. Will a newcomer be confused by seeing terminology like feature map and anstaz here? Sure you can concatentate one of the library feature map circuits with one its ansatz circuits (as TwoLocal did before) but is this a good basic description or would it be more an example. Maybe the part given over to input params we want to refer to it as a feature map but I think only this above text here does that right, nowhere else in the qnn - same for ansatz.

adekusar-drl: I'm not sure I got the point. I updated the description slightly. Feel free to suggest changes if you think it's necessary.

woodsp-ibm: My point was more that when you come to QNNs this just takes a circuit however you make it. The description seemed to lean to heavily into feature map/ansatz as terms to describe its operation which this does not have (unlike TwoLocal where it has like named parameters). The text is much better now in that it makes the feature map + anstaz an example of what can be done so it seems you got the basic point I was trying to make, I think.

Maybe we could even go a bit further and have a small code example in the class docstring that builds an instance with a feature map and ansatz combined circuit, that follows on from that text, which I think may helpful for someone coming to ML. Sure we will have tutorials elsewhere that should cover this, but something very simple here in the API ref to show this.

adekusar-drl: Added a code snippet.

woodsp-ibm (approved):

PR #478, opened 2022-09-26
Tutorial for Quantum Autoencoder to compress quantum states

murphytarra:
### Summary
This is a tutorial which goes through the theory of a Quantum Autoencoder, followed by several examples of compressing different quantum states.


Zoufalc (changes requested):
Thanks for this great contribution.
Unfortunately, I cannot directly comment into the file so I'll put my points down here.

- While QAE is the natural abbreviation for Quantum Auto Encoder - it is the abbreviation that is generally used for Quantum Amplitude Estimation. Better change it to QAuEn or something along those lines.
- There is a minor typo in the classical auto encoder section `which has a less nodes`.
- Please try to avoid the term "ancilla". This term is connected to some negative history. The preferred terminology is "auxiliary qubit".
- The goal of the algorithm should be explained in detail in the introduction.
- Why do you use a Swap test instead of a simple overlap? The increase in the number of qubits could be problematic for near-term applications.
- I suggest renaming Sec. 7. to `A Quantum Auto Encoder for Digit Compression`
- What I would also like to see is a discussion on applications of quantum auto encoders. This can be very beneficial for the user's understanding.

adekusar-drl:
Thanks for the comments. There are a few updates, see below what has been changed.

>  While QAE is the natural abbreviation for Quantum Auto Encoder - it is the abbreviation that is generally used for Quantum Amplitude Estimation. Better change it to QAuEn or something along those lines.

Now it is just plain "Quantum Autoencoder" to avoid any confusions.

>  There is a minor typo in the classical auto encoder section `which has a less nodes`.

Fixed.

>  Please try to avoid the term "ancilla". This term is connected to some negative history. The preferred terminology is "auxiliary qubit".

Good point, replaced with "auxiliary".

>  The goal of the algorithm should be explained in detail in the introduction.

Added some additional explanations in the introduction.

>  Why do you use a Swap test instead of a simple overlap? The increase in the number of qubits could be problematic for near-term applications.

The tutorial does not pretend to be the only right solution. It is merely implements what is stated in the referenced paper. Other approaches may require additional research and trials.

>  I suggest renaming Sec. 7. to `A Quantum Auto Encoder for Digit Compression`

Renamed.

>  What I would also like to see is a discussion on applications of quantum auto encoders. This can be very beneficial for the user's understanding.

Added a few applications.

adekusar-drl:
@woodsp-ibm I have updated the tutorial to the new QNNs. I'm happy with the content, Christa had a look as well. If you could take a look it would be nice, but no pressure now. I guess, if I mark it as "stable backport" it will be backported and published on the website, right?

woodsp-ibm:
It needs to go over to stable to get published yes. I think the docs need to be triggered @manoelmarques 

I am curious on this below though in regards of what warning is being ignored here. Whatever it is I guess its of no concern, but then if its raising some warning - or maybe this is from something its using - can it not be sorted?

```
# we suppress a warning that comes from RawFeatureMap
warnings.filterwarnings("ignore")

qnn = SamplerQNN(
    circuit=qc,
    input_params=fm.parameters,
    weight_params=ae.parameters,
    interpret=identity_interpret,
    output_shape=2,
)
```

woodsp-ibm:
At the start of section 5 it has some imports which still include Aer - I do not think thats needed any longer right (I cannot comment in the file at all since github says its too large to show the diff)
```
import json
import time
import warnings

import matplotlib.pyplot as plt
import numpy as np
from IPython.display import clear_output
from qiskit import ClassicalRegister, QuantumRegister
from qiskit import QuantumCircuit, Aer <<<-------------
```

woodsp-ibm:
At the end of the notebook, in the applications of section 8.

`This is commonly used to describe electron - electron interactions in molecules.`

I had to re-read that as I first tried to read it as a dash - not a hyphen.

Think it should more be (no spaces around the - so words (electron) are joined i.e. use it as a hyphen.

`This is commonly used to describe electron-electron interactions in molecules.`

adekusar-drl:
> 
> I am curious on this below though in regards of what warning is being ignored here. Whatever it is I guess its of no concern, but then if its raising some warning - or maybe this is from something its using - can it not be sorted?
> 
> ```
> # we suppress a warning that comes from RawFeatureMap
> warnings.filterwarnings("ignore")
> 
> qnn = SamplerQNN(
>     circuit=qc,
>     input_params=fm.parameters,
>     weight_params=ae.parameters,
>     interpret=identity_interpret,
>     output_shape=2,
> )
> ```

Not required anymore, good point. This was need  to suppress warnings from `CircuitQNN` with `RawFeatureVector` when the former tries to evaluate a gradient circuit. Removed now.

adekusar-drl:
> At the start of section 5 it has some imports which still include Aer - I do not think thats needed any longer right (I cannot comment in the file at all since github says its too large to show the diff)

Aer is removed now.

adekusar-drl:
> At the end of the notebook, in the applications of section 8.

Updated `-` and the numbered list, should look good now (locally it is good).

woodsp-ibm (approved):

PR #485, opened 2022-10-05
Fix QuantumKernel evaluate_duplicates docstring

### LINKED ISSUE #484: QuantumKernel `evaluate_duplicates` docstring is not formatted well

The evaluate_duplicates was added in #432 but its seems the formatting of the str values and their explanation could do with improvement. It comes out as below which brings part of the text along with the value making it harder to deal with. (I noticed when looking at the docs from another newer as yet unmerged PR)

image

manoelmarques:
### Summary
Fixes #484


woodsp-ibm (approved):
Looks fine now thanks - well as good as we've figured with bullet lists in an arg docstring. For some reason the arg text always splits to next line when it has a list like this in it rather than starting following the arg itself, oh well!

PR #414, opened 2022-06-14
Pytorch QGAN Notebook

Zoufalc:
### Summary
This PR includes a notebook that illustrates how to build a qGAN with PyTorch and qiskit-machine-learning resources.

### Details and comments
@ElePT Your help and feedback is much appreciated!

PR #495, opened 2022-10-20
Deprecate distribution_learners package

adekusar-drl:
### Summary
This PR deprecates everything in the `qiskit_machine_learning.algorithm.distribution_learners` package. The package is replaced with a new tutorial introduced in #414. This PR to be merged after the new tutorial.


qiskit_machine_learning/algorithms/distribution_learners/__init__.py
```diff
@@ -20,6 +20,17 @@
     QuantumGenerator,
     QGAN,
 )
+from ...deprecation import warn_deprecated, DeprecatedType, MachineLearningDeprecationWarning
+
+warn_deprecated(
```
woodsp-ibm, line 25: I am not sure with this. In algorithms.init file we import these so I think users are going to see messages whether they use the classes or not. We may need to put deprecation more local in constructors. @manoelmarques any thoughts here. If we do that we may also want to change the first line of the class doctring to prefix it with Deprecated:  so its evident in the docs. Since all the classes are documented in algorithms init though an alternative/addition would be to state they are deprecated there too.

Also we normally adjust unit tests of deprecated code, since we know it will emit such messages, so as to restrict what comes out in CI to deprecations we should be taking care of.

releasenotes/notes/deprecate-distribution-learners-196741e53a8cecc5.yaml
```diff
@@ -0,0 +1,7 @@
+---
+deprecations:
+  - |
+    The `qiskit_machine_learning.algorithms.distribution_learners` package is deprecated. There's no
+    a direct replacement for the classes from this package. Instead, please refer to the new QGAN
```
woodsp-ibm, line 5:
```suggestion
    direct replacement for the classes from this package. Instead, please refer to the new QGAN
```

releasenotes/notes/deprecate-distribution-learners-196741e53a8cecc5.yaml
```diff
@@ -0,0 +1,7 @@
+---
+deprecations:
+  - |
+    The `qiskit_machine_learning.algorithms.distribution_learners` package is deprecated. There's no
```
woodsp-ibm, line 4: I think we normally state, after noting its deprecated something like - and will be removed in a future version.

qiskit_machine_learning/algorithms/distribution_learners/__init__.py
manoelmarques: If `distribution_learners` folder is included in any `init` above it, then unfortunately the deprecation can't be at the `init` level otherwise any unrelated import will import it and show the deprecation msg. In  that case it needs to be in the `init` of each class inside the folder.

qiskit_machine_learning/deprecation.py
```diff
@@ -140,10 +148,10 @@ def warn_deprecated(
         type_str = new_type.value if new_type is not None else old_type.value
         msg += f". Instead use the {new_name} {type_str}"
     if additional_msg:
-        msg += f" {additional_msg}"
+        msg += f". {additional_msg}"
```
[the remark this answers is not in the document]
adekusar-drl: Rolled back the update. Hope the deprecation message does not look very ugly now.

qiskit_machine_learning/algorithms/distribution_learners/__init__.py
adekusar-drl: Moved deprecations to the class level.

releasenotes/notes/deprecate-distribution-learners-196741e53a8cecc5.yaml
adekusar-drl: Updated the reno, added "and will be removed no sooner than 3 months after the release".

woodsp-ibm:
Ideally this should update the unit tests as well - since we know they will emit deprecation messages to suppress them. Maybe this can be done in a separate PR since it seems there are already a bunch of units tests using former QuantumKernel etc that are emitting a bunch of warnings that CI is catching that we should be suppressing.

adekusar-drl:
> Ideally this should update the unit tests as well - since we know they will emit deprecation messages to suppress them. Maybe this can be done in a separate PR since it seems there are already a bunch of units tests using former QuantumKernel etc that are emitting a bunch of warnings that CI is catching that we should be suppressing.

Do you mean update the test to suppress deprecation messages? Or something else?

woodsp-ibm:
> Do you mean update the test to suppress deprecation messages?

Yes, any unit tests which we know is testing deprecated logic, while we continue to support it ahead of its removal, which will emit deprecated messages when used/called, these warnings should be suppressed by the unit tests.

adekusar-drl:
> > Do you mean update the test to suppress deprecation messages?
> 
> Yes, any unit tests which we know is testing deprecated logic, while we continue to support it ahead of its removal, which will emit deprecated messages when used/called, these warnings should be suppressed by the unit tests.

Done for QGANs.

woodsp-ibm (approved):

PR #499, opened 2022-10-25
fix cross reference of classes from terra

### LINKED ISSUE #474: Update references to Terra's classes in docstrings

Review all docstrings in the package and update references to Terra's classes. For instance, in the `VQC` documentation we reference \`\`ZZFeatureMap\`\` that is in `qiskit.circuit.library`. When the package documentation is generated, the reference becomes just a plain text. Such references should be replaced with cross references, that are replaced in the generated documentations with corresponding links. In this example, in the documentation \`\`ZZFeatureMap\`\` to be replaced with \`:class:\`~qiskit.circuit.library.ZZFeatureMap\`. Then in the html documentation this cross reference will become a link to the corresponding class in Terra.

Ryand1234:
### Summary

Fixes #474 

### Details and comments
Replaced plain class names from terra with cross references.

### DIFF
--- a/qiskit_machine_learning/algorithms/classifiers/vqc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/vqc.py
@@ -66,11 +66,11 @@ class VQC(NeuralNetworkClassifier):
                 are adjusted to this number if required.
             feature_map: The (parametrized) circuit to be used as a feature map for the underlying
                 :class:`~qiskit_machine_learning.neural_networks.CircuitQNN`. If ``None`` is given,
-                the ``ZZFeatureMap`` is used if the number of qubits is larger than 1. For a single
-                qubit classification problem the ``ZFeatureMap`` is used per default.
+                the :class:`~qiskit.circuit.library.ZZFeatureMap` is used if the number of qubits is larger than 1. For a single
+                qubit classification problem the :class:`~qiskit.circuit.library.ZZFeatureMap` is used per default.
             ansatz: The (parametrized) circuit to be used as an ansatz for the underlying
                 :class:`~qiskit_machine_learning.neural_networks.CircuitQNN`. If ``None`` is given
-                then the ``RealAmplitudes`` circuit is used.
+                then the :class:`~qiskit.circuit.library.RealAmplitudes` circuit is used.
             loss: A target loss function to be used in training. Default value is ``cross_entropy``.
             optimizer: An instance of an optimizer to be used in training. When ``None`` defaults
                 to SLSQP.
--- a/qiskit_machine_learning/algorithms/regressors/vqr.py
+++ b/qiskit_machine_learning/algorithms/regressors/vqr.py
@@ -51,11 +51,11 @@ class VQR(NeuralNetworkRegressor):
                 are adjusted to this number if required.
             feature_map: The (parametrized) circuit to be used as a feature map for the underlying
                 :class:`~qiskit_machine_learning.neural_networks.TwoLayerQNN`. If ``None`` is given,
-                the ``ZZFeatureMap`` is used if the number of qubits is larger than 1. For a single
-                qubit regression problem the ``ZFeatureMap`` circuit is used per default.
+                the :class:`~qiskit.circuit.library.ZZFeatureMap` is used if the number of qubits is larger than 1. For a single
+                qubit regression problem the :class:`~qiskit.circuit.library.ZZFeatureMap` circuit is used per default.
             ansatz: The (parametrized) circuit to be used as an ansatz for the underlying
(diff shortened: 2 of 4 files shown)


adekusar-drl (commented):
Thanks for taking care of this issue. The PR is a good starting point, but there are other places to look at:
- `QuantumKernel`
- `QuantumKernelTrainer`, there local and non-local references
- `RawFeatureVector`

And other classes to check out.

Ryand1234:
Hi @adekusar-drl in this issue and PR we were changing references from class that belongs to terra. Will it be okay to update references for classes which belongs to this repo like QuantumKernal in this PR? or should I create new PR for it?

adekusar-drl:
@Ryand1234 certainly, you can update reference to all classes you find in the docstrings, either they are in Terra or here in QML. The point is that if the references are properly formatted, they become hyper links in the online documentation.

woodsp-ibm:
For the CLA it goes by the email in the commits which comes from your local git config. If it cannot resolve that email to a github user id then it shows the user name from your git config. Maybe you used different machines to do this PR. One way to fix this is to add the other email to your github account.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -84,11 +84,13 @@ def __init__(
 
         Raises:
             ValueError:
-                - if ``quantum_kernel`` is passed and ``precomputed`` is set to ``True``. To use
-                a precomputed kernel, ``quantum_kernel`` has to be of the ``None`` type.
+                - if :class:`~qiskit_machine_learning.kernels.FidelityQuantumKernel` is passed and
```
woodsp-ibm, line 87: I think this change to make things specific makes it incorrect. The `quantum_kernel` parameter must be None if you say you will use a `precomputed` kernel when that parameter is True. A quantum_kernel by type just has to be subclass of BaseKernel - so by making some specific reference it changes the meaning somewhat.

Now its true that if precomputed is False and you do not pass a quantum_kernel, i.e. you leave it at None, then it creates an instance of that specific subclass - but overall its not limited to that. Arguably this default when None and precomputed is False ought to be noted in the docstring above.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -84,11 +84,13 @@ def __init__(
...
+                ``precomputed`` is set to ``True``. To use a precomputed kernel,
+                 :class:`~qiskit_machine_learning.kernels.FidelityQuantumKernel` has to be of
+                 the ``None`` type.
             TypeError:
-                - if ``quantum_kernel`` neither instance of
-                  :class:`~qiskit_machine_learning.kernels.BaseKernel` nor ``None``.
+                - if :class:`~qiskit_machine_learning.kernels.FidelityQuantumKernel` neither instance
+                  of :class:`~qiskit_machine_learning.kernels.BaseKernel` nor ``None``.
```
woodsp-ibm, line 93: This I think can be completely removed - the code did raise a TypeError but that seems to have been removed in a recent change since things are now more general around  a BaseKernel rather than what was there before.


On thing I do note is that is that this routine raises a ValueError if C <= 0 i.e. not a positive number, which is not included above.

qiskit_machine_learning/algorithms/classifiers/vqc.py
```diff
@@ -66,11 +66,12 @@ def __init__(
...
+                the :class:`~qiskit.circuit.library.ZZFeatureMap` is used if the number of qubits
+                is larger than 1. For a single qubit classification problem the
+                :class:`~qiskit.circuit.library.ZZFeatureMap` is used per default.
             ansatz: The (parametrized) circuit to be used as an ansatz for the underlying
                 :class:`~qiskit_machine_learning.neural_networks.CircuitQNN`. If ``None`` is given
-                then the ``RealAmplitudes`` circuit is used.
+                then the :class:`~qiskit.circuit.library.RealAmplitudes` circuit is used.
             loss: A target loss function to be used in training. Default value is ``cross_entropy``.
             optimizer: An instance of an optimizer to be used in training. When ``None`` defaults
                 to SLSQP.
```
woodsp-ibm, line 77: SLSQP here could be written as ``:class:`~qiskit.algorithms.optimizers.SLSQP` `` since this is the actual optimizer class it defaults to. This default is actually done in `TrainableModel` class where it could be altered the same way,

adekusar-drl:
@Ryand1234 will you continue to work on this pull request?

adekusar-drl:
@Ryand1234 Are there any  updates here?

adekusar-drl:
Closing the PR as stale.

PR #506, opened 2022-11-01
Deprecate QNNs

adekusar-drl:
### Summary
This PR deprecates `CircuitQNN`, `OpflowQNN`, and `TwoLayerQNN`. To be merged after #436.


qiskit_machine_learning/neural_networks/two_layer_qnn.py
```diff
@@ -18,17 +18,25 @@
...
 from ..utils import derive_num_qubits_feature_map_ansatz
 
 
 class TwoLayerQNN(OpflowQNN):
-    """Two Layer Quantum Neural Network consisting of a feature map, a ansatz,
+    """Pending deprecation: Two Layer Quantum Neural Network consisting of a feature map, a ansatz,
     and an observable.
     """
 
+    @deprecate_function(
```
woodsp-ibm, line 32: Jus checking - If I create a TwoLayerQNN do I just get the one deprecation message or a second when it calls super.init0 from OpflowQNN. I am thinking the latter is more an internal call so I wold just have the one message.

releasenotes/notes/pending-deprecate-qnns-ef003277a841096d.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - |
+    Class :class:`qiskit_machine_learning.neural_networks.CircuitQNN` is pending deprecation and
+    is superseded by :class:`qiskit_machine_learning.neural_networks.SamplerQNN`.
+
+    Class :class:`qiskit_machine_learning.neural_networks.OpflowQNN` is pending deprecation and
+    is superseded by :class:`qiskit_machine_learning.neural_networks.EstimatorQNN`.
+
+    Class :class:`qiskit_machine_learning.neural_networks.TwoLayerQNN` is pending deprecation and
+    has no direct replacement. Please make use of
+    :class:`qiskit_machine_learning.neural_networks.EstimatorQNN` instead.
```
woodsp-ibm, line 12: Should we note all of the above will be deprecated and subsequently removed in future releases. By default you don't get to see pending deprecation messages so while that extra info is in them maybe a single sentence covering all of the above classes would be helpful here too.

woodsp-ibm (dismissed):

qiskit_machine_learning/neural_networks/two_layer_qnn.py
adekusar-drl: By default, there are no warnings printed. If I enable them, I get two warning. I don't know if that should be changed. I'd leave as now.

releasenotes/notes/pending-deprecate-qnns-ef003277a841096d.yaml
adekusar-drl: Added "These classes will be deprecated in a future release and subsequently removed after that."
Initially the reno was copied from the QK deprecation reno. I updated it similarly as well.

qiskit_machine_learning/neural_networks/two_layer_qnn.py
woodsp-ibm: Possibly the TwoLayerQNN could suppress the warning around its call to super().init We may want to check that when we change to deprecated as then the warnings will print by default.

woodsp-ibm (approved):

PR #517, opened 2022-11-09
Fix formatting in 02a Training a Quantum Model on a Real Dataset

adekusar-drl:
### Summary
Now formatting is broken despite everything was fine when the original tutorial was merged:

image

Now it should be like this:

image


woodsp-ibm (approved):
This will be the first stable backport - it should be to the stable/0.5 branch. In Nature despite updating the config mergify was still targeting the 0.4 branch. Lets see, hopefully it fine here. 

**Update**: worked fine, the right branch was targeted, so the problem we had with Nature did not manifest here.

PR #523, opened 2022-11-14
Grammar Fixes

AngeloDanducci:
### Summary
Fixes an issue in another repo that points to here https://github.com/Qiskit/qiskit/issues/1291

### Details and comments
A few grammar fixes in `01_neural_networks.ipynb`

### DIFF
--- a/docs/tutorials/01_neural_networks.ipynb
+++ b/docs/tutorials/01_neural_networks.ipynb
@@ -9,7 +9,7 @@
     "\n",
     "This notebook demonstrates the different generic quantum neural network (QNN) implementations provided in Qiskit Machine Learning.\n",
     "The networks are meant as application-agnostic computational units that can be used for many different use cases. \n",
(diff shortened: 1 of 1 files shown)


adekusar-drl:
@woodsp-ibm looks good to me. What do you think?

woodsp-ibm (approved):
It looks fine to me. Maybe we want to mark it as stable backport to get it over to stable - though perhaps we do not need from there to update/re-publish docs quite yet, since you have more updates to these notebooks coming?

adekusar-drl:
For the minor update like this, I think, we can postpone documentation update.

woodsp-ibm:
Ok it still could have been backported. We'll have to check when we do backport docs that the change done here is catered to so that we do not just backport changes done after this and perhaps miss this.

adekusar-drl:
It is still can be backported.

PR #538, opened 2022-12-06
Fix CI aer build

manoelmarques:
### Summary
Github actions updated Ubuntu-latest OS so it needed a change in the Aer build


adekusar-drl (approved):
we still have to fix the problem with CI: https://github.com/Qiskit/qiskit-terra/pull/9242

adekusar-drl (approved):

PR #548, opened 2022-12-14
Add StatevectorKernel

### LINKED ISSUE #551: Add FidelityStatevectorKernel, an optimized kernel that makes use of the statevector simulator
### What should we add?

Adds a `FidelityStatevectorKernel` class implementation that has been optimized for use with statevector-implemented feature maps. Computed statevector arrays are also stored in a statevector cache to avoid repeated computation.

The performance of this implementation should be closer to the previous implementation of `QuantumKernel` running on a statevector simulator. Thus complexity is reduced from O(N^2) to O(N).

declanmillar:
### Summary

Adds a `StatevectorKernel` class implementation that has been optimized for use with statevector-implemented feature maps. Computed statevector arrays are also stored in a statevector cache to avoid repeated computation.

The performance of this implementation should be closer to the previous implementation of QuantumKernel running on a statevector simulator. Thus complexity is reduced from O(N^2) to O(N).

### Details and comments

~~Currently, the implementation includes an `enforce_psd` argument in conjunction with `FidelityQuantumKernel`, however, I'm fairly positive that the kernel matrix is guaranteed to be PSD when using the statevector-based feature mapping. If this is confirmed, I will remove this argument and functionality.~~ Removed.

Closes #551


qiskit_machine_learning/kernels/statevector_kernel.py
```diff
@@ -0,0 +1,190 @@
...
+from __future__ import annotations
+
+import numpy as np
+from qiskit import QuantumCircuit
+from qiskit.quantum_info import Statevector
+
+from .base_kernel import BaseKernel
+
+
+class StatevectorKernel(BaseKernel):
```
ElePT, line 23: Naming question, should we go for `StatevectorQuantumKernel` for consistency with `FidelityQuantumKernel`?? I normally prefer shorter names, but you know, the usual discussion.

qiskit_machine_learning/kernels/statevector_kernel.py
```diff
@@ -0,0 +1,190 @@
...
+
+    These are stored in a statevector cache for reuse to avoid repeated computation. This stash
+    can be cleared using :meth:`clear_cache()`.
+    """
+
+    def __init__(
+        self,
+        *,
+        feature_map: QuantumCircuit | None = None,
+        enforce_psd: bool = True,
```
ElePT, line 66: I agree with your comment, I think that the kernel matrix in absence of noise is guaranteed to be PSD, so we could probably remove this argument.

ElePT (commented):
Thanks for the PR!

qiskit_machine_learning/kernels/statevector_kernel.py
declanmillar: Yeah, I think that's more consistent as you say, and am happy to change it to that.

qiskit_machine_learning/kernels/statevector_kernel.py
woodsp-ibm: The PSD conversion was added in the past for qasm simulation/real device. It was not needed for statevector (ideal outcome).

qiskit_machine_learning/kernels/statevector_kernel.py
```diff
@@ -0,0 +1,190 @@
...
+    def _compute_kernel_element(x: Statevector, y: Statevector) -> float:
+        return np.abs(np.conj(x) @ y) ** 2
+
+    def _get_statevector(self, param_values) -> Statevector:
+        param_values = tuple(param_values)
+        statevector = self._statevector_cache.get(param_values, None)
+
+        if statevector is None:
+            qc = self._feature_map.bind_parameters(param_values)
+            statevector = Statevector(qc).data
```
woodsp-ibm, line 146: This relates to Anton's comment https://github.com/Qiskit/qiskit-machine-learning/pull/548#pullrequestreview-1217421920. Should this here use AerStatevector, if Aer is installed, dropping back to the reference one in Terra - on basis Aer should be more performant? I guess as per the comment one could have some sort of parameter eg pass the class so even a 3rd party StateVector subclass could be used - wondering if a parameter is really needed though since to me this more akin to a reference kernel and limited to what can be done classically.

adekusar-drl: Aer is not always faster, a quick test showed that on a small number of qubits, eg. 10 or so, Terra's Statevector is faster, but when I increased the number of qubits AerStatevector became faster than the original one. But this may be true only in my case. I'd add a parameter, type of the statevector implementation.

gentinettagian:
Thank you for this PR! As I designed the new QuantumKernel interface I thought I might drop some comments, feel free to ignore. 
- This kernel is still a fidelity-kernel, so having a new class with a name that doesn't include `Fidelity` might be confusing. In theory the ideal solution would be to implement a `StatevectorStateFidelity` class that implements the `StateFidelity` interface in way that allows for caching. I haven't thought this through entirely but if it is possible I think this would be the cleaner way to go and would also allow this class to be used wherever the `StateFidelity` is used in the codebase. 
- If we stick to implementing this feature in a separate kernel class, I would try to avoid overriding the `evaluate` method and using the private `_get_kernel_entries` method instead. This will allow for an easy extension for a trainable version of this class similarly to the `TrainableFidelityQuantumKernel`.

ElePT:
Hi @gentinettagian! Thanks so much for taking the time and going through the PR. I have reflected a bit on your comments, and I have a few conclusions I'd like to share. TL;DR: I don't think it's worth it to change the current implementation. 

1. About having a `StatevectorStateFidelity` class: our implementation is highly adapted to the case of quantum kernels, where we know that we have a single circuit and two parameter sets to get state vectors from. This is why, in this implementation, we only cache the parameter values. If we wanted to extend a `StatevectorStateFidelity` implementation to deal with all possible inputs, no assumptions can be made about `circuit_1` and `circuit_2`, and thus, we would have to cache circuits and parameter values, which is something I would like to avoid if possible (there are ongoing discussions about this, the way it is done in the primitives has lead to problems). On top of that, I discussed it with @Cryoris, and it seems like the caching feature in particular seems to benefit a specific subset of problems we are tackling with kernels, but no to the generic set of problems where fidelities are used. In other words, it doesn't seem to be worth it to go though the hassle of implementing a generic class for something that might be useful only for us. 
2. About having the `StatevectorKernel` inherit from `FidelityQuantumKernel`: Yes, we could inherit and then overwrite the `_get_kernel_entries` method (without a new fidelity class), but then we would be inheriting a bunch of other methods and properties based on primitives/fidelities that we will not be using. I don't think this is clean.
3. About extending to trainable/pseudo/etc: The way we are currently overwriting `evaluate` allows for easy extension to both. Yes, this will mean having 3 new kernel classes, but I don't think we can avoid this if we want to be explicit about how each class works. And I personally believe that having a clear separation between the primitives and these "hacks" for performance in simulation will be helpful to avoid confusion among users. It is already hard enough to transmit the whole abstraction that comes with the primitives to then use something that looks like it should be using them but it's not. 
4. The current implementation is very lean and easy to understand, we skip unnecessary checks and parameter reshapings because of the assumptions we are working with, benefitting the performance overall. And it still inherits from `BaseKernel` :)
5. We can always discuss naming changes.

Sorry for the long reply, but I hope it helps convey my message.

qiskit_machine_learning/kernels/statevector_kernel.py
```diff
@@ -0,0 +1,190 @@
...
+        eval_duplicates = evaluate_duplicates.lower()
+        if eval_duplicates not in ("all", "off_diagonal", "none"):
+            raise ValueError(
+                f"Unsupported value passed as evaluate_duplicates: {evaluate_duplicates}"
+            )
+        self._evaluate_duplicates = eval_duplicates
+
+        self._statevector_cache = {}
+
+    def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray | None = None) -> np.ndarray:
```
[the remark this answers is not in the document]
woodsp-ibm: As a different thought on the cache you could look at the functools caching decorators e.g  lru_cache which I think could be applied to something like your get statevector method so that you do not have to do explicit caching - an lru cache can be size limited so would alleviate concerns about uncontrolled growth and the user needing to manage things via clear cache.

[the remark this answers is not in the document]
gentinettagian: > I'd hide cache in the implementation without exposing anything and clear it once `evaluate` is called.

This should definitely be optional. For example in `PegasosQSVC` evaluate is called in every step of the training loop, so there it would be crucial to keep the cache.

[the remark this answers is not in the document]
adekusar-drl: > > I'd hide cache in the implementation without exposing anything and clear it once `evaluate` is called.
> 
> This should definitely be optional. For example in `PegasosQSVC` evaluate is called in every step of the training loop, so there it would be crucial to keep the cache.

Not necessarily. `Pegasos` randomly samples elements to compute, so cache should be on the algorithm level, not on the kernel level. Moreover, current implementation does not have any cache anyway :)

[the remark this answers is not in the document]
gentinettagian: > Not necessarily. `Pegasos` randomly samples elements to compute, so cache should be on the algorithm level, not on the kernel level. Moreover, current implementation does not have any cache anyway :)

`Pegasos` samples a new point every iteration and then computes the kernel entries between this point and the (iteratively accumulated) support vectors. So the in a state vector based implementation of the kernel, the state vectors of the support vectors are used in every iterations and caching them would provide a speedup.

Of course so far there is no caching, but if caching is added I think limiting it to a single `evaluate` call might be too restrictive.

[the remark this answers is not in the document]
gentinettagian: Yes, `Pegasos` knows what should be cached, but `Pegasos` only has access to the kernel entries, not to the actual state vectors. So, unless you suggest writing a separate `StatevectorPegasos` that doesn't rely on the kernel interface, the caching has to be implemented within the kernel (or even better in the fidelity instance, but I accept that this is too challenging).

qiskit_machine_learning/kernels/statevector_kernel.py
declanmillar: As discussed elsewhere, I've now renamed the class `FidelityStatevectorKernel`.

adekusar-drl: > As discussed elsewhere, I've now renamed the class `FidelityStatevectorKernel`.

Thanks!

qiskit_machine_learning/kernels/statevector_kernel.py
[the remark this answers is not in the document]
adekusar-drl: Okay, thanks for being patient with me. I see the problem now. Certainly, I'm against `StatevectorPegasos`, but on the other hand I don't like the idea of exposing a caching interface at the `FidelityStatevectorKernel` level as well. Looks like caching makes sense only in the statevector kernel implementation. Any ideas where else any kind of caching may emerge? I'm thinking what bits and pieces can be unified.

qiskit_machine_learning/kernels/statevector_kernel.py
```diff
@@ -0,0 +1,191 @@
...
+
+from .base_kernel import BaseKernel
+
+
+class FidelityStatevectorKernel(BaseKernel):
+    r"""
+    A reference implementation of the quantum kernel interface limited to classically simulated
+    statevectors.
+
+    The general task of machine learning is to find and study patterns in data. For many
```
Cryoris, line 28: This is a nice explanation! Shouldn't this better go into a more general place like the `kernels` init file or so? Hiding general docs into a specific class is makes them harder accessible (and might lead to duplications) 🙂

I would say that starting from 
```
Here, the kernel function is defined as ...
``` 
it really is only class specific 🙂

qiskit_machine_learning/kernels/statevector_kernel.py
```diff
@@ -0,0 +1,191 @@
...
+    .. math::
+
+        K(x,y) = |\langle \phi(x) | \phi(y) \rangle|^2.
+
+    In this implementation, :math:`\phi` is represented by a ``Statevector.data`` array,
+    thus the kernel function is given simply by
+
+    .. math::
+
+        K(x,y) = (\phi(x)^\dagger \phi(y))^2.
```
Cryoris, line 57: This misses the absolute value
```suggestion
        K(x,y) = |\phi(x)^\dagger \phi(y)|^2.
```

qiskit_machine_learning/kernels/statevector_kernel.py
```diff
@@ -0,0 +1,191 @@
...
+
+    These arrays are stored in a statevector cache for reuse to avoid repeated computation. This stash
+    can be cleared using :meth:`clear_cache()`.
+    """
+
+    def __init__(
+        self,
+        *,
+        feature_map: QuantumCircuit | None = None,
+        evaluate_duplicates: str = "off_diagonal",
```
Cryoris, line 67: Since every evaluation is exact, why would one ever want to re-evaluate the same datapoint?

qiskit_machine_learning/kernels/statevector_kernel.py
```diff
@@ -0,0 +1,191 @@
...
+
+        if statevector is None:
+            qc = self._feature_map.bind_parameters(param_values)
+            statevector = self._statevector_type(qc).data
+            self._statevector_cache[param_values] = statevector
+
+        return statevector
+
+    @property
+    def evaluate_duplicates(self):
```
Cryoris, line 153: Without a setter this is "immutable" in theory, is this on purpose? If not, we could also avoid the properties and just make it a public variable 😄

qiskit_machine_learning/kernels/statevector_kernel.py
```diff
@@ -0,0 +1,191 @@
...
+
+        kernel_matrix = np.ones(kernel_shape)
+        for i, x in enumerate(x_svs):
+            for j, y in enumerate(y_svs):
+                if self._is_trivial(i, j, x, y, is_symmetric):
+                    continue
+                kernel_matrix[i, j] = self._compute_kernel_element(x, y)
+
+        # due to truncation and rounding errors we may get complex numbers
+        kernel_matrix = np.real(kernel_matrix)
```
Cryoris, line 133: Is this still the case? Per typehint (and from what numpy returns) it seems that the kernel can only have floats as values 😄

qiskit_machine_learning/kernels/statevector_kernel.py
```diff
@@ -0,0 +1,191 @@
...
+    def _compute_kernel_element(x: Statevector, y: Statevector) -> float:
+        return np.abs(np.conj(x) @ y) ** 2
+
+    def _get_statevector(self, param_values) -> Statevector:
+        param_values = tuple(param_values)
+        statevector = self._statevector_cache.get(param_values, None)
+
+        if statevector is None:
+            qc = self._feature_map.bind_parameters(param_values)
+            statevector = self._statevector_type(qc).data
```
Cryoris, line 147: Here this returns a `numpy.array` not a `Statevector` -- is the typehint wrong or do you want to cache the `Statevector` here? 😄

qiskit_machine_learning/kernels/statevector_kernel.py
ElePT: Naive question, since this is squared... does the absolute value serve any purpose?

qiskit_machine_learning/kernels/statevector_kernel.py
ElePT: Mmmm I believe that this is on purpose, we are generally not using public variables in qiskit-ml, and try to minimize the setters. I think that the idea here is not to let the user modify the behavior of the class mid-use, and encourage creating a new instance instead.

qiskit_machine_learning/kernels/statevector_kernel.py
gentinettagian: the overlap can be complex in general :)

Cryoris: E.g. `\phi(x) = [0, 1]` and `\phi(y) = [0, i]` would evaluate to `-i` and then squaring would give a negative kernel entry 😄

qiskit_machine_learning/kernels/statevector_kernel.py
declanmillar: The typehint is wrong. Thanks!

qiskit_machine_learning/kernels/statevector_kernel.py
declanmillar: I moved this section of text into the `kernels` init file and removed it from the fidelity kernel docstring.

qiskit_machine_learning/kernels/statevector_kernel.py
```diff
@@ -0,0 +1,191 @@
...
+    These arrays are stored in a statevector cache for reuse to avoid repeated computation. This stash
+    can be cleared using :meth:`clear_cache()`.
+    """
+
+    def __init__(
+        self,
+        *,
+        feature_map: QuantumCircuit | None = None,
+        evaluate_duplicates: str = "off_diagonal",
+        statevector_type: type = Statevector,
```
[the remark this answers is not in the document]
declanmillar: I tried it out with Protocol, but think a `np.ndarray` `data` attribute might be too non-specific for duck typing in this case, so I went with your original suggestion as I'm assuming that all Statevectors will inherit the reference implementation. If it's foreseen that some Statevectors won't inherit the reference Statevector, I can change this though.

qiskit_machine_learning/kernels/statevector_kernel.py
```diff
@@ -0,0 +1,129 @@
...
+                    continue
+                kernel_matrix[i, j] = self._compute_kernel_element(x, y)
+
+        return kernel_matrix
+
+    @staticmethod
+    def _compute_kernel_element(x: np.ndarray, y: np.ndarray) -> float:
+        return np.abs(np.conj(x) @ y) ** 2
+
+    def _get_statevector(self, param_values) -> np.ndarray:
```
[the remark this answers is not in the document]
declanmillar: I've refactored the code to use `functools.lru_cache(maxsize=None)` as a Python3.7- and 3.8-compatible `functools.cache`. Since we're wrapping an instance method here, I've made the cache container local to the instance so it can be cleaned up properly, rather than use a decorator.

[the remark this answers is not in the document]
declanmillar: It probably makes sense to put a limit on the cache size, but I'm not sure what a typical user would require. @ElePT do you have an idea of cache size requirements based on what the client that required the protocol implementation was using?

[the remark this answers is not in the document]
woodsp-ibm: Given how you have done things you could have say a `cache_size` parameter on the constructor so whatever default you end up with a user can still adjust it via that parameter.

qiskit_machine_learning/kernels/statevector_kernel.py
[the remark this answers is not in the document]
adekusar-drl: That's fine. Python is not easy for strong typing/type hints.

test/kernels/test_statevector_kernel.py
```diff
@@ -0,0 +1,389 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2022, 2023.
```
ElePT, line 3:
```suggestion
# (C) Copyright IBM 2023.
```

qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
```diff
@@ -0,0 +1,135 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2022, 2023.
```
ElePT, line 3:
```suggestion
# (C) Copyright IBM 2023.
```

test/kernels/test_new_vs_old_kernel_sv.py
```diff
@@ -0,0 +1,93 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021, 2023.
```
ElePT, line 3:
```suggestion
# (C) Copyright IBM 2023.
```

ElePT (commented):
Extremely minor suggestion, I think that the copyright year for the new classes should be directly 2023. That's all I could find to comment on :)

test/kernels/test_statevector_kernel.py
ElePT: Also not a big deal, but shouldn't the name of this file be `test_fidelity_statevector_kernel.py` for consistency?

declanmillar: I did think about that but the other test files aren't all consistent. E.g. `test_fidelity_qkernel.py`, so I didn't bother. I will make the change in this case though.

qiskit_machine_learning/kernels/statevector_kernel.py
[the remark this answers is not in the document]
woodsp-ibm: Out of curiosity do you know how much difference the caching makes to the performance?

qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
woodsp-ibm: Just a minor comment here - the PR was created last year and I think these files at the same time. So technically 2022 would be the first year of copyright if my statement hold true.

qiskit_machine_learning/kernels/statevector_kernel.py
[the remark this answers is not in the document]
adekusar-drl: I can try to test on Pegasos, but honestly, I don't expect dramatic improvements.

docs/tutorials/02a_training_a_quantum_model_on_a_real_dataset.ipynb
```diff
@@ -201,7 +201,7 @@
     },
```
adekusar-drl, line 201: Wonder why figures in this tutorial were updated.

adekusar-drl (commented):
Just left two comments.
@woodsp-ibm is there anything from your side?

woodsp-ibm:
It should have a reno release note right for this new feature.

qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
```diff
@@ -0,0 +1,116 @@
...
+    produced by a parametrized quantum circuit (called feature map):
+
+    .. math::
+
+        K(x,y) = |\langle \phi(x) | \phi(y) \rangle|^2.
+
+    In this implementation, :math:`|\phi(y)\rangle` is given by a ``Statevector.data`` array.
+    These arrays are stored in a statevector cache for reuse to avoid repeated evaluation of the
+    quantum circuit. This cache can be cleared using :meth:`clear_cache`. By default the cache is
+    cleared when :meth:`evaluate` is called, unless ``retain_cache`` is ``True``.
```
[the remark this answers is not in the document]
woodsp-ibm: > These arrays are stored in a statevector cache for reuse to avoid repeated evaluation of the quantum circuit.

Pretty minor comment, could change it to add commas since the for reuse should be implied by it being a cache

These arrays are stored in a statevector cache`,` for reuse`,` to avoid repeated evaluation of the quantum circuit.

or simply remove

These arrays are stored in a statevector cache to avoid repeated evaluation of the quantum circuit.

docs/tutorials/02a_training_a_quantum_model_on_a_real_dataset.ipynb
declanmillar: That's an accidental inclusion. I'll remove those changes.

woodsp-ibm:
I was looking at the docs, and see types like QuantumCircuit here do not become links via intersphinx - only basic types from Python. In looking at the docs overall it seems to be a more general problem - at first sight it appears typehints done using the old Union way, e.g [CircuitQNN](https://qiskit.org/documentation/machine-learning/stubs/qiskit_machine_learning.neural_networks.CircuitQNN.html#qiskit_machine_learning.neural_networks.CircuitQNN) the outcome is fine, done the new | way, e.g [VQC](https://qiskit.org/documentation/machine-learning/stubs/qiskit_machine_learning.algorithms.VQC.html#qiskit_machine_learning.algorithms.VQC), not so much. I guess it warrants a separate issue really - but as I noticed it here I figured I'd note it here to start.

adekusar-drl (approved):
Looks good to me!

PR #553, opened 2023-01-23
[WIP] Add Sampling QNN

### LINKED ISSUE #556: Add sampling quantum neural network on top of primitives
### What should we add?

Sampling feature is available in `CircuitQNN` but is missing in the new primitive based QNNs. Sampling is required for the QGAN tutorial.

adekusar-drl:
### Summary
The goal of this PR is to add a new QNN that brings in sampling functionality on top of primitives. A new `SamplingQNN` closed this gap.

### Details and comments
Changes:
- a new `SamplingQNN` that extends `SamplerQNN`. Since direct sampling is not possible with current primitives, the QNN samples samples from quasi-distributions returned by Sampler. To be revised once new primitives are introduced.
- Some fixes in `SamplerQNN` to allow circuits without inputs and/or weights.
- Some tweaks in the backward pass of `EstimatorQNN` to make it aligned with `SamplerQNN`.
- Removed unnecessary normalization in `SamplerQNN`.
- Some other tweaks in the QNNs.

Updates in the QGAN tutorial to follow in a separate PR.
Closes #556.


adekusar-drl:
@ElePT I'm kindly asking to review the PR since you are familiar with QNNs very well. Thanks.

qiskit_machine_learning/neural_networks/sampling_qnn.py
```diff
@@ -0,0 +1,177 @@
...
+        interpret: Callable[[int], int | tuple[int, ...]] | None = None,
+    ):
+        """
+        Args:
+            circuit: The parametrized quantum circuit that generates the samples of this network.
+            sampler: The sampler primitive used to compute the neural network's results.
+                If ``None`` is given, a default instance of the reference sampler defined
+                by :class:`~qiskit.primitives.Sampler` will be used.
+            num_shots: A positive number of shots for each circuit to sample from the
+                quasi-distribution generated by the sampler. The default is 1024.
```
woodsp-ibm, line 85: Since a Sampler, which this takes, has shots would it be helpful to explain the difference to this, and how would I normally set them say wrt to one another if I need to care about this.   

I wonder whether SamplerQNN and SamplingQNN are too similar in name and will lead to confusion - having said that I have no idea for anything different for the latter.

adekusar-drl: This is tricky. In general, they both have the same meaning, but they are interpreted differently. Shots in the Sampler is used to generate distributions. In the QNN shots (or `num_shots`) is used to generate samples (circuit measurements/outcomes?) from the generated distributions. Basically, the QNN does the reverse operation of the Sampler since we don't have direct access to the circuit execution results. The problem is that `shots` has a proper meaning in quantum computing. Same for `samples`, it has a proper meaning of the number of samples in a dataset. In fact, the QNN can re-use the value set to `shots` if present. So, it is confusing, I agree. I'm happy to rename `num_shots` to somewhat else, but I have no idea too. And I don't like the name of the QNN as well.

Of course, I can document the difference between these two parameters/meanings in more details. Any ideas on the names are really welcome.

ikkoham: For naming of this parameter, `num_samples` might be nice.
BTW, since `QuasiDistribution` sometimes has `shots`. It would be nice to have the option of using this `shots` to generate the sequence of outputs. In this case, I would suggest shuffling outcomes instead of bootstrap (resampling).

a-matsuo:
Do we really need to create another new class for "sampling" functionality? We can make a `sampling: bool` option to `SamplerQNN`, right?

qiskit_machine_learning/neural_networks/sampling_qnn.py
```diff
@@ -0,0 +1,177 @@
...
+import numpy as np
+from qiskit import QuantumCircuit
+from qiskit.circuit import Parameter
+from qiskit.primitives import BaseSampler, SamplerResult
+from qiskit.utils import algorithm_globals
+
+from .sampler_qnn import SamplerQNN
+
+
+class SamplingQNN(SamplerQNN):
```
ikkoham, line 28: How about having this feature in `SamplerQNN` like `CircuitQNN` using `sampling: bool` option?

ikkoham (commented):
Just today I need this feature. Great.

PR #558, opened 2023-02-03
Add shot noise emulation to FidelityStatevectorKernel

### LINKED ISSUE #557: Add shot noise emulation to the fidelity statevector kernel
### What should we add?

Model shot noise in `FidelityStatevectorKernel` by simulating sampling noise in the fidelity measurement.

declanmillar:
### Summary

Adds shot noise emulation to `FidelityStatevectorKernel`. Closes #557.

### Details and comments

A `shots` argument is added. If `shots` is `None`, the exact fidelity is used. Otherwise, the mean is taken of samples drawn from a binomial distribution with probability equal to the exact fidelity.

### DIFF
--- a/qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
+++ b/qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
@@ -69,7 +69,7 @@ class FidelityStatevectorKernel(BaseKernel):
         cache_size: int | None = None,
         auto_clear_cache: bool = True,
         shots: int | None = None,
-        enforce_psd: bool = False,
+        enforce_psd: bool = True,
     ) -> None:
         """
         Args:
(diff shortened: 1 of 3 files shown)


woodsp-ibm:
Do we end up needing the `enforce_psd` option back again for if shots is used?

declanmillar:
> Do we end up needing the `enforce_psd` option back again for if shots is used?
Good point; I'll add that back in.

qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
```diff
@@ -65,14 +84,21 @@ def __init__(
             cache_size: Maximum size of the statevector cache. When ``None`` this is unbounded.
             auto_clear_cache: Determines whether the statevector cache is retained when
                 :meth:`evaluate` is called. The cache is automatically cleared by default.
+            shots: The number of shots. If ``None``, the exact fidelity is used. Otherwise, the
+                mean is taken of samples drawn from a binomial distribution with probability equal
+                to the exact fidelity.
+            enforce_psd: Project to the closest positive semidefinite matrix if ``x = y``.
+                This is only required if the number of shots is specified.
```
woodsp-ibm, line 91: Since enforce_psd has a default value, can we say more like  `This is only used when number of shots given is not None.`

adekusar-drl (approved):
Looks good to me, thanks

PR #563, opened 2023-02-13
Fix SamplerQNN when no input parameters and no weights are set in the quantum circuit

adekusar-drl:
### Summary
`SamplerQNN` can now correctly handle quantum circuits without both input parameters and weights. If such a circuit is passed to the QNN then this circuit executed once in the forward pass and backward returns ``None`` for both gradients.

### Details and comments
Changes:
- Some fixes in SamplerQNN to allow circuits without inputs and/or weights.
- Some tweaks in the backward pass of EstimatorQNN to make it aligned with SamplerQNN.
- Removed unnecessary normalization in SamplerQNN.
- Some other minor tweaks in the QNNs.

### DIFF
--- a/qiskit_machine_learning/neural_networks/neural_network.py
+++ b/qiskit_machine_learning/neural_networks/neural_network.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2020, 2022.
+# (C) Copyright IBM 2020, 2023.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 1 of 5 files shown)


qiskit_machine_learning/neural_networks/neural_network.py
```diff
@@ -164,7 +164,8 @@ def _preprocess_forward(
                 num_samples = 1
                 parameters = np.broadcast_to(weights, (num_samples, len(weights)))
             else:
-                return None, None
+                # no input, no weights, just execute circuit once
+                return np.asarray([]), 1
```
woodsp-ibm, line 168: Pretty minor comment: I guess having a return was explicit here before, but this is just one case out of the 4 of the if/else logic, where the others set things up to use the return at the end. Maybe it would be nice to do the same here and set num_samples = 1 and parameters = np.asarray)[]) and use the common return at the end (in this case num_samples = 1 applies to the whole else here i.e. where input_data is None)

adekusar-drl: It makes sense, updated.

woodsp-ibm (approved):

PR #571, opened 2023-03-09
Enable sparse support in TorchConnector and other minor updates

adekusar-drl:
### Summary
The PyTorch connector `TorchConnector` supports sparse output in both passes. That means a hybrid network that has a quantum layer must rely on PyTorch sparse operations only.

### Details:
Added more tests for the connector to `test_torch_connector.py`

~~**On hold** until we drop support of python 3.7 since the code depends on the newer version of `sparse` that does not support 3.7 anymore.~~

Unit tests don't test sparse configuration on python 3.7.


adekusar-drl:
Coverage went down on python 3.7 which is not supported in sparse test cases.
@woodsp-ibm Can you please take a look at the PR?

qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -218,13 +277,29 @@ def __init__(
...
+                3.8 or higher.
+
+        Raises:
+            QiskitMachineLearningError: If the connector is configured as sparse and the underlying
+                network is not sparse. Or if python version is 3.7.
         """
         super().__init__()
         self._neural_network = neural_network
+        if sparse is None:
+            sparse = False
```
woodsp-ibm, line 290: In looking at `sparse` it says this
>If sparse is set to None, then the setting from the given neural network is used.

This here is forcing it to False though. Was wondering why it defaulted to None in init and then if None gets set to False here that the default would not just be False rather than None.

This adopting the sparse setting from the network seems no longer applicable as per the exception below

>"TorchConnector configured as sparse, the network must be sparse as well"

woodsp-ibm:
> Unit tests don't test sparse configuration on python 3.7.

We could check they raise the exception as expected, which is all that can be done, to assert its behaving as expected in this case (instead of skipping the tests completely)

qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -94,15 +100,30 @@ def forward(  # type: ignore
...
-                from sparse import SparseArray, COO
+            if ctx.sparse:
+                if neural_network.sparse:
+                    _optionals.HAS_SPARSE.require_now("SparseArray")
+                    # pylint: disable=import-error
+                    from sparse import SparseArray, COO
 
-                result = cast(COO, cast(SparseArray, result).asformat("coo"))
-                result_tensor = sparse_coo_tensor(result.coords, result.data)
+                    # todo: replace output type from DOK to COO?
```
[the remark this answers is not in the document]
adekusar-drl: It is a long term comment. I'm not sure it is worth replacing DOK to COO, but at least it is somewhat to consider in future updates/changes.

qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -133,8 +154,8 @@ def backward(ctx: Any, grad_output: Tensor) -> Tuple:  # type: ignore
             neural_network = ctx.neural_network
 
             # if sparse output is requested return None, since PyTorch does not support it yet.
-            if neural_network.sparse and ctx.sparse:
-                return None, None, None, None
+            # if neural_network.sparse and ctx.sparse:
+            #     return None, None, None, None
```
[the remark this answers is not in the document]
adekusar-drl: I forgot to remove the commented out code. Thanks.

qiskit_machine_learning/connectors/torch_connector.py
adekusar-drl: Honestly, I missed the docstring. Updated `sparse` handling and the docstring. Should be better now.

releasenotes/notes/add-sparse-torch-connector-a3b9e3d50b405a01.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    The PyTorch connector :class:`~qiskit_machine_learning.connector.TorchConnector` supports
+    sparse output in both passes. That means a hybrid network that has a quantum layer must rely
+    on PyTorch sparse operations only.
```
woodsp-ibm, line 6: Do you think this enough for the release note for this change? Would an example improve it? 

What does this mean for an end user seeing this.

>That means a hybrid network that has a quantum layer must rely on PyTorch sparse operations only.

If I have a regular PyTorch network, without quantum, is this still a true statement. I am trying to figure as an end user who is going to create a hybrid network what this means to me say as different from a pure classical set that I might have used before.

adekusar-drl: Updated the reno: added an example and rephrased this limitation. Hopefully, it is better now.

releasenotes/notes/add-sparse-torch-connector-a3b9e3d50b405a01.yaml
```diff
@@ -0,0 +1,46 @@
...
+features:
+  - |
+    The PyTorch connector :class:`~qiskit_machine_learning.connector.TorchConnector` now fully
+    supports sparse output in both forward and backward passes. To enable sparse support, the
+    `sparse` parameter must be set in the connector itself along with the corresponding `sparse`
+    parameter of :class:`~qiskit_machine_learning.neural_networks.SamplerQNN`. This snippet
+    illustrates how to create a sparse instance of the connector.
+
+    .. code-block:: python
+      import torch
```
woodsp-ibm, line 11:
```suggestion

      import torch
```

releasenotes/notes/add-sparse-torch-connector-a3b9e3d50b405a01.yaml
```diff
@@ -0,0 +1,46 @@
+---
+features:
+  - |
+    The PyTorch connector :class:`~qiskit_machine_learning.connector.TorchConnector` now fully
+    supports sparse output in both forward and backward passes. To enable sparse support, the
+    `sparse` parameter must be set in the connector itself along with the corresponding `sparse`
+    parameter of :class:`~qiskit_machine_learning.neural_networks.SamplerQNN`. This snippet
```
woodsp-ibm, line 7: > must be set in

Since it picks up the value of the neural network anyway by default its not needed if you set in on the QNN right? Do these always have to match? In which case could we consider deprecating and dropping it there or is there some valid mix/match where conversion is done?

> corresponding `sparse`  parameter of :class:`~qiskit_machine_learning.neural_networks.SamplerQNN`.

Should this be a more general statement about a QNN class rather than SamplerQNN which is more an example - though at present I guess its really the only one we have, since EstimatorQNN does not allow sparse to be set.

adekusar-drl: I tried to be more precise in the reno. The connector can be dense and the network is sparse, this is the only case when they don't match each other. 

No, there's no need to deprecate and drop.

And yes, there's only `SamplerQNN` that supports sparsity.

woodsp-ibm (approved):

PR #573, opened 2023-03-15
Remove unnecessary `html_theme_path` option

Eric-Arellano:
See https://github.com/Qiskit/qiskit-finance/pull/244 for the motivation.

I built the docs locally on `main` vs this branch and could find no differences between the two, as expected.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -88,13 +88,6 @@ nbsphinx_prolog += link_str + "{{ docname }}"
 
 # -- General configuration ---------------------------------------------------
 
-# If your documentation needs a minimal Sphinx version, state it here.
-#
-# needs_sphinx = '1.0'
-
-# Add any Sphinx extension module names here, as strings. They can be
-# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
-# ones.
 extensions = [
     "sphinx.ext.napoleon",
     "sphinx.ext.autodoc",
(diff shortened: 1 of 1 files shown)


adekusar-drl (commented):
Seems like the PR introduced the same problem as in optimization here: https://github.com/Qiskit/qiskit-optimization/pull/475. Spell checker fails on the class names in the warning messages. I'm not sure I want to add all that class names to the dictionary.

woodsp-ibm:
Since #571 passed yesterday I am re-running that to see if the failure might be related to a change in a library or dependency. I tried looking at versions but the spelling checker etc seem the same versions.

woodsp-ibm:
It has to be something that changed very recently but not what was done by this PR. The sparse TorchConnector PR #571 that passed yesterday ok, fails today the same way as this PR now having just re-run it.

adekusar-drl:
Oh, I see now. Then you are right, something has been changed very recently.

woodsp-ibm:
It seems to be the class names at the end of the long class paths in the deprecation messages that it did pull off and check before. Not sure why now. Would any alteration to the message , say putting the whole path say in singles/double quotes stop it as a alternative to adding it to spell checker - something we'd have to try I guess if we don't want to add those class names to the dict.

woodsp-ibm:
I was wondering why Nature did not hit this since it has a lot of deprecations. With QuantumKernel here that is failing the class names are directly in the message (string) here versus in params of a decorator. Maybe both can go in the new name string parameter of a decorator rather than as currently done as a different option.

adekusar-drl:
At the first glance, I have not spotted the difference between two setups: when CI failed and when CI checks were successful. Looks like the library versions are identical.

woodsp-ibm:
Now the failure here with the spelling in QuantumKernel its using a Terra decorator - that was just changed and now includes the deprecation message into the doctring which the spelling ends up seeing. So its not its presence in the source I think more that it gets injected into the docstring. You can see the result in the html built
image

adekusar-drl:
Perhaps, this commit https://github.com/Qiskit/qiskit-terra/pull/9790 introduced that behavior. I will check it out later.

woodsp-ibm:
Yes, Eric has been working on improving the decorators. The problem in Optimization is related but a bit different. Hopefully we can figure what the best way forward is given that it was recognized it would be very helpful for the deprecation info to be in the docs too for an end user and for it to be automated,

woodsp-ibm (approved):

PR #574, opened 2023-03-17
Fix spelling in deprecated functions/classes

adekusar-drl:
### Summary
There are several ways to fix the problem appeared in #573:

1. Remove packages and leave only class names in the deprecated messages. I don't like it, fully qualified names are better in this case.
2. Add class names to the dictionary. Not great though, but the easiest way to fix the problem, this PR does it.
3. Format class names with :class:`class_name`. Convenient for html documentation, but inconvenient for developers, they would see the prefix, etc
4. Remove lowercase filter. Solves the issue, but breaks down the dictionary, it becomes case sensitive.
5. Add another filter to sphinx or hack it in another way.

I prefer 2.

### DIFF
--- a/.pylintdict
+++ b/.pylintdict
@@ -70,11 +70,13 @@ dt
 eigenstates
 endian
 entangler
+estimatorqnn
 et
 eval
 expressibility
 farrokh
 fidelities
+fidelityquantumkernel
 formatter
 frontend
 func
@@ -161,6 +163,7 @@ ovo
 ovr
 param
 parameterization
+parametrize
 parametrized
 params
 pauli
@@ -206,6 +209,7 @@ runtimes
 rx
 ry
 rz
+samplerqnn
 scalability
 scikit
 scipy
@@ -240,6 +244,7 @@ toctree
 todo
 traceback
 trainability
+trainablefidelityquantumkernel
 transpilation
 transpile
 transpiled
--- a/qiskit_machine_learning/kernels/quantum_kernel.py
+++ b/qiskit_machine_learning/kernels/quantum_kernel.py
@@ -67,7 +67,6 @@ class QuantumKernel(TrainableKernel, BaseKernel):
         "removed after that.",
         stacklevel=3,
         category=PendingDeprecationWarning,
-        since="0.5.0",
     )
     def __init__(
         self,
--- a/qiskit_machine_learning/kernels/quantum_kernel.py
+++ b/qiskit_machine_learning/kernels/quantum_kernel.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021, 2022.
+# (C) Copyright IBM 2021, 2023.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
@@ -67,6 +67,7 @@ class QuantumKernel(TrainableKernel, BaseKernel):
         "removed after that.",
         stacklevel=3,
         category=PendingDeprecationWarning,
+        since="0.5.0",
     )
     def __init__(
         self,
--- a/qiskit_machine_learning/kernels/quantum_kernel.py
+++ b/qiskit_machine_learning/kernels/quantum_kernel.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021, 2023.
+# (C) Copyright IBM 2021, 2022.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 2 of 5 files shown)


woodsp-ibm:
I would like to see what @Eric-Arellano has to say in regards of the problem. The decorator takes that message and injects it into the doctring. The class names that are put in there could be manipulated at that time to improve things for a user so these are links. Of course one needs to detect these in some way so whatever is done is reliable. - there is also a different outcome in Optimization where the deprecation of the parent class is injected into a subclass - ended up with the same spelling problem but seeing the parents deprecation message there is arguably confusing. A spelling fix was already merged in Optimization - arguably we can do that here too since having them in the dict is not a real issue. Going forwards with this where deprecation messages will exist for all deprecation messages I would like to consider whether things can be improved, since this change has just been done, before the use is more pervasive. Note the change is only in main so far in Terra so the built docs and spelling in stable would not have this issue.

woodsp-ibm:
>Deprecation authors are welcome to start writing :class: in their deprecation message. It will show that as raw text during the runtime warning & calling help(), so a little weird. But possible.

Now we could do the opposite - put the :class: and :meth: etc in the docstring where that pattern would be detectable enough and could be removed when raising the error. This would allow docs with links in them but not expose this odd syntax in the message of the warning that is finally raised.

Eric-Arellano:
> Now we could do the opposite - put the :class: and :meth: etc in the string where that pattern would be detectable enough and could be removed when raising the error. This would allow docs with links in them but not expose this odd syntax in the message of the warning that is finally raised.

Agreed, that is 100% reasonable for us to process the deprecation message before we call `warnings.warn()`. `help()` would still show the `:class:` syntax, but that seems okay; it's the runtime warning I'm more worried about.

I will work on this today.

Eric-Arellano:
See https://github.com/Qiskit/qiskit-terra/pull/9815

woodsp-ibm:
> There are several ways to fix the problem appeared in

I could add a 6th

6. Switch to using the deprecate decorators that we have in ML here in the couple of places, that for some reason, we are using the Terra ones.

This would make the docs independent of the change to Terra, that is just affecting up now in a couple of places here. We can decide later to transition to where these messages are appearing consistently throughout here. Whether moving to use the decorators in Terra and/or updating the local decorators to also inject the text into the docstring.

Adding the names to the dictionary, as is done here, lets the code stay as is. I'm fine with that too. With the more recent change that has been done as a PR in Terra to have allow :class: etc in the message so the docs come out nice, but remove it from the message in the warning so we may want also to update some messages.

That docs update will only affect that built against the next version of Terra - which is what is done in main here. Over in stable, the docs are built against stable Terra and will not be impacted by this for now, until that gets released. Which will be after the next release of ML which should be sooner.

adekusar-drl:
Let me add my two cents. Since we were celebrating St. Patrick last Friday I did not follow the discussion.
1. I'm in favor of using Terra decorations, they are maintained centrally and they should be universal, I think.
2. I don't like adding class names as words to the dictionary. It is already large.
3. If we could have nice warnings and formatted documentation, it would be very nice.
4. We can't use recent changes in Terra due to `neko` limitations.

So, I think I'll keep the words in the dictionary, keep plain text in the warning, and after the next Terra release we can rollback the dictionary and make use of formatted messages in deprecation decorators. For such purposes, I'll create an issue if you agree with the proposed solution.

woodsp-ibm:
>  I'll create an issue if you agree with the proposed solution.

Sure its fine by me. The reason we have the decorators here (and elsewhere) in the apps is they they went beyond what was in Terra function wise and in standardizing messages etc. Eric is now improving the decorators in Terra with a similar goal though also to have the text become part of the docs - which having messages standarized is arguably even more important. Since we more or less have the same decorators across the apps repos it will be good to be able to use/maintain just the one common set. Already here though a bunch of decorators are done from the ones in ML, hence to me it was a choice to use these throughout to avoid this problem. I am not sure how we ended up using the Terra ones in a couple of places here in ML. Anyway I would see the apps migrating over time to the new Terra decorators, rather than us updating them here to inject docs etc. so as just to be using/maintaining one set where possible. What exactly this entails I guess we need to work out.

I was reacting to your not wanting the class names that got impacted in the custom dict. In the iterim we can solve it either way - use the decorators from here in ML and avoid the issue or do what you are proposing and leave the decorators using Terra and add to dict. This latter way will not affect the stable branch at present, since that uses Terra stable, but once Terra releases with the new decorators and becomes the new stable then any new bugs fix releases off that branch will affect the docs and the deprecation messages as they are coming out now will do so there. Normally we do not do code updates for anything else other than critical bugs - though we do doc fixes. Here this would be an intersection - maybe we'll just live with the docs coming out with the deprecations in a couple of places until the subsequent release where things can be addressed.

woodsp-ibm (approved):

PR #578, opened 2023-03-21
Fix docstrings in VQR

### LINKED ISSUE #577: VQR `estimator` parameter indicates incorrect default
>estimator (BaseEstimator | None) – An estimator to be used to evaluate expectation values of the observable. If None the [qiskit.primitives.BaseEstimator](https://qiskit.org/documentation/stubs/qiskit.primitives.BaseEstimator.html#qiskit.primitives.BaseEstimator) is used. The underlying QNN is [EstimatorQNN](https://qiskit.org/documentation/machine-learning/stubs/qiskit_machine_learning.neural_networks.EstimatorQNN.html#qiskit_machine_learning.neural_networks.EstimatorQNN).

This should be the `Estimator` class there not the abstract parent base class `BaseEstimator` type

adekusar-drl:
### Summary
Fixes #577 .

### Details and comments
Now the docstring states:
```
estimator: An estimator to be used to evaluate expectation values of the observable.
    If ``None`` the :class:`qiskit.primitives.Estimator` is used. If an estimator
    instance is set, the underlying QNN will be of type
    :class:`~qiskit_machine_learning.neural_networks.EstimatorQNN`, and the estimator
    primitive will be used to compute the neural network's results.
```

### DIFF
--- a/qiskit_machine_learning/algorithms/regressors/vqr.py
+++ b/qiskit_machine_learning/algorithms/regressors/vqr.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021, 2022.
+# (C) Copyright IBM 2021, 2023.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
--- a/qiskit_machine_learning/algorithms/regressors/vqr.py
+++ b/qiskit_machine_learning/algorithms/regressors/vqr.py
@@ -89,7 +89,7 @@ class VQR(NeuralNetworkRegressor):
                 optimized. This allows to track how well optimization / training process is going on.
             estimator: An estimator to be used to evaluate expectation values of the observable.
                 If ``None`` and quantum instance is not set, the
-                :class:`qiskit.primitives.Estimator` is used, the underlying QNN will be of type
+                :class:`qiskit.primitives.Estimator` is used. The underlying QNN will be of type
                 :class:`~qiskit_machine_learning.neural_networks.EstimatorQNN`, and the estimator
                 primitive will be used to compute the neural network's results.
         Raises:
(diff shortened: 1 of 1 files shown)


qiskit_machine_learning/algorithms/regressors/vqr.py
```diff
@@ -88,9 +88,8 @@ def __init__(
                 as an array and a computed value as a float of the objective function being
                 optimized. This allows to track how well optimization / training process is going on.
             estimator: An estimator to be used to evaluate expectation values of the observable.
-                If ``None`` the :class:`qiskit.primitives.BaseEstimator` is used. The underlying QNN
-                is :class:`~qiskit_machine_learning.neural_networks.EstimatorQNN`.
-            If an estimator instance is set, the underlying QNN will be of type
+                If ``None`` the :class:`qiskit.primitives.Estimator` is used. If an estimator
+                instance is set, the underlying QNN will be of type
```
woodsp-ibm, line 92: I noticed the BaseEstimator when answering a question, but in reading this again I note that we say
> If an estimator instance is set, the underlying QNN will be of type....

but arguably the underlying type is always EstimatorQNN, whether an `estimator is instance is set` or not - the default estimator stated here is that of that class (unlikely to change I guess). Maybe we can drop the bit about an instance being set?

adekusar-drl: What about this?
```
            estimator: An estimator to be used to evaluate expectation values of the observable.
                If ``None`` and quantum instance is not set, the 
                :class:`qiskit.primitives.Estimator` is used, the underlying QNN will be of type
                :class:`~qiskit_machine_learning.neural_networks.EstimatorQNN`, and the estimator
                primitive will be used to compute the neural network's results.
```

woodsp-ibm: I might do a full stop not a comma, after `is used` thus

> If ``None`` and quantum instance is not set, the  :class:`qiskit.primitives.Estimator` is used. 
> The underlying QNN will be of type....

so as it does not seem the latter part is simply qualifying that initial text so it still only applies in the None case.

adekusar-drl: Updated.

woodsp-ibm (approved):

woodsp-ibm (approved):

PR #585, opened 2023-03-27
Fix Terra installation in CI

adekusar-drl:


adekusar-drl:
There's no need for this fix.

PR #598, opened 2023-04-06
Remove qgan and runtime packages as deprecated

adekusar-drl:
### Summary
This PR removes `qgan` and `runtime` packages from the sources, tests and documentation. The packages are deprecated as of version 0.5.


releasenotes/notes/remove-qgan-runtime-aa74bec5d95ffa00.yaml
```diff
@@ -0,0 +1,8 @@
+---
+
+upgrade:
+  - |
+    The previously deprecated `qgan` and `runtime` packages have been removed. Please refer to
+    the new QGAN tutorial and new primitive based quantum neural networks such as
+    :class:`~qiskit_machine_learning.neural_networks.EstimatorQNN` and
+    :class:`~qiskit_machine_learning.neural_networks.SamplerQNN` respectively.
```
woodsp-ibm, line 8: Should we mention `Torch runtime` here explicitly. I guess both replacements are more or less using the TorchConnector with the new primitive based neural nets.

adekusar-drl: I have updated the reno, should be better now.

PR #603, opened 2023-04-14
Replace sparse with sparse_output in the OneHotEncoder constructor

adekusar-drl:
### Summary
This PR replaces usage of `sparse` with `sparse_output` where `OneHotEncoder` is instantiated.


woodsp-ibm:
We should update the requirements txt too, I imagine, as I see that change is effective from 1.2 but we allow older sklearn versions 

`scikit-learn>=0.20.0`

so given the change we now need 1.2 at a minimum it seems from sklearn:

> New in version 1.2: sparse was renamed to sparse_output
> Deprecated since version 1.2: sparse is deprecated in 1.2 and will be removed in 1.4. Use sparse_output instead.

adekusar-drl:
Agree. Updated.

woodsp-ibm (approved):

woodsp-ibm:
3.7 support in Qiskit ends with Terra 0.25 at end July. Place this on hold until such time as 3.7 support is removed from here commensurate with that?

adekusar-drl:
On hold until we drop support of python 3.7

PR #606, opened 2023-04-15
Fix complex kernel matrix entries for FidelityStatevectorKernel

### LINKED ISSUE #605: FidelityStatevectorKernel can return a complex-valued matrix
### Environment

- **Qiskit Machine Learning version**: 9b1c56e3
- **Python version**: Python 3.10.11
- **Operating system**: macOS 13.3.1

### What is happening?

FidelityStatevectorKernel can return a complex-valued matrix when enforcing a PSD matrix.

With an appropriately large dataset and with `shots!=None` tiny non-zero complex values will populate the kernel matrix.

Snippet:
```python
kernel = FidelityStatevectorKernel(enforce_psd=True, shots=100)
```

### What should happen?

The matrix should be real-valued, with float-type dtype.

### Any suggestions?

Just take the real part of the kernel array.

declanmillar:
### Summary

Takes the real part of the kernel array when enforcing a PSD matrix. 

Closes #605.

### DIFF
--- a/qiskit_machine_learning/kernels/fidelity_quantum_kernel.py
+++ b/qiskit_machine_learning/kernels/fidelity_quantum_kernel.py
@@ -13,6 +13,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Sequence
 from typing import List, Tuple
 
 import numpy as np
@@ -114,9 +115,6 @@ class FidelityQuantumKernel(BaseKernel):
         if is_symmetric and self._enforce_psd:
             kernel_matrix = self._make_psd(kernel_matrix)
 
-        # due to truncation and rounding errors we may get complex numbers
-        kernel_matrix = np.real(kernel_matrix)
-
         return kernel_matrix
 
     def _get_parameterization(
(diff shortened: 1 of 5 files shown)


adekusar-drl (approved):

woodsp-ibm:
> where kernel entries could potentially have nonzero complex components due to truncation and rounding errors when enforcing a PSD matrix.

As the psd method is used by other kernels would it be applicable for the fix to be in the method itself, rather than external here for it's one use here and only applies to this specific kernel?

qiskit_machine_learning/kernels/fidelity_quantum_kernel.py
```diff
@@ -215,7 +215,7 @@ def _get_kernel_entries(self, left_parameters: np.ndarray, right_parameters: np.
                 left_parameters,
                 right_parameters,
             )
-            kernel_entries = np.real(job.result().fidelities)
+            kernel_entries = job.result().fidelities
```
declanmillar, line 218: I've removed this use of `np.real()`. Behind the scenes I think this was effectively just using `np.asanyarray`, as the fidelities should be a sequence of floats. These entries are later added to a np.ndarray with np.float64 dtype anyway.

qiskit_machine_learning/kernels/fidelity_quantum_kernel.py
```diff
@@ -114,9 +115,6 @@ def evaluate(self, x_vec: np.ndarray, y_vec: np.ndarray | None = None) -> np.nda
         if is_symmetric and self._enforce_psd:
             kernel_matrix = self._make_psd(kernel_matrix)
 
-        # due to truncation and rounding errors we may get complex numbers
-        kernel_matrix = np.real(kernel_matrix)
-
```
declanmillar, line 119: I believe the only possible source of these errors is `_make_psd`, so this may also be removed.

adekusar-drl (approved):

PR #609, opened 2023-04-25
Simplify SamplerQNN and EstimatorQNN interfaces with QNNCircuit

### LINKED ISSUE #597: Enhance SamplerQNN and EstimatorQNN with QNNCircuit
### What should we add?

A newly introduced `QNNCircuit` should simplify interfaces of QNNs. Right now to create an QNN instance users have to pass at least three parameters: a circuit itself, a list of input parameters, and a list of trainable parameters or weights. With the new `QNNCircuit` the task should become easier to accomplish. Since the qnn circuit keeps track of feature map and ansatz, all circuit parameters are known as well. Therefore it is enough to pass an instance of `QNNCircuit` to a quantum neural network without passing lists of parameters. The networks should figure out the required lists of parameters themselves.

A user code should look like this:

```python
fm = ...
ansatz = ...
qc = QNNCircuit(feature_map=fm, ansatz=ansatz)
qnn = SamplerQNN(qc)
```

david-alber:
### Summary

The ``QNNCircuit`` class can be passed as a circuit to ``SamplerQNN`` and ``EstimatorQNN``.
This simplifies the interfaces to build a Sampler/Estimator-based neural network implementation from a feature map and an ansatz circuit.

Closes #597 

### Details and comments

Using the ``QNNCircuit`` for the construction of ``SamplerQNN`` and ``EstimatorQNN`` comes with two benefits:

1)  The explicit composition of a feature map and ansatz circuit can be omitted.

2)  the input and weight parameters, previously required inputs, can be omitted (since these properties are already held in the ``QNNCircuit``internally).


qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -45,19 +46,38 @@ class EstimatorQNN(NeuralNetwork):
...
         from qiskit.circuit.library import ZZFeatureMap, RealAmplitudes
+        from qiskit_machine_learning.circuit.library import QNNCircuit
 
         from qiskit_machine_learning.neural_networks import EstimatorQNN
 
         num_qubits = 2
+
+        # Using the QNNCircuit:
+        # Create a parameterized 2 qubit circuit composed of the default ZZFeatureMap feature map
+        # and RealAmplitudes ansatz with 3 repetitions.
```
woodsp-ibm, line 69: The 3 reps here is the default value that RealAmplitudes has right. So its `a default ZZFeatureMap feature map with a default RealAmplitudes ansatz` - the latter has 3 reps by default. Instead of explicitly configuring the number or reps below, as was originally done, if we want it to match it might be better to remove the explicit setting and let it just set to the same default. It could be noted that this is 3 reps here if thats important.

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -98,13 +118,20 @@ def __init__(
...
+            circuit: The quantum circuit to represent the neural network. If a
+                :class:`~qiskit_machine_learning.circuit.library.QNNCircuit` is passed,
+                input_params and weight_params are omitted. In this case these two properties are
+                held in the :class:`~qiskit_machine_learning.circuit.library.QNNCircuit`.
             observables: The observables for outputs of the neural network. If ``None``,
                 use the default :math:`Z^{\otimes num\_qubits}` observable.
             input_params: The parameters that correspond to the input data of the network.
                 If ``None``, the input data is not bound to any parameters.
+                If a :class:`~qiskit_machine_learning.circuit.library.QNNCircuit` is passed as
+                circuit, this input is omitted.
```
woodsp-ibm, line 130: Saying it another way - maybe we can say something more like this.

`If a QNNCircuit is provided the input_params value here is ignored and the value instead is taken from the QNNCircuit input_parameters.` where I used ignored rather than omitted.

Similar for weight_params. And maybe the text for circuit can be updated - something similar is said there

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -85,7 +105,7 @@ class EstimatorQNN(NeuralNetwork):
     def __init__(
         self,
         *,
-        circuit: QuantumCircuit,
+        circuit: QuantumCircuit | QNNCircuit,
```
woodsp-ibm, line 108: Since a QNNCircuit is a QuantumCircuit its already really covered by the QuantumCircuit typehint. Yes we recognize this specific subclass and do things differently...

qiskit_machine_learning/neural_networks/estimator_qnn.py
david-alber: Mentioning the default reps for RealAmplitudes (reps=3) is not of particular importance. It was only meant to give the reader a hint why the length of the weights array is 8 😄 . Since the default reps of ZZFeatureMap (reps=2) are also not mentioned, let's ignore it.

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -98,13 +118,23 @@ def __init__(
...
+                :class:`~qiskit_machine_learning.circuit.library.QNNCircuit` is passed, the
+                input_params and weight_params do not have to be provided, because these two
+                properties are taken from the
+                :class:`~qiskit_machine_learning.circuit.library.QNNCircuit`.
             observables: The observables for outputs of the neural network. If ``None``,
                 use the default :math:`Z^{\otimes num\_qubits}` observable.
             input_params: The parameters that correspond to the input data of the network.
                 If ``None``, the input data is not bound to any parameters.
+                If a :class:`~qiskit_machine_learning.circuit.library.QNNCircuit` is provided the
+                input_params value here is ignored. Instead the value is taken from the
```
adekusar-drl, line 131:
```suggestion
                `input_params` value here is ignored. Instead the value is taken from the
```

qiskit_machine_learning/neural_networks/estimator_qnn.py
```diff
@@ -98,13 +118,23 @@ def __init__(
...
+                :class:`~qiskit_machine_learning.circuit.library.QNNCircuit` input_parameters.
             weight_params: The parameters that correspond to the trainable weights.
                 If ``None``, the weights are not bound to any parameters.
+                If a :class:`~qiskit_machine_learning.circuit.library.QNNCircuit` is provided the
+                weight_params value here is ignored. Instead the value is taken from the
```
adekusar-drl, line 136:
```suggestion
                `weight_params` value here is ignored. Instead the value is taken from the
```

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -121,16 +141,26 @@ def __init__(
                 If ``None`` is given, a default instance of the reference sampler defined
                 by :class:`~qiskit.primitives.Sampler` will be used.
             circuit: The parametrized quantum circuit that generates the samples of this network.
-            input_params: The parameters of the circuit corresponding to the input.
-            weight_params: The parameters of the circuit corresponding to the trainable weights.
+                If a :class:`~qiskit_machine_learning.circuit.library.QNNCircuit` is passed, the
+                input_params and weight_params do not have to be provided, because these two
```
adekusar-drl, line 145:
```suggestion
                `input_params` and `weight_params` do not have to be provided, because these two
```

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -121,16 +141,26 @@ def __init__(
...
+                properties are taken from the
+                :class:`~qiskit_machine_learning.circuit.library.QNNCircuit`.
+            input_params: The parameters of the circuit corresponding to the input. If a
+                :class:`~qiskit_machine_learning.circuit.library.QNNCircuit` is provided the
+                input_params value here is ignored. Instead the value is taken from the
```
adekusar-drl, line 150:
```suggestion
                `input_params` value here is ignored. Instead the value is taken from the
```

qiskit_machine_learning/neural_networks/sampler_qnn.py
```diff
@@ -121,16 +141,26 @@ def __init__(
...
+                :class:`~qiskit_machine_learning.circuit.library.QNNCircuit` input_parameters.
+            weight_params: The parameters of the circuit corresponding to the trainable weights. If
+                a :class:`~qiskit_machine_learning.circuit.library.QNNCircuit` is provided the
+                weight_params value here is ignored. Instead the value is taken from the
```
adekusar-drl, line 154:
```suggestion
                `weight_params` value here is ignored. Instead the value is taken from the
```

adekusar-drl (commented):
Thanks. I have only formatting comments.

david-alber:
> Thanks. I have only formatting comments.

Yes, very legit! I committed your suggestions 👍

adekusar-drl (approved):

PR #619, opened 2023-05-04
Fix language toggle in docs

Eric-Arellano:
### Summary

The language toggle was broken for two reasons:

1. Sphinx 6 does not activate jquery by default anymore, so the interactivity was broken
2. The `versionutils.py` file should be setting `language_label`, not `version_label`

This PR also makes some other docs config improvements:

* Pin the sphinx theme version for more stability
* Remove override of the `page.html` and `versions.html` templates. We should use qiskit-sphinx-theme instead.


woodsp-ibm (approved):
Looks good to me. Unlike the equivalent in Finance Qiskit/qiskit-finance#256 you changed the name of the method here in versionutils as well - just noting this that's all.

Eric-Arellano:
> Unlike the equivalent in Finance https://github.com/Qiskit/qiskit-finance/pull/256 you changed the name of the method here in versionutils as well - just noting this that's all.

Indeed, good eye. I should have updated it in Finance and didn't realize. I didn't go back to fix it since I plan to delete `versionutils.py` in ~1-2 weeks with the theme 1.12 release including versionutils.py for you automatically.

PR #621, opened 2023-05-06
Terra main vs stable install switch

woodsp-ibm:
### Summary

As Terra main has dropped Python 3.7 support (and with it CI here - see last nights scheduled build - using Terra main fails against 3.7) this enables a switch of the install between main or stable and it's set at present not to use main (i,e, pip install latest stable)

See Qiskit/qiskit-nature#1159 for similar

Note: Nature and Opt had already dropped building Aer locally in favor of just simply installing the PyPi release of Aer. I changed that here to be the same since building Aer is complex and not really needed anymore.

I labelled it to be backported just to keep the workflows in sync - stable only installs stable versions anyway (do not use this main dependencies action).

### DIFF
--- a/.github/actions/install-main-dependencies/action.yml
+++ b/.github/actions/install-main-dependencies/action.yml
@@ -19,13 +19,15 @@ inputs:
   python-version:
     description: 'Python version'
     required: true
+  terra-main:
+    description: 'Use Terra main'
+    required: true
 runs:
   using: "composite"
   steps:
     - name: Get main last commit ids
       run: |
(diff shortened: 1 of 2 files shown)


adekusar-drl (approved):
Correct me if I wrong, starting from this PR CI will use a release version of Terra. And how do I change the switch value?

woodsp-ibm:
> Correct me if I wrong, starting from this PR CI will use a release version of Terra. 

Yes

> And how do I change the switch value?

There is a flag in `main.yml` workflow `terra-main: "false"` and that would be set "true" to switch it back to using Terra main branch. Thats the only change in the main.yml file here so ts easy to see where it is - its basically a parameter/flag passed to the main_dependencies action, the other file that was changed here

PR #626, opened 2023-05-12
Remove unused custom CSS files and Sphinx directives

Eric-Arellano:
No one was using the custom directives `galleryitem` and `customgalleryitem`. After removing this, it's safe to remove `sphinx-gallery` because it's not used. That allows removing `gallery.css`, which is meant to override `sphinx-gallery`'s default styling.

This also removes two CSS files:

* `custom.css`: There is no `.toggle` element anymore
* `style.css`: its 3 rules are holdovers from the Read The Docs theme


docs/conf.py
```diff
@@ -208,12 +202,9 @@
 
 html_context = {"analytics_enabled": True}
 
-import sphinx
-
-if sphinx.__version__ >= "4.1.0":
```
Eric-Arellano, line 213: You pin Sphinx to >=5, so this is unnecessary.

woodsp-ibm: I imagine it was just overlooked when sphinx was recently pinned that way following Terra - since it makes no net difference to the outcome. But yes it can be dropped now.

PR #628, opened 2023-05-12
Remove the `includenodoc` and `customcarditem` directives

Eric-Arellano:
According to grep, these weren't actually used. 

`CustomCalloutItemDirective` is used though. It's used in a few other projects too, so I'm likely going to move this to `qiskit_sphinx_theme` in its 1.13 release so that you don't have to copy and paste the code. Tracked by https://github.com/Qiskit/qiskit_sphinx_theme/issues/323.


woodsp-ibm (approved):

woodsp-ibm:
@Mergifyio backport stable/0.6

PR #660, opened 2023-06-09
Improve links and list formatting for tutorials

woodsp-ibm:
### Summary

While changing urls for the repo move in #657 I noted:

>This also changed the urls similarly in `03_quantum_kernel.ipynb`. For 2 notebook references I think these would be better as relative urls so that the result ends up pointing the published html version. And for the `FidelityQuantumKernel` link that would be better to the published API ref than direct to the source. I did not change these rather I just changed all urls likewise and figured this would be better in separate PR to improve the links for it.

This PR changes the links according:
* Notebook urls are relative so if you are looking at source code, running in jupyter or looking at the published html it takes you to the corresponding context in the same folder.
* I also changed some urls to a) point at the new docs location in ecosystem and b) make the link their name rather than the generic class or primitive word next to them.

I also fixed the html formatting as a couple of lists were missing the preceding blank line that Sphinx needs to avoid all the text being just wrapped into the preceding text.

### DIFF
--- a/docs/tutorials/01_neural_networks.ipynb
+++ b/docs/tutorials/01_neural_networks.ipynb
@@ -59,9 +59,9 @@
     "\n",
     "The QNNs in `qiskit-machine-learning` are meant as application-agnostic computational units that can be used for different use cases, and their setup will depend on the application they are needed for. The module contains an interface for the QNNs and two specific implementations:\n",
     "\n",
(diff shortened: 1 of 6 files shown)


woodsp-ibm:
While I noticed the urls in one of the notebooks when searching to replace repo links, I figured I would check for other links to older API refs that are moved to ecosystem. Since this now hits a few more notebooks I changed the title to reflect this.

adekusar-drl (approved):

Issue #613, opened 2023-05-02
Update the Quantum Neural Networks tutorial with QNNCircuit


adekusar-drl:
### What should we add?

With the recent introduction of `QNNCircuit` it is important to show how to use it in the QML tutorials. The most suitable opportunity is to update the `01_neural_networks.ipynb` and add a simplified way of instantiating quantum neural networks. Other tutorials can be updated as well.

david-alber:
@adekusar-drl great idea 👍 !
I'd love to pick up this one 😃 .

athanzxyt:
Hey @adekusar-drl ! Would it be possible for me also to be assigned to this issue? I'd love to help improve the QML tutorials.

adekusar-drl:
@david-alber are there any updates?

david-alber:
@adekusar-drl  I am still working on this issue. We can split the work though: I will update the ``01_neuronal_networks.ipynb`` tutorial. @athanzxyt  if you find any other tutorials where the ``QNNCircuit`` can be applied, please go ahead and do the updates 😀.

adekusar-drl:
@athanzxyt You can think where `QNNCircuit` can be applied else. Thanks.

Jay-Patel-257:
@adekusar-drl Can you reference the API documentation for `QNNCircuit`? I could not find anything related to `QNNCircuit`.

adekusar-drl:
The documentation is not available since it is not released yet.

Jay-Patel-257:
Ahh, alright!

PR #664, opened 2023-06-13
updated QNN tutorials to demonstrate usage of QNNCircuit class

david-alber:
### Summary

The recently introduced `QNNCircuit` class is introduced in the tutorials
- 02_neural_network_classifier_and_regressor.ipynb
- 05_torch_connector.ipynb
- 10_effective_dimension.ipynb
- 11_quantum_convolutional_neural_networks.ipynb
to demonstrate the simplified interface of composing a parameterized quantum circuit from a feature map and an ansatz.

### Details and comments

The QNNCircuit class is intentionally not introduced in the other tutorials (01, 02a, 03, 04, 06, 07, 08, 09, 12) because either one of the below reasons apply: 
- The `QNNCircuit` class would not simplify the code/usage.
- A parameterized Quantum circuit is not composed of a feature map and an ansatz.
- The tutorial intends to demonstrate a usage scenario that would be obscured by utilizing the `QNNCircuit` class.


david-alber:
Closes #613 
(sorry I forgot to put this in the PR msg.)

docs/conf.py
```diff
@@ -94,7 +94,7 @@
 ]
 templates_path = ["_templates"]
 
-nbsphinx_timeout = 360
+nbsphinx_timeout = 600
```
adekusar-drl, line 97: Please don't increase, first you should figure out why the tutorial is timing out.

docs/tutorials/11_quantum_convolutional_neural_networks.ipynb
```diff
@@ -732,37 +733,33 @@
...
-    "circuit.compose(feature_map, range(8), inplace=True)\n",
-    "circuit.compose(ansatz, range(8), inplace=True)\n",
+    "circuit = QNNCircuit(feature_map=feature_map, ansatz=ansatz)\n",
     "\n",
     "observable = SparsePauliOp.from_list([(\"Z\" + \"I\" * 7, 1)])\n",
     "\n",
     "# we decompose the circuit for the QNN to avoid additional data copying\n",
     "qnn = EstimatorQNN(\n",
-    "    circuit=circuit.decompose(),\n",
+    "    circuit=circuit,\n",
```
woodsp-ibm, line 742: I am not sure of the statement around decompose and additional data copying - I would check this in regards of performance by adding back decompose and see what happens. It maybe that this decomposition speeds things up around binding parameters - maybe internally some copies are avoided during binding this way. Just so happens that a PR was opened in Qiskit earlier today, that is likely related, to flatten off these blocks in NLocal circuits as it leads to way faster binding. I imagine this is more or less what decompose was used here for.

david-alber: @woodsp-ibm  thanks for your insights!
@adekusar-drl I will try to figure out what caused the long eval times and set back the nbsphinx timeout to its original value :).

woodsp-ibm: FYI: related to decompose above (has a plot of times as comparison)

* Qiskit/qiskit-terra/issues/10282

david-alber: A small increase in execution time was observed when decomposing the QNNCircuit (several times - using reps =2,3,4). The respective execution time was still much longer than the original version. This is most definitely linked to the issue @woodsp-ibm referred to.
For now, the QNNCircuit class will be demonstrated in tutorials: 02 & 10.  
Once the issue  [Qiskit/qiskit-terra#10282](https://github.com/Qiskit/qiskit-terra/pull/10284) is closed I will pick up the adjustments for tutorial 05 & 11.

adekusar-drl (approved):
Looks good, thanks!

docs/tutorials/11_quantum_convolutional_neural_networks.ipynb
woodsp-ibm: From my perspective its curious that something which should be more or less identical to the circuit that was formed separately before here behaves differently - is this the case elsewhere too, ie that the performance of using QNNCircuit elsewhere is worse too, just that here it got noticed. The only thing that occurs to me is that the combined circuit before was a plain QuantumCircuit that contained the composed BlueprintCircuit's (i.e feature map and ansatz) whereas QNNCircuit is a BlueprintCircuit itself. Whatever the difference it maybe that the changes going on in Qiskit improve things in any regards making the decompose hopefully unneeded and maybe even making things more comparable. Anyway I just figured I'd mention this since ideally things ought to be better understood so as QNNCircuits use as a convenience is not adversely impacting performance more generally but probably that's best left to after whatever changes Qiskit brings to the parameter binding.

PR #680, opened 2023-08-30
Switch from qiskit-terra to qiskit

woodsp-ibm:
### Summary

Switches over to `qiskit` from `qiskit-terra` 

* requirements.txt updated
* CI updated for install main dependencies
* CONTRIBUTING.MD links etc updated. Plus I removed/simplified some bits which are no correct and may have led to confusion down the line.

I also removed the Neko integration/workflow

### Details and comments

I added the `on-hold` label, as to test these changes I set it for CI to use Qiskit source from main. Nowadays by default it's set to install the release version and I will set it back that way once I check the from main path. Arguably we could drop this from main install but I guess since we have it I left it for now in case we ever want to switch to main to leverage/test against/with some new feature ahead of its release.

With it set to install qiskit from source it does this ok then when it comes to installing ML it does this
```
  Attempting uninstall: qiskit-terra
    Found existing installation: qiskit-terra 0.45.0
    Uninstalling qiskit-terra-0.45.0:
      Successfully uninstalled qiskit-terra-0.45.0
```
which I presume is because the requirements says qiskit >= 0.44 and the metapkg is installed from pypi - since I did not do that aspect from qiskit since its going away - and that forces qiskit-terra back to the version in that meta as can be seen from the install list out of CI log which follows the above.
```
Successfully installed MarkupSafe-2.1.3 cmake-3.27.2 fastdtw-0.3.4 filelock-3.12.3 importlib-metadata-6.8.0 jinja2-3.1.2 joblib-1.3.2 lit-16.0.6 llvmlite-0.40.1 networkx-3.1 numba-0.57.1 nvidia-cublas-cu11-11.10.3.66 nvidia-cuda-cupti-cu11-11.7.101 nvidia-cuda-nvrtc-cu11-11.7.99 nvidia-cuda-runtime-cu11-11.7.99 nvidia-cudnn-cu11-8.5.0.96 nvidia-cufft-cu11-10.9.0.58 nvidia-curand-cu11-10.2.10.91 nvidia-cusolver-cu11-11.4.0.1 nvidia-cusparse-cu11-11.7.4.91 nvidia-nccl-cu11-2.14.3 nvidia-nvtx-cu11-11.7.91 qiskit-0.44.1 qiskit-machine-learning-0.7.0 qiskit-terra-0.25.1 scikit-learn-1.3.0 sparse-0.14.0 threadpoolctl-3.2.0 torch-2.0.1 triton-2.0.0 zipp-3.16.2
```
the relevant part of the above being `qiskit-0.44.1 qiskit-machine-learning-0.7.0 qiskit-terra-0.25.1`

I will let this complete for now to check. The install from main seems to work as it installed 0.45.0 which is the latest main version, but the metapkg aspect which I did no bother with since its going away then undoes this. Since it will work later I think its ok for now especially since we have it set by default to install the release rather than main. If anything needs tweaking later it can be done then but for now it seems good enough to me. And of course in the interim if one wanted to have things running here off source switching the requirements.txt back to qiskit-terra would keep the source installed above.


woodsp-ibm:
I set main.yml as we normally have things now to install the release version so I removed the `on-hold`

adekusar-drl (approved):

PR #690, opened 2023-09-14
[WIP] Remove deprecated code

adekusar-drl:
### Summary
This PR removes:
- CircuitQNN
- EstimatorQNN
- All opflow related code


constraints.txt
```diff
@@ -1,2 +1,3 @@
 numpy>=1.20.0
 ipython<8.13;python_version<'3.9'
+python-subunit<1.4.3
```
woodsp-ibm, line 3: There was a new version of stestr recently uploaded that fixes this incompatibility so the constraint is no longer needed. The Qiskit PR that did similar to workaround the incompatibility was simply closed as a result. The Opt PR that similar dropped that too qiskit-community/qiskit-optimization#559 though it is limiting stestr version (which strictly is not needed - you can see comments there).

adekusar-drl: Removed

woodsp-ibm (approved):
It all looks fine to me. Since its closing the issue, and since it dealt with all the task items listed there I checked them all off in the issue.

One comment I will make, since it had to be changed (well only the format not the content) is the 0..5 migration guide. The classes, like the older QNNs and QuantumKernel often are linked, and these presently do work, but will no longer do so once the new docs, with this guide, are republished. Same is true of older release notes too of course where the same may happen. In the case of the guide there are lot of links though.

woodsp-ibm:
Though in this case they are all absolute links so it would not help - in principle yes!

woodsp-ibm:
Unless they were all relative urls so once you navigated to whatever version folder they were copied to, once they were no longer the current version, they would all still work. Anyway something for the future if/when versions of docs ever happen

PR #691, opened 2023-09-19
Migrate to qiskit_algorithms

### LINKED ISSUE #687: Switch from qiskit.algorithms to qiskit_algorithms
Switch to use qiskit_algorithms so typehints etc refer to this. As its compatible with qiskit.algorithms users existing primitive based use of ML should work without change. I suggest changing the code first and pushing this to CI, leaving tests and tutorials unchanged using qiskit.algorithms just to double check. If there are no issues then changing these over. For algorithm_globals this should now be imported from qiskit_algorithms.utils not qiskit.utils.

- [x] Add qiskit-algorithms to requirements
- [x] Add intersphinx in docs conf.py for qiskit-algorithms
- [x] Change `qiskit.algorithms` to `qiskit_algorithms` for both code and docs class refs etc
- [x] Change unit tests, tutorials, readme sample etc (older migration quides, old release notes can stay as-is)
- [x] Import `algorithm_globals` from `qiskit_algorithms.utils` not `qiskit.utils`.

adekusar-drl:
### Summary
This PR migrates QML from `qiskit.algorithms` to `qiskit_algorithms` as the former has been deprecated.
Closes #687.


requirements.txt
```diff
@@ -1,4 +1,5 @@
 qiskit>=0.44
+qiskit-algorithms
```
woodsp-ibm, line 2:
```suggestion
qiskit-algorithms>=0.2.0
```
The `algorithms_globals` was overlooked in the first release when it was moved out of Qiskit. This was added in 0.2.0 so this update needs that as a min version.

adekusar-drl:
Added a short reno and slightly update the init file.

woodsp-ibm (approved):
LGTM

PR #698, opened 2023-10-12
Used trusted publisher for PyPi

woodsp-ibm:
### Summary

This PR follows Qiskit/rustworkx#1001 to update the release CI workflow to use PyPI's trusted publisher mechanism.

### DIFF
--- a/.github/workflows/deploy-code.yml
+++ b/.github/workflows/deploy-code.yml
@@ -20,6 +20,9 @@ on:
 jobs:
   code_publish:
     runs-on: ubuntu-latest
+    environment: release
+    permissions:
+      id-token: write
     strategy:
       matrix:
         python-version: [3.8]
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-code.yml
```diff
@@ -28,12 +31,12 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Install deps
+        run: pip install -U pip setuptools virtualenv wheel
+      - name: Build sdist
+        run: python3 setup.py sdist bdist_wheel
```
mtreinish, line 37: The future proof way to do this is to use:

```suggestion
        run: pip install -U pip setuptools build
      - name: Build sdist
        run: python3 -m build .
```

but for now this works fine.

mtreinish (approved):

mtreinish:
This LGTM, you'll need to add a release environment but otherwise this is fine. I left an inline comment, but that's more future facing and we can look at that in a follow up PR as it's not related to using a new publishing mechanism

adekusar-drl (approved):

.github/workflows/deploy-code.yml
woodsp-ibm: Ok, sounds good for the future.

PR #700, opened 2023-10-22
Use cross-references to Terra and QML objects in docstrings

desmonty:
### Summary

Fix Issue https://github.com/qiskit-community/qiskit-machine-learning/issues/474

### Details and comments

- Reviewed mentions of QML and Terra classes and updated them into cross-references.
- Used https://github.com/qiskit-community/qiskit-machine-learning/pull/499 work and discussions.

### DIFF
--- a/qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
@@ -69,7 +69,7 @@ class PegasosQSVC(ClassifierMixin, SerializableModelMixin):
         Args:
             quantum_kernel: A quantum kernel to be used for classification.
                 Has to be ``None`` when a precomputed kernel is used. If None,
-                and ``precomputed``is ``False``, the quantum kernel will default to
+                and ``precomputed`` is ``False``, the quantum kernel will default to
                 :class:`~qiskit_machine_learning.kernels.FidelityQuantumKernel`.
             C: Positive regularization parameter. The strength of the regularization is inversely
                 proportional to C. Smaller ``C`` induce smaller weights which generally helps
--- a/qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
@@ -67,8 +67,9 @@ class PegasosQSVC(ClassifierMixin, SerializableModelMixin):
     ) -> None:
         """
         Args:
-            quantum_kernel: a quantum kernel to be used for classification. Has to be ``None`` when
-                a precomputed kernel is used.
+            quantum_kernel: A subclass of :class:`~qiskit_machine_learning.kernels.BaseKernel`
+                to be used for classification.
+                Has to be ``None`` when a precomputed kernel is used.
             C: Positive regularization parameter. The strength of the regularization is inversely
                 proportional to C. Smaller ``C`` induce smaller weights which generally helps
                 preventing overfitting. However, due to the nature of this algorithm, some of the
--- a/qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
+++ b/qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
@@ -67,9 +67,9 @@ class PegasosQSVC(ClassifierMixin, SerializableModelMixin):
     ) -> None:
         """
         Args:
-            quantum_kernel: A subclass of :class:`~qiskit_machine_learning.kernels.BaseKernel`
-                to be used for classification.
-                Has to be ``None`` when a precomputed kernel is used.
+            quantum_kernel: A quantum kernel to be used for classification.
+                Has to be ``None`` when a precomputed kernel is used. If None,
+                default to :class:`~qiskit_machine_learning.kernels.FidelityQuantumKernel`.
             C: Positive regularization parameter. The strength of the regularization is inversely
                 proportional to C. Smaller ``C`` induce smaller weights which generally helps
                 preventing overfitting. However, due to the nature of this algorithm, some of the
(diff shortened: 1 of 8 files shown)


qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -67,8 +67,9 @@ def __init__(
     ) -> None:
         """
         Args:
-            quantum_kernel: a quantum kernel to be used for classification. Has to be ``None`` when
-                a precomputed kernel is used.
+            quantum_kernel: A subclass of :class:`~qiskit_machine_learning.kernels.BaseKernel`
```
woodsp-ibm, line 70: Having the BaseKernel linked on this parameter, when the parameter types are included in the docs from the typehint on the parameter seems like its unnecessary. This is how the docs look now and to me the typehint there already says it needs to be of type BaseKernel and is a link (you can see this in the existing docs too).

I will only comment on this one but there are more instances. My take would be to not have duplication like this.

image

Having it duplicated when it's brought in from the typehint and linked via that seems like

qiskit_machine_learning/algorithms/classifiers/qsvc.py
```diff
@@ -45,7 +45,9 @@ class QSVC(SVC, SerializableModelMixin):
     def __init__(self, *, quantum_kernel: Optional[BaseKernel] = None, **kwargs):
         """
         Args:
-            quantum_kernel: Quantum kernel to be used for classification.
+            quantum_kernel: A subclass of :class:`~qiskit_machine_learning.kernels.BaseKernel`
+                to be used for classification.
+                If None, default to :class:`~qiskit_machine_learning.kernels.FidelityQuantumKernel`.
```
woodsp-ibm, line 50: This sentence could also be added to improve PegasosQSVC with the additional note that this is only the case when precomputed is False.

qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
```diff
@@ -40,7 +40,8 @@ class FidelityStatevectorKernel(BaseKernel):
 
         K(x,y) = |\langle \phi(x) | \phi(y) \rangle|^2.
 
-    In this implementation, :math:`|\phi(y)\rangle` is given by a ``Statevector.data`` array. These
+    In this implementation, :math:`|\phi(y)\rangle` is given by a
+    :attr:`qiskit.quantum_info.Statevector.data` array. These
```
woodsp-ibm, line 44: > is given by a :attr:`qiskit.quantum_info.Statevector.data`

Well yes and no. AerStatevector can also be used here https://qiskit.org/ecosystem/aer/stubs/qiskit_aer.quantum_info.AerStatevector.data.html 

Maybe if it says something like

> is given by a :attr:`qiskit.quantum_info.Statevector.data` array or the same field of a compatible sub-class

My concern is here in making it look too specific to the subclass by having the specific link which might lead to the belief that only that can be used. The parameter itself does have a note about StateVector

qiskit_machine_learning/utils/loss_functions/kernel_loss_functions.py
```diff
@@ -55,7 +55,8 @@ def evaluate(
 
         Args:
             parameter_values: an array of values to assign to the user params
-            quantum_kernel: A ``QuantumKernel`` object to evaluate
+            quantum_kernel: A :class:`~qiskit_machine_learning.kernels.BaseKernel` object
```
woodsp-ibm, line 58: In this case the extra information is incorrect - the type is a TrainableKernel as per the parameter above. This is how it comes out. Maybe it could say `A trainable quantum kernel object to evaluate` so it better describes things and leaves link/type to be generated from the typehint.

image

woodsp-ibm (commented):
Thanks for looking at this. A few comments...

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
desmonty: Reset the first sentence to the original and add the additional note that the quantum kernel defaults to FidelityQuantumKernel if precomputed is `False` and quantum_kernel is `None`.

qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
desmonty: Right. Because AerStatevector is a subclass of Statevector, I've replaced this sentence with the following:

> In this implementation, :math:`|\phi(y)\rangle` is given by the ``data`` attribute of a :class:`~qiskit.quantum_info.Statevector` object or one of its subclasses.

qiskit_machine_learning/utils/loss_functions/kernel_loss_functions.py
desmonty: Thanks, I replaced the wrong class info with your suggestion.

qiskit_machine_learning/algorithms/classifiers/pegasos_qsvc.py
```diff
@@ -67,29 +67,28 @@ def __init__(
     ) -> None:
         """
         Args:
-            quantum_kernel: a quantum kernel to be used for classification. Has to be ``None`` when
-                a precomputed kernel is used.
+            quantum_kernel: A quantum kernel to be used for classification.
+                Has to be ``None`` when a precomputed kernel is used. If None,
+                and ``precomputed``is ``False``, the quantum kernel will default to
```
woodsp-ibm, line 72:
```suggestion
                and ``precomputed`` is ``False``, the quantum kernel will default to
```
Missing a space

image

desmonty: Thanks!

adekusar-drl:
Thank you for the good job!

PR #702, opened 2023-10-26
Fixed typo in 05_torch_connector.ipynb

FelixLehn:
### Summary
simple feature map -> simple parameterized ansatz

### Details and comments
Changed the comment in the regression section as a simple parameterized ansatz is generated and not a simple feature map again.


woodsp-ibm:
@FelixLehn The PR seems to be have been done based on the stable 0.6 branch - it needs to be done on the `main` branch here. Patches to stable are backported, when applicable, from main to stable, so that any fix will also be in future versions too.

FelixLehn:
Okay I've done a PR based on the main branch! :)

woodsp-ibm:
Thanks lets just close then since you have done #703

PR #704, opened 2023-10-31
Typo - Table of Content - 05_torch_connector.ipynb

FelixLehn:
### Summary
SamplerQNN -> EstimatorQNN

### Details and comments
In the table of contents, the Regression part is listed as Regression with SamplerQNN but in the article itself, the EstimatorQNN is used.


adekusar-drl (approved):
Thank you.

PR #705, opened 2023-10-31
Implement evaluation mode for model evaluations

FelixLehn:
### Summary
In PART 1 of the tutorial, the classification and regression models use the model.train() setting but not the model.eval() setting with torch.no_grad().

### Details and comments
The model.eval() setting with torch.no_grad() is used in PART 2, which makes sense as no test data is used in PART 1, but for the overall understanding and the PyTorch alignment, shouldn't it be there, too?


FelixLehn:
You're right. It is otherwise explained twice.

adekusar-drl (approved):
Thank you.

PR #706, opened 2023-10-31
Update Code in part 4

FelixLehn:
### Summary
Trying to fix the issue from https://github.com/qiskit-community/qiskit-machine-learning/issues/685.

### Details and comments
The issue is about the confusion of generating the test kernel matrix with only test featuresand not with the test and training features. When using that the kernel_pca needs to do fit_transform() for the testing which is not the common way, as the kernelPCA should not fit on the testing features.


adekusar-drl (approved):
Thanks for taking care of this issue. Keep going :)

PR #707, opened 2023-11-03
Update CI to use latest windows

woodsp-ibm:
### Summary

Changed `windows-2019` to `windows-latest` (current same as windows-2022) so it running on a newer version and use latest like is done on others. My recollection of it using windows-2019 specifically was when Aer was being built from source. As this is no longer the case it seems better to switch to a newer version and do -latest like is done with Ubuntu and Mac

I already did this in Nature qiskit-community/qiskit-nature#1266 as part of updating CI there.

### Details and comments

The branch rules will need updating to account for the job name change.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -118,9 +118,9 @@ jobs:
             python-version: 3.8
           - os: macos-latest
             python-version: 3.11
-          - os: windows-2019
+          - os: windows-latest
             python-version: 3.8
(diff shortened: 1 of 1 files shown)


adekusar-drl (approved):
Thank you. 
I guess CI checks to be updated as well.

woodsp-ibm:
>I guess CI checks to be updated as well.

Yes, I mentioned in the Details above. I'll update them now.

PR #713, opened 2023-11-10
Unpin mypy

woodsp-ibm:
### Summary

I altered source and config to allow mypy to be unpinned as was done in #711 

For adhoc I changed the code to assign to a different variable so it does not use the same one where the type was already inferred

For torch connector. I could not find a reasonable way to support the newer 1.7.0 and earlier versions via any easy changes to the code. I decided just not to have the type ignore raise an error if its determined to be unused (which varied by version) just for the torch connector and its test file by adding extra specific configuration for that in mypy.ini

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -141,7 +141,7 @@ jobs:
       - run: make lint
         shell: bash
       - run: make mypy
-        if: ${{ !cancelled() && matrix.python-version != '3.10' && matrix.python-version != 3.11 }}
+        if: ${{ !cancelled() }}
         shell: bash
       - name: Machine Learning Unit Tests under Python ${{ matrix.python-version }}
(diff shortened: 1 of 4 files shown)


woodsp-ibm:
I had noticed when CI failed before mypy was pinned it was just 3.8 and 3.9. I thought perhaps that to Python version related. But in checking that mypy indeed was 1.7.0 in 3.8 thru 3.11 I see that for 3.10 and 3.11 it skips the mypy check. We don't do that in other app repos and I assume its just perhaps an artifact of when things were first switched to support that Python version where it was not doable to run mypy for whatever reason. I removed the specific Python checks so mypy runs across the board - it seemed fine locally under 3.11 as I was using that and 3.9 to determine what to change to unpin mypy.

adekusar-drl (approved):

PR #720, opened 2023-11-27
Fix blog post link

ElePT:
### Summary
Updates link to blog post in reno prelude.


mrossinek:
> I wonder if a 0.7.1 release is required.

Indeed, we just saw in Qiskit Nature that without a new release tag, the prelude reno file will simply be filtered because its modification time is newer than any tag and, thus, will be filtered out. Rather than messing with the filter and reverting that, we now opted for https://github.com/qiskit-community/qiskit-nature/pull/1288 which should resolve this and have the prelude re-appear :+1:

Issue #725, opened 2023-12-04
Update tutorials for deprecated circuit draw style


woodsp-ibm:
A number of Tutorials have this deprecation warning message. The drawing should be updated to avoid this.

`qc.draw("mpl")`

>/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/qiskit/visualization/circuit/matplotlib.py:266: FutureWarning: The default matplotlib drawer scheme will be changed to "iqp" in a following release. To silence this warning, specify the current default explicitly as style="clifford", or the new default as style="iqp".
  self._style, def_font_ratio = load_style(self._style)

These are the tutorials - I just put the number not the full name. There can be multiple places within each that needs fixing.

- [x] 01
- [x] 02
- [x] 02a
- [x] 04
- [x] 05
- [x] 10
- [x] 11 (updated by #728)
- [x] 12

edoaltamura:
Fixed by 30aa648b1eb82527a9bfb6be9173c5f526f18b4e for all tutorials: `qc.draw("mpl")` -> `qc.draw("mpl", style="clifford")`. This closes also https://github.com/OkuyanBoga/hc-qiskit-machine-learning/issues/11 and the update is available there after syncing the fork.

PR #728, opened 2023-12-13
Fix QNN for input and weights ordering

### LINKED ISSUE #678: Wrong parameters assignment in the tutorial 11_quantum_convolutional_neural_networks
### Environment

- **Qiskit Machine Learning version**: 0.6.1
- **Python version**: 3.10.9
- **Operating system**: Linux (Ubuntu 20.04)

### What is happening?

I suspect that wrong parameters assignment occurs in the `NeuralNetworkClassifier` of [11_quantum_convolutional_neural_networks.html](https://qiskit.org/ecosystem/machine-learning/tutorials/11_quantum_convolutional_neural_networks.html).

It is difficult to explain in words, so I created a notebook: [11_quantum_convolutional_neural_networks_issue.ipynb](https://github.com/derwind/qiskit-machine-learning/blob/issue/wrong-parameters-assign/docs/tutorials/11_quantum_convolutional_neural_networks_issue.ipynb)

So you can reproduce the issue by executing the attached notebook.

----

In short, `train_images` data are assigned to the conv layer whose name is `"c2"` and the parameters from `11_qcnn_initial_point.json` are assigned to `ParameterVectorElement(c2[8])`, `ParameterVectorElement(c2[9])`, ...

The contents up to cell No.14 are the same as in the tutorial. However, only cell No.13 has `maxiter=0` to finish immediately. My explanation will start from the cell No.15.

### What should happen?

I think `train_images` data should be assigned to `x[0]`, `x[1]`, ... and the parameters from `11_qcnn_initial_point.json` should be assigned to `c1[0]`, `c1[1]`, ...

Note that `c` of `c1` in the tutorial is actually not a latin character`c` (LATIN SMALL LETTER C; U+0063) but a **cyrillic character** `с` (CYRILLIC SMALL LETTER ES; U+0441). After c2, `c` of c2, c3 are LATIN SMALL LETTER C.

### Any suggestions?

_No response_

### LINKED ISSUE #727: QNN parameter binding incorrect
inputs and weights are concatenated as parameters for the primitive. However the primitives bind the data to the circuit in the order in which the parameters are returned by the circuit.parameters function (which is sorted so its always the same for the circuit). Hence depending on parameter naming the order may very well not end up being such that the inputs and weights are not bound as per the parameter sequences supplied defining the input and weight parameters. 

Findings from investigating this Qiskit Slack workspace message https://qiskit.slack.com/archives/C7SS31917/p1702175603460089

woodsp-ibm:
### Summary

Fixes #727

Reassigns QNN circuit parameters so the order of circuit.parameters, which is alphanumerically ordered when returned, comes out as as desired for the primitive and the data (parameters) as ordered there which is inputs followed by weights in the given array.

### Details and comments

I added a method to reassign parameters in the circuit so it's a new circuit copy from that supplied, but whose parameters are in the order needed by the data passed to the primitive when it runs. This I put in `self._circuit` so that internally its using this circuit now when running etc. As there are parameter and circuit getters, I left these all returning what was passed in originally ( as before) so the original circuit is assigned to a different named instance var which is returned by the getter. 

So what is run is equivalent but not the circuit as that passed. An alternative would be to alter the data order so it corresponds to the order of the parameters. This seemed less desirable on initial discussion since it changes the data. 

This PR still needs a release note but I figured lets settle that this is how we want to proceed before I do that. Not that I imagine it would change the text thereof though. **_Update_**: I added one anyway...

- [x] Add release note
- [x] Address change in tutorial `11_quantum_convolutional_neural_networks` (see comment below in this page)

Fixes #678

In updating the tutorial with a new initial point I saw the issue described in #678 where "c1" was not a latin c from the overall parameter order (indeed searching for c in the text did not locate it) So this addresses that issue too.
As part of the update I also set a random_state to the test_train_split so the result in the notebook in the same each time its run. Otherwise the split is different each time which made the results different each time.

This is one of the tutorials, per issue #725, where the circuit.draw("mpl") has deprectation messages so since this is being modified I corrected that as well, just for this tutorial, and used the clifford style which keeps the same look as before


woodsp-ibm:
A note:

As the standard feature map is mostly used in tests and tutorials with "x" as the name and the standard ansatzes used with greek theta the order was naturally as needed so no change in tests was needed and the tutorials look the same as before etc....

But with the exception of 11_quantum_convolutional_neural_networks whose behavior is now different and may need to be reviewed. In it it says

> Since model training may take a long time we have already pre-trained the model for some iterations and saved the pre-trained weights. We’ll continue training from that point by setting initial_point to a vector of pre-trained weights.

The plot no longer looks flat like it did so most likely the initial point needs reviewing and the accuracy is lower now - most likely because maxiter is limited to 200 and it no longer reaches a more optimal solution

image

adekusar-drl (approved):
Thanks. Looks good to me.

PR #739, opened 2024-01-19
Deploy docs to GitHub Pages

arnaucasau:
### Summary

This PR changes the `deploy-docs.yml` workflow to deploy the documentation to GitHub Pages to migrate the project away from the qiskit.org domain.

The deployment to the `gh-pages` branch will not overwrite the `locale` folder, where I will save all the translations in a follow-up once the first deployment is done.

More information about the change and a quick guide to set up GitHub pages can be found at: https://github.com/Qiskit/ecosystem/issues/578

Reminder task for maintainers here to update the doc link in the About text for the repo on the main page once this change happens and the docs are re-deployed.

- [ ] Update About text doc link


woodsp-ibm (approved):
Thanks for the changes. I labelled this so it will be backported to stable after which I will deploy the docs again so things can checked out as/if needed.

PR #766, opened 2024-02-15
Remove all other languages other than Spanish and Japanese

SooluThomas:
### Summary

Spanish and Japanese are the only languages that was 100% translated & proofread as of Jan 2024

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -138,15 +138,8 @@ numfig_format = {"table": "Table %s"}
 
 translations_list = [
     ("en", "English"),
-    ("bn_BN", "Bengali"),
-    ("fr_FR", "French"),
-    ("hi_IN", "Hindi"),
     ("ja_JP", "Japanese"),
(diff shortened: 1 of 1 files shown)


Eric-Arellano (approved):

woodsp-ibm:
This just basically removed them from being selected from the dropdown right - the translated pages (locales) will still be there, unless removed, from the gh_pages branch where all this is published  - so you if you know you could still alter the url and see things if this is all that is changed. Or are well also going to delete the other locales, ie all apart from Spanish and Japanese, there too?

Eric-Arellano:
> so if you know you could still alter the url and see things if this is all that is changed.

I think that's fine. There isn't much downside. There's also an upside that it will make sure we don't 404 for old Google results for the translated docs that used to be hosted on qiskit.org and now redirect.

woodsp-ibm (approved):

PR #769, opened 2024-02-16
Fix tutorials to work with Qiskit 1.0

woodsp-ibm:
### Summary

Qiskit 1,0 removed the deprecated qiskit.tools which included some .jupyter from which the magics that did the version and copyright were imported. This PR adds a local file in the tutorials folder that contains just these magics and changes the import to use this.

Similar fix to qiskit-community/qiskit-algorithms#156

I also removed an import of BasicAer in one tutorial as that fails now - it was unused - probably from a time before primitives. 

### Details and comments

I do not expect this to pass until the other failures are sorted. A new release of Qiskit Algorithms is required I believe, but at least its ready.

### DIFF
--- a/docs/tutorials/01_neural_networks.ipynb
+++ b/docs/tutorials/01_neural_networks.ipynb
@@ -1047,7 +1047,7 @@
     }
    ],
    "source": [
-    "import qiskit.tools.jupyter\n",
+    "import tutorial_magics\n",
     "\n",
     "%qiskit_version_table\n",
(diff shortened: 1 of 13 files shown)


woodsp-ibm:
Qiskit Algorithms 0.3.0 was released so all passes - well its failing in testing the stable set of tutorials that are there against main here. Since this is intended to be backported to fix them too that will have to be ignored.

adekusar-drl (approved):

PR #774, opened 2024-02-21
Add Python 3.12 support

### LINKED ISSUE #759: Add Python 3.12 support
Qiskit 0.45.1 added Python 3.12 support, and its been added to Algorithms see qiskit-community/qiskit-algorithms#108 (although at the time of writing this algorithms has yet to officially release a version supporting this).

CI will likely need to be specific to 3.12.0 until an updated testtools is released - see qiskit-community/qiskit-algorithms#113 which goes into the failure using 3.12.1

~Note #729 in regards of any newer version of qiskit algorithms which will most likely need sorting around the same time~ This now has already been taken care of in a way that supported 0.2.0 as well as the changes that were done in 0.3.0 - see #761 for more info.

- [X] #759 adds 3.12 support but needs to limit CI to 3.12.0 for the reason above
- [ ] ~When possible change `3.12.0` to `3.12` to allow newer bug fix version of Python 3.12 to be used in CI~ No longer needed as the breaking change in 3.12.1 was reverted in 3.12.2

woodsp-ibm:
### Summary

~See #759 I will not mark this to close that since this still CI needs to use 3.12.0 specifically as the breaking change to testtools, referred there has not yet been addressed. So rather than have it close it and then open another specifically around just using 3.12 when we are able, I just put a task list in there so this does one of the entries. The other will need a release of testtools first.~ Fixed in 3.12.2 which reverted the breaking change

Closes #759

I marked this stable backport in order to facilitate a release that supports 3.12. 

See also qiskit-community/qiskit-nature#1338 if you want to see a similar change to CI to add 3.12
**Update**: qiskit-community/qiskit-nature#1341 unpins back to just 3.12

### Details and comments

If/when merged will require branch rules updating - due the changes in jobs being run now versus what was done before e.g. macos-latest 3.11 is no more as we now do macos-latest 3.12. Which jobs are Required to run/pass is configured via the Branch Rules so until that is updated there will be a mismatch. Once it is updated these will be the new rules and any existing PRs would need updating so the new rules kick in for them too.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,6 @@
 # This code is part of a Qiskit project.
 #
-# (C) Copyright IBM 2021, 2023.
+# (C) Copyright IBM 2021, 2024.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
@@ -112,16 +112,16 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8, 3.9, '3.10', 3.11]
+        python-version: [3.8, 3.9, '3.10', 3.11, 3.12.0]
         include:
           - os: macos-latest
             python-version: 3.8
           - os: macos-latest
-            python-version: 3.11
+            python-version: 3.12.0
           - os: windows-latest
             python-version: 3.8
           - os: windows-latest
-            python-version: 3.11
+            python-version: 3.12.0
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
@@ -183,7 +183,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8, 3.11]
+        python-version: [3.8, 3.12]
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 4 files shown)


woodsp-ibm:
Note mainly for new maintainers, While for Ubuntu it's tested against all the Python versions supported, for Mac and Windows its done just against the min and max versions. Only the unit tests need to use the exact 3.12.0 version (testtools issue only affects unittests) - the tutorials we again do min and max versions and there its simply 3.12 since it does not need to be down to the exact bug fix version.

.github/workflows/main.yml
```diff
@@ -183,7 +183,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8, 3.11]
+        python-version: [3.8, 3.12]
```
adekusar-drl, line 186: I wonder why it is 3.12 and 3.12.0 in some places. Is it something new in the python versioning?

woodsp-ibm: From the https://github.com/qiskit-community/qiskit-machine-learning/pull/774#issuecomment-1958091770 above:

> Only the unit tests need to use the exact 3.12.0 version (testtools issue only affects unittests) - the tutorials we again do min and max versions and there its simply 3.12 since it does not need to be down to the exact bug fix version.

adekusar-drl (dismissed):
Thank you

woodsp-ibm:
@adekusar-drl, @t-imamichi did a PR for Opt last night where he used just 3.12 and it passed. That surprised me as the testtools issue, while fixed by @mtreinish testtools has not yet released. In looking it seems that the breaking change that was causing the failure in 3.12.1 has been reverted https://github.com/python/cpython/pull/114470 in 3.12.2 which is the latest and what Opt ran on last night. I will check with algs, where I know it failed before, and drop it to 3.12. If that all passes I will do that here before this is merged so we don't have to bother with updating it later (and I can update Nature too which was already done)

woodsp-ibm:
I marked it OnHold while I investigate as per above comment.

adekusar-drl (approved):

PR #778, opened 2024-02-26
Bug/607 fidelity statevector kernel cannot be pickled

### LINKED ISSUE #607: FidelityStatevectorKernel cannot be pickled
### Environment

- **Qiskit Machine Learning version**: qiskit-machine-learning==0.6.0
- **Python version**:  Python 3.10.4
- **Operating system**:  Ubuntu 20.04.5 LTS

### What is happening?

qiskit_machine_learning.kernels.fidelity_statevector_kernel.FidelityStatevectorKernel cannot be pickled.  I.e., it gives error:

> _pickle.PicklingError: Can't pickle <functools._lru_cache_wrapper object at 0x7f7de13c7e20>: it's not the same object as qiskit_machine_learning.kernels.fidelity_statevector_kernel.FidelityStatevectorKernel._get_statevector_

Being able to pickle objects is required for a lot of frameworks we'd want to use machine learning components with.  
I.e., this error prevents a lot of things - such as using it in a parallelization framework like ray (which requires pickling) or using it with common machine learning approaches such as sklearn GridSearchCV which also requires pickling.

The cause is a known issue of using lru_cache as is done in the code for FidelityStatevectorKernel (line 100-101):
```
# Create the statevector cache at the instance level.
self._get_statevector = lru_cache(maxsize=cache_size)(self._get_statevector_)
```

E.g., see: https://stackoverflow.com/questions/54363043/pickling-lru-cached-function-on-object
There are a number of fixes, including some listed in the SO post above.

```
from qiskit.circuit.library import PauliFeatureMap

from qiskit_machine_learning.kernels.fidelity_statevector_kernel import (
    FidelityStatevectorKernel,
)

import pickle


kernel = FidelityStatevectorKernel(
    feature_map=PauliFeatureMap(
        feature_dimension=3,
        reps=1,
        entanglement="linear",
        alpha=1.0,
        paulis=("Z", "ZZ"),
    )
)

kernel_obj = pickle.dumps(kernel)
```

### What should happen?

Fix the class so it can be pickled.  I.e., the above code should not throw an error, but allow successfully pickling the object.

### Any suggestions?

You can look at the stack overflow post linked above, and other similar posts.  There are some easy fixes, such as writing __getstate__ and __setstate__ to handle that attribute specially.

oscar-wallis:
### Summary
Fixes #607 where FidelityStatevectorKernel was throwing an error when pickled due to a known pickling error with functools.lru_cache. This has been fixed by adding a self.cache_size attribute and custom __getstate__ and __setstate__ functions used when pickling and unpickling.

### Details and comments
When pickled the lru_cache is set to None in the object dict. When unpicked the cache is re-initiliased with the new cache_size variable and __getstate__ and __setstate__. Unit tests added to check object pickling and unpickling as well as manually using __getstate__ and __setstate__ to initialise a FidelityStatevectorKernel. Passed all other unit tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,4 +127,3 @@ venv.bak/
 
 # stestr repostiory
 .stestr/
-bug607_replicated.py
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,4 @@ venv.bak/
 
 # stestr repostiory
 .stestr/
+bug607_replicated.py
--- a/qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
+++ b/qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
@@ -161,7 +161,7 @@ class FidelityStatevectorKernel(BaseKernel):
         # pylint: disable=no-member
         self._get_statevector.cache_clear()
 
-    def __getstate__(self) -> dict:
+    def __getstate__(self) -> dict[str, Any]:
         kernel = dict(self.__dict__)
         kernel["_get_statevector"] = None
         return kernel
--- a/qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
+++ b/qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
@@ -162,10 +162,10 @@ class FidelityStatevectorKernel(BaseKernel):
         self._get_statevector.cache_clear()
 
     def __getstate__(self) -> dict:
-        picklable_kernel = dict(self.__dict__)
-        picklable_kernel["_get_statevector"] = None
-        return picklable_kernel
+        kernel = dict(self.__dict__)
+        kernel["_get_statevector"] = None
+        return kernel
 
-    def __setstate__(self, unpickled_kernel):
-        self.__dict__ = unpickled_kernel
+    def __setstate__(self, kernel):
+        self.__dict__ = kernel
         self._get_statevector = lru_cache(maxsize=self.cache_size)(self._get_statevector_)
--- a/qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
+++ b/qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
@@ -14,7 +14,7 @@
 from __future__ import annotations
 
 from functools import lru_cache
-from typing import Type, TypeVar
+from typing import Type, TypeVar, Any
 
 import numpy as np
 
--- a/qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
+++ b/qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
@@ -166,6 +166,6 @@ class FidelityStatevectorKernel(BaseKernel):
         kernel["_get_statevector"] = None
         return kernel
 
-    def __setstate__(self, kernel):
+    def __setstate__(self, kernel: dict[str, Any]):
         self.__dict__ = kernel
         self._get_statevector = lru_cache(maxsize=self._cache_size)(self._get_statevector_)
--- a/qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
+++ b/qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
@@ -1,6 +1,6 @@
 # This code is part of a Qiskit project.
 #
-# (C) Copyright IBM 2023.
+# (C) Copyright IBM 2023, 2024.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
@@ -97,7 +97,7 @@ class FidelityStatevectorKernel(BaseKernel):
         self._auto_clear_cache = auto_clear_cache
         self._shots = shots
         self._enforce_psd = enforce_psd
-
+        self.cache_size = cache_size
         # Create the statevector cache at the instance level.
         self._get_statevector = lru_cache(maxsize=cache_size)(self._get_statevector_)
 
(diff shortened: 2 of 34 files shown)


qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
```diff
@@ -97,7 +97,7 @@ def __init__(
         self._auto_clear_cache = auto_clear_cache
         self._shots = shots
         self._enforce_psd = enforce_psd
-
+        self.cache_size = cache_size
```
woodsp-ibm: Should this be `self._cache` ie something we consider private and uses should not change. At present its a public instance var and part of the public API - something we document via Attributes in this case - but in this case its not something I think we want users to change other that via the constructor.

releasenotes/notes/fix-fid_statevector_kernel-pickling-b7fa2b13a15ec9c6.yaml
```diff
@@ -0,0 +1,10 @@
+---
+fixes:
+  - |
+    Fixed a bug where :class:`~qiskit_machine_learning.kernels.FidelityQuantumKernel` threw an error when pickled. This was due
```
woodsp-ibm, line 4: Just as a note you can do this too, as long as its a class that is unique ''':class:`.FidelityQuantumKernel` ``` where with the `.` sphinx will find the class without you having to type the full thing. 

Unless any of how it was fixed affects the end user then just stating its fixed and can be pickled as expected now suffices. The release notes are really just the public API and changes/fixes perceived when using the API for end-users - internal details of how it was fixed, unless they have additional impacts in some way are not really relevant to such users. Eg. to fix things we could have stopped user lru_cache and done something else internally. All the end user needs to know is that the problem is fixed. Hence to me only the content of the first sentence is needed here for a release note for this.

woodsp-ibm (commented):
Normally if we have a bug that does not show up in the test cases it suggests we need one to test that aspects for better coverage and to make sure it things stay working. I think a test case that pickles and unpickles is needed.

qiskit_machine_learning/kernels/fidelity_statevector_kernel.py
oscar-wallis: yes it should be _cache_size, changed now

releasenotes/notes/fix-fid_statevector_kernel-pickling-b7fa2b13a15ec9c6.yaml
oscar-wallis: Ah, okay I will reduce it down to the first sentence.

oscar-wallis:
> Normally if we have a bug that does not show up in the test cases it suggests we need one to test that aspects for better coverage and to make sure it things stay working. I think a test case that pickles and unpickles is needed.

Does that mean a test is needed where we pickle and unpickle all qml objects?

woodsp-ibm:
> Does that mean a test is needed where we pickle and unpickle all qml objects?

Seems like a lot of objects! This is much more an edge case I guess due to lru_cache so I would leave it at just a test for this object - makes sure it works in CI over the platforms and continues to do so.

oscar-wallis:
Added unit testing and edited release notes.

test/kernels/test_fidelity_statevector_kernel.py
```diff
@@ -343,6 +344,52 @@ def test_properties(self):
...
+            self.assertNotEqual(kernel1, kernel2)
+
+        with self.subTest("Fail if objects 1 and 3 are not unique"):
+            self.assertNotEqual(kernel1, kernel3)
+
+        with self.subTest("Fail if caches 1 and 2 are incorrectly initialised"):
+            self.assertEqual(type(kernel1._get_statevector), type(kernel2._get_statevector))
+
+        with self.subTest("Fail if caches 1 and 3 are incorrectly initialised"):
+            self.assertEqual(type(kernel1._get_statevector), type(kernel3._get_statevector))
```
OkuyanBoga, line 375: To make testing clearer, can you please test messages more informative?

OkuyanBoga: When the tests fail, it would be hard to understand for the people, for example, where it is originated
Instead of `"Fail if objects 1 and 3 are not the same type"`, maybe, something like 
`"Pickle test failed, kernels are not the same type"`.

oscar-wallis: Are these better 83a00d3 ? B)

OkuyanBoga: Sounds better, thanks

test/kernels/test_fidelity_statevector_kernel.py
```diff
@@ -343,6 +344,55 @@ def test_properties(self):
...
+        with self.subTest("Pickle fail, kernels are not unique objects"):
+            self.assertNotEqual(kernel1, kernel2)
+
+        with self.subTest("Pickle fail, kernels are not unique objects"):
+            self.assertNotEqual(kernel1, kernel3)
+
+        with self.subTest("Pickle fail, caches are not the same type"):
+            self.assertEqual(type(kernel1._get_statevector), type(kernel2._get_statevector))
+
+        with self.subTest("Pickel fail, caches are not the same type"):
```
declanmillar, line 375:
```suggestion
        with self.subTest("Pickle fail, caches are not the same type"):
```

declanmillar (changes requested):
Thanks very much for addressing this issue! I just have a few minor suggestions.

.gitignore
```diff
@@ -127,3 +127,4 @@ venv.bak/
 
 # stestr repostiory
 .stestr/
+bug607_replicated.py
```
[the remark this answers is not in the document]
oscar-wallis: Yes, it was added accidentally. I just wanted to make sure I didn't upload the bug replicating file, I didn't realise the gitignore is added here too.

oscar-wallis:
> Thanks very much for addressing this issue! I just have a few minor suggestions.

I appreciate the suggestions! I'm surprised some of them weren't picked up by CI. All the changes were approved.

declanmillar:
You'll need to add `from typing import Any` to `fidelity_statevector_kernel.py`, but, unfortunately, I'm not allowed to make that suggestion as it's in an unedited code block.

oscar-wallis:
> `from typing import Any`

Can do, why does it need to be added?

adekusar-drl (approved):
Looks good to me. Thank you.

edoaltamura (approved):

Issue #726, opened 2023-12-04
test_change_kernel and test_qsvr fail in clean stable 0.7 environment


iyanmv:
### Environment

- **Qiskit Machine Learning version**: 0.7.1
- **Python version**: 3.11.6
- **Operating system**: Arch Linux

### What is happening?

I'm trying to improve [the PKGBUILD for AUR](https://aur.archlinux.org/packages/python-qiskit-machine-learning) and run  the python tests in the `check()` function.

The following tests fail in a clean chroot environment:

- `test/algorithms/regressors/test_qsvr.py::TestQSVR::test_change_kernel`
- `test/algorithms/regressors/test_qsvr.py::TestQSVR::test_qsvr`
- `test/algorithms/classifiers/test_fidelity_quantum_kernel_pegasos_qsvc.py::TestPegasosQSVC::test_save_load` (core dumped)

```python
_________________________ TestQSVR.test_change_kernel __________________________

self = <test.algorithms.regressors.test_qsvr.TestQSVR testMethod=test_change_kernel>

    def test_change_kernel(self):
        """Test QSVR with QuantumKernel later"""
        qkernel = FidelityQuantumKernel(feature_map=self.feature_map)
    
        qsvr = QSVR()
        qsvr.quantum_kernel = qkernel
        qsvr.fit(self.sample_train, self.label_train)
        score = qsvr.score(self.sample_test, self.label_test)
    
>       self.assertAlmostEqual(score, 0.38359, places=4)
E       AssertionError: 0.38411965819305205 != 0.38359 within 4 places (0.0005296581930520627 difference)

test/algorithms/regressors/test_qsvr.py:71: AssertionError
```
```python
______________________________ TestQSVR.test_qsvr ______________________________

self = <test.algorithms.regressors.test_qsvr.TestQSVR testMethod=test_qsvr>

    def test_qsvr(self):
        """Test QSVR"""
        qkernel = FidelityQuantumKernel(feature_map=self.feature_map)
    
        qsvr = QSVR(quantum_kernel=qkernel)
        qsvr.fit(self.sample_train, self.label_train)
        score = qsvr.score(self.sample_test, self.label_test)
    
>       self.assertAlmostEqual(score, 0.38359, places=4)
E       AssertionError: 0.38411965819305205 != 0.38359 within 4 places (0.0005296581930520627 difference)

test/algorithms/regressors/test_qsvr.py:60: AssertionError
```

<details>

```
[...]

Issue #434, opened 2022-07-07
Improve QSVR unit tests 


adekusar-drl:
### What is the expected enhancement?

Currently unit test in [TestQSVR](https://github.com/Qiskit/qiskit-machine-learning/blob/main/test/algorithms/regressors/test_qsvr.py) are based on the classification dataset, e.g., labels are `[0, 0, 1, 1]`. Such a dataset is not very well suited to test regressors. This issue intends to fix this problem and implement unit tests on a more suitable dataset.

adekusar-drl:
@AbdullahKazi500 do you have ideas what exactly to be fixed in this issue?

adekusar-drl:
Sure, I can. But the question I posted above is still valid.

adekusar-drl:
Thanks for getting back. A few comments:
- A dataset should be really small, like a couple of data points. Currently, there are 4 of them in the test. One of the ideas to try out is to use the regression dataset from https://github.com/Qiskit/qiskit-machine-learning/blob/main/docs/tutorials/02_neural_network_classifier_and_regressor.ipynb. But you are welcome to propose something else.
- I suggest to modify existing test case unless it is inconvenient to update them.
- There's no need to split the dataset, these are just tests.
- One metric is enough. Currently, tests compare evaluated score vs true value.
- Not sure hyperparameterization is required here.

So, the work should be pretty straight forward. Let me know if you have questions.

adekusar-drl:
No, all of them should fixed to certain values and that's it. Tests must check that algorithms work and results are meaningful, but no more.

adekusar-drl:
@AbdullahKazi500 is there any progress here?

edoaltamura:
Hi @AbdullahKazi500 has there been any recent progress? If so, please organize and submit the commits in a PR.

edoaltamura:
I haven't heard recent updates on this test update and wanted to touch base and share a few timeline considerations FYI. We aim to include this update in the v.0.8 initial release, which is expected around the end of October 2024. This means that we would like to support long-standing community contributions - like this one - pull-requested by Friday 25 October, and leave enough time to coordinate the merging and release preparation.

Are you still available to work on this with the suggestions in https://github.com/qiskit-community/qiskit-machine-learning/issues/434#issuecomment-1476375330 by then or rather step out on this occasion?

RishiNandha:
[...]

PR #779, opened 2024-02-28
Add M1 image for MacOS for tests

woodsp-ibm:
### Summary

#726 documents issues using M2. Github recently made M1 images available and this updates the main workflow to include jobs running under ~3.8~ 3.10 and 3.12 (~lower~ (3.10 is is the min available at present for arm64) and upper Python versions supported).

Adding M1 had been talked about in Optimization and and issue exists there qiskit-community/qiskit-optimization#593 which has a link to the github announce for the M1 capability. Linking the issue to this PR can also inform Opt on how I did it here.

### Details and comments

I marked on OnHold - lets see how this goes. I kinda expect it may fail the tests that #726 does. If failed as expected, I altered them as described below and removed the OnHold as it all passes.


woodsp-ibm:
The lowest version arm64 Python is presently 3.10 so I switched to that as a minimum as the 3.8 had before failed since it could not find a compatible Python

It did indeed fail on the unit tests indicated in #726. I added enforce_psd=False to both tests, as they should be PSD since its an ideal statevector computation the sampler uses.

````
==============================
Failed 2 tests - output below:
==============================

test.algorithms.regressors.test_qsvr.TestQSVR.test_qsvr
-------------------------------------------------------

Captured traceback:
~~~~~~~~~~~~~~~~~~~

      File "/Users/runner/work/qiskit-machine-learning/qiskit-machine-learning/test/algorithms/regressors/test_qsvr.py", line 60, in test_qsvr
    self.assertAlmostEqual(score, 0.38359, places=4)

      File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/unittest/case.py", line 939, in assertAlmostEqual
    raise self.failureException(msg)

    AssertionError: 0.38411965819305227 != 0.38359 within 4 places (0.0005296581930522848 difference)


test.algorithms.regressors.test_qsvr.TestQSVR.test_change_kernel
----------------------------------------------------------------

Captured traceback:
~~~~~~~~~~~~~~~~~~~

      File "/Users/runner/work/qiskit-machine-learning/qiskit-machine-learning/test/algorithms/regressors/test_qsvr.py", line 71, in test_change_kernel
    self.assertAlmostEqual(score, 0.38359, places=4)

      File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/unittest/case.py", line 939, in assertAlmostEqual
    raise self.failureException(msg)

    AssertionError: 0.38411965819305227 != 0.38359 within 4 places (0.0005296581930522848 difference)
```

woodsp-ibm:
These are the new M1 jobs

image

woodsp-ibm:
> Machine Learning Unit Tests / MachineLearning (macos-latest, 3.12) (pull_request) Failing after 10m 

Hmmm, this installed Python 3.12.1 for some reason rather than the new 3.12.2 and 3.12.1 causes a known issue with testtools (used by stestr that we use for testing) where this issue was a breaking change in the unit test result that was reverted in 3.12.2. I'll try just re-running the job and see. It passed the nightly CI so unless there is some more recent change - I do not think anything done by  this PR should cause this.

woodsp-ibm:
This all passes, but as expected the couple of unit tests that we saw fail on an M2 machine, failed here too with M1. I updated them to remove the enforce psd as described above. At this point its ready. If we want to have these jobs Required too, I do not see why we wouldn't, then the Branch Rules need to be updated for that. Since this is only CI, and to keep CI current there in stable, since otherwise I thinks it the same as main, I think adding the StableBackportPotential label would be good to have it backported as well.

adekusar-drl:
It is interesting, but I can't reproduce the issue on M1 and python 3.11. I think, putting `enforce_psd` to false is not a proper way for tests. May be it is better to either reduce the accuracy or replace the tests. If I recall correctly, the tests have been constructed out of a classification problem and the tests are for a regression model.

woodsp-ibm:
> It is interesting, but I can't reproduce the issue on M1 and python 3.11. I think, putting enforce_psd to false is not a proper way for tests. May be it is better to either reduce the accuracy or replace the tests.

Strange - we have not really determined why this happens. Its not only M1/M2, where this seems to have surfaced more, as #776 has the same issue with a different OS on x86 This is qiskit.primitives Sampler so the result should be ideal and not need enforce_psd. I can tell you if you replace that by the Aer sampler in ideal mode (shots=None) then the test passes. So its not clear the exact cause of why, in certain environments, this comes up with a different answer. Yes one could reduce the accuracy of the test, but if enforce_psd is not required - it should not be required for an ideal outcome right, that seemed better to me - the fact that it changes things under certain conditions,  presumably a slightly different sampler result,  and causes a different outcome, without further investigation of what is happening I guess we will not know.

adekusar-drl:
@oscar-wallis What do you think how we should proceed here? I'm fine either way: merge as it is now, update the tests, or something else.

woodsp-ibm:
Let me link #434 which is intended to improve these tests - if we have M1 job in place at that point it will be interesting to see if there is any difference.

OkuyanBoga (approved):
I am happy with the current, however, do we need a release note for this?

woodsp-ibm:
> do we need a release note for this?

Release notes are for changes to/around the public API that are of concern to end users. This can include bug fixes of course since these can be perceived when using the public API, and performance. Internal refactoring that does not reflect in public API would not need one, nor do changes to CI like this.

Issue #716, opened 2023-11-13
Dimension Mismatch in torch_connector.py


miles0428:
### Environment

- **Qiskit Machine Learning version**: 0.7.0
- **Python version**: 3.9.6
- **Operating system**: macOS

### What is happening?

I am attempting to build a quantum version of a convolutional layer with Qiskit and PyTorch, but I encountered an error related to the Einstein summation method. I believe this issue arises because the dimensions do not match when I execute `loss.backward()`.

The problem specifically stems from the fact that the `torch_connector.py` file utilizes `weights_grad = torch.einsum("ij,ijk->k", grad_output.detach().cpu(), weights_grad)` for the 3D case. However, in my implementation, both `grad_output.detach().cpu()` and `weights_grad` are in the 4D case. To resolve this, I modified the expression from `"ij,ijk->k"` to `"ijl,ijlk->k"`, and this corrected the problem.

### How can we reproduce the issue?

Quantum Convolution Layer
```python 
import torch
import qiskit as qk
from qiskit import QuantumCircuit
import torch.nn as nn
from qiskit_machine_learning.neural_networks import SamplerQNN
# from torch_connector import TorchConnector
from qiskit_machine_learning.connectors import TorchConnector
import torch.nn.functional as F


# Define the quantum circuit

class Quanv2d(nn.Module):
    def __init__(self,input_channel,output_channel,num_qubits,num_weight, kernel_size=3, stride=1):
        super(Quanv2d, self).__init__()
        self.kernel_size = kernel_size
        self.stride = stride
        self.input_channel = input_channel
        self.output_channel = output_channel
        self.qnn = TorchConnector(self.Sampler(num_weight,kernel_size * kernel_size * input_channel, num_qubits))
        #check if 2**num_qubits is greater than output_channel
        assert 2**num_qubits >= output_channel, '2**num_qubits must be greater than output_channel'

    def Sampler(self, num_weights, num_input, num_qubits = 3):
        qc = QuantumCircuit(num_qubits)
        weight_params = [qk.circuit.Parameter('w{}'.format(i)) for i in range(num_weights)]
        input_params = [qk.circuit.Parameter('x{}'.format(i)) for i in range(num_input)]
        #construct the quantum circuit with the parameters
        """
            build quantum circuit here
        """

        #use SamplerQNN to convert the quantum circuit to a PyTorch module
[...]

PR #793, opened 2024-04-03
Patches einsum dimensionality in `torch_connector` - #716

edoaltamura:
### Summary
- Closes https://github.com/qiskit-community/qiskit-machine-learning/issues/716
- Closes https://github.com/OkuyanBoga/hc-qiskit-machine-learning/pull/18 (merged)
- Adds release notes since the dimensionality is automatically matched to the input at the user level.

✅ I have added the tests to cover my changes.
✅ I have updated the documentation accordingly.
✅ I have read the CONTRIBUTING document.


test/connectors/test_torch_connector.py
```diff
@@ -44,6 +45,24 @@ def setup_test(self):
             torch.tensor([[[1.0], [2.0]], [[3.0], [4.0]]]),
         ]
 
+    def test_get_einsum_signature(self):
```
woodsp-ibm, line 48: Unit tests should really (best practice) test the public API and not implementation specific private methods etc that are an implementation detail. Such methods should be covered of course overall by unit tests which may mean calling the public API with certain values etc. In this case can the function working be covered by using the connector with suitably sized data and making sure a result is returned as expected?

edoaltamura: Doing that is possible. Would you advise rebuilding the test using something similar to https://github.com/qiskit-community/qiskit-machine-learning/issues/716#issuecomment-1990945075?

adekusar-drl: I agree with @woodsp-ibm. I think the best way is exactly to have a test like in the issue. If possible, I'd generalize it.

qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -52,12 +52,53 @@ class Module:  # type: ignore
...
+        RuntimeError: If the number of dimensions exceeds the character limit.
+        ValueError: If an invalid return type is provided.
+    """
+    trace = ""
+    char_limit = 26
+    for i in range(n_dimensions):
+        # pylint: disable=wrong-spelling-in-comment
+        # chr(97) == 'a'
+        trace += chr(97 + i)
+        if i >= char_limit:
```
woodsp-ibm, line 78: Can we not just check n_dimensions right at the start rather than doing it in the for loop each time.  Also as CHAR_LIMIT is more a constants can be not have it so (ie CAPS).

qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -52,12 +52,53 @@ class Module:  # type: ignore
...
+            raise RuntimeError(
+                f"Cannot define an Einstein summation with more tha {char_limit:d} dimensions."
+            )
+
+    if return_type == "input":
+        signature = f"{trace[:-1]},{trace:s}->{trace[0] + trace[2:]}"
+    elif return_type == "weight":
+        signature = f"{trace[:-1]},{trace:s}->{trace[-1]}"
+    else:
+        raise ValueError(
```
woodsp-ibm, line 88: Since this is an internal private method ideally this should not occur right - it would be just some coding error on our part - a user should not be using this method. We could avoid this altogether by just having a bool parameter, say for_weights which could say default to False which is taken to mean a return for inputs. Would we ever want more values - one could use an Enum too rather than have to code literal strings and make mistakes.

releasenotes/notes/fix-716-mismatch-dimension-pytorch-ba01bea90eba1435.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixes the dimension mismatch error in the `torch_connector` raised when using other-than 3D datasets.
+    The updated implementation defines the Einstein summation signature dynamically based on the number of
+    dimensions `ndim` of the input data (up to 26 dimensions).
```
woodsp-ibm, line 6: Should we also note this in the TorchConnector somewhere. Maybe in its main docstring that datasets are limit to 26 dimensions and any larger with raise an error.

test/connectors/test_torch_connector.py
```diff
@@ -241,3 +242,191 @@ def test_estimator_qnn(self, num_qubits, observables):
...
+          using a 4-D input tensor (batch, channel, height, width) to ensure compatibility.
+          This test mirrors Issue https://github.com/qiskit-community/qiskit-machine-learning/issues/716
+
+        Note: Backward test is run implicitly within the context of convolutional layer execution.
+
+        pylint disable=SPELLING
+        """
+        # Test valid inputs and outputs
+        self.assertEqual(_get_einsum_signature(3, "input"), "ab,abc->ac")
+        self.assertEqual(_get_einsum_signature(3, "weight"), "ab,abc->c")
```
woodsp-ibm, line 411: Ideally unit tests test the public API not private methods explicitly though of course by exercising the public API it should cover these completely too

qiskit_machine_learning/connectors/torch_connector.py
edoaltamura: I replaced the for loop with `from string import ascii_lowercase`, and slice that string to find the trace. The CHAR_LIMIT is now defined as a constant and sits outside the function scope.

releasenotes/notes/fix-716-mismatch-dimension-pytorch-ba01bea90eba1435.yaml
edoaltamura: I agree it would be useful to mention it in the main docstring. Now added.

test/connectors/test_torch_connector.py
edoaltamura: These are covered by the examples defined in the test above, which should be comprehensive enough. So, we should be able to remove these two assertEqual tests since they are a bit redundant.

qiskit_machine_learning/connectors/torch_connector.py
```diff
@@ -52,12 +53,50 @@ class Module:  # type: ignore
...
+
+    Returns:
+        str: The Einstein summation signature.
+
+    Raises:
+        RuntimeError: If the number of dimensions exceeds the character limit.
+    """
+    if n_dimensions >= CHAR_LIMIT:
+        raise RuntimeError(
+            f"Cannot define an Einstein summation with more tha {CHAR_LIMIT:d} dimensions."
```
woodsp-ibm, line 78:
```suggestion
            f"Cannot define an Einstein summation with equal or more than {CHAR_LIMIT:d} dimensions,, got {n_dimensions}."
```
There is a spelling mistake `tha` in the message. Now you could change things to say more that CHAR_LIMIT-1 rather than what I did or adding equal or. Also adding the value we got which triggered the error is always helpful.

edoaltamura:
The Windows CI seems to be affected by this issue https://github.com/pytorch/pytorch/pull/124932, so we might need to wait for the PyTorch update to reflect in the CI before triggering the tests once again.

woodsp-ibm (approved):

PR #796, opened 2024-05-02
Pin torch==2.2.2 in CI to avoid a UTF-8 error in Windows (temporary bug in Torch)

edoaltamura:
### Summary
Pins `torch==2.2.2` in CI to avoid a UTF-8 error in Windows. This change is temporary and should be reverted to the latest Torch version as soon as https://github.com/pytorch/pytorch/issues/124897 is closed.


woodsp-ibm (approved):
Maybe you want to create an issue to revert this change once a new release of torch is done that includes the fix so as we don;t forget to do this at some point,.

PR #806, opened 2024-05-23
Include GitHub security policy

edoaltamura:
### Summary
Implements a security policy for GitHub use, similar to the one used in Qiskit: https://github.com/Qiskit/qiskit/blob/main/SECURITY.md

### Details and comments
Closes https://github.com/qiskit-community/qiskit-machine-learning/issues/791.


SECURITY.md
```diff
@@ -0,0 +1,28 @@
...
+The last version of the previous major release, `0.46.x`, is supported with bug fixes only until six months after the final release of `1.0.0`.
+
+As an additional resource, you can find more details on the release and support schedule of Qiskit in the [documentation](https://docs.quantum.ibm.com/start/install#release-schedule).
+
+## Reporting a Vulnerability
+
+To report vulnerabilities, you can privately report a potential security issue
+via the GitHub security vulnerabilities feature. This can be done here:
+
+https://github.com/Qiskit/qiskit/security/advisories
```
woodsp-ibm, line 20: This should a pointer here not to Qiskit's repo for where to report things. If its in Qiskit they should report it there but ML ones I think not.

edoaltamura: Indeed it should be https://github.com/qiskit-community/qiskit-machine-learning/security/advisories, just updated it.

SECURITY.md
```diff
@@ -0,0 +1,28 @@
+# Security Policy
```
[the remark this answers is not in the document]
edoaltamura: I have removed the third point and kept support for major releases (such as 0.x), with stable backports. What do you and others think about this proposed solution?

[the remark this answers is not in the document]
woodsp-ibm: You can see the text that Qiskit had pre-1.0 https://github.com/Qiskit/qiskit/blob/stable/0.46/SECURITY.md The statement is a lot less. Until qiskit reached 1.0 and its API was considered stable, change was much more. More like in the apps which you can see within the 0.x major version (which it has been so far) the minor versions e.g 0.7.0, 0.6.0 may deprecate stuff and introduce new features such that earlier versions may well have a completely different API eg in the changeover from using QuantumInstance to Primitives where the former was deprecated and later removed after primitive support was added.

And this
>For example, if the most recent release is `0.7.2`, then the current major release series is `0.7.x` and the current minor release is `0.7.2`.

Is more like

For example, if the most recent release is `0.7.2`, then the current major release series is `0.x` the current minor release is `0.7.x`. with `0.7.2` being the current patch release.

I could see if/when deciding things were stable enough and going to 1.0 that doing things more as qiskit is now doing makes sense. Ahead of that..... well for one the ISA change and V2 support is going to change things up a bit!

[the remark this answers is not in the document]
edoaltamura: Thanks for the pointer - I change it to make it more similar to Qiskit pre-1.0.

SECURITY.md
```diff
@@ -0,0 +1,47 @@
...
+   - The link to the filed issue stub.
+   - A description of the vulnerability and its impact.
+   - Steps to reproduce the issue or a proof-of-concept (PoC) for independent confirmation.
+   - Any potential fixes or recommendations you might have.
+
+2. **Response Time**: We will acknowledge your report within 3 business days and provide you with an estimated time frame for resolving the issue.
+
+
+### Untrusted models
+Be careful when running untrusted models. Models are programs running them is equivalent to exposing your system to potentially untrusted code[^data-poisoning-sources].
```
OkuyanBoga, line 36: Terminology here can be misleading, could be understood as malware, but it is about model security. Please make it clearer.

edoaltamura: Updated the statement about data poisoning and backdoor trigger safety as discussed.

OkuyanBoga (approved):

PR #816, opened 2024-07-05
Restore latest dependencies in CI environment

edoaltamura:
### Summary

Restores the latest dependencies for Numpy and PyTorch.

Closes https://github.com/qiskit-community/qiskit-machine-learning/issues/815
Closes https://github.com/qiskit-community/qiskit-machine-learning/issues/797
Closes https://github.com/qiskit-community/qiskit-machine-learning/issues/812

Updates the GitHub Actions workflow to skip the `mypy` job when testing on the Windows platform. The `mypy` type checker is known to have an issue with PyTorch UTF-8 and skipping this step temporarily will streamline the CI process.


.github/workflows/main.yml
```diff
@@ -146,7 +146,7 @@ jobs:
       - run: make lint
         shell: bash
       - run: make mypy
-        if: ${{ !cancelled() }}
+        if: ${{ !cancelled() && matrix.os != 'windows-latest' }}
```
woodsp-ibm, line 149: Dont we want to do this either if the run is not cancelled `or` if the os is not windows (in general the workflow will not be cancelled and even if it is we always want to skip i.e run if not cancelled so it needs to be an `or` not `and` I think)

woodsp-ibm:
Seems it skipped the mypy checks as was wanted but now hits errors in unit test.

edoaltamura:
It only appears with Windows and only with Py3.12. @OkuyanBoga suggested it could be connected to outstanding compatibility issues between Numpy 2.0 and `qiskit_aer_gpu` or `torch_gpu`. Unfortunately, I could not find anything on this topic online, and a quick local debugging didn't give useful information either. Do you have any suggestions?

woodsp-ibm:
I see there was some recent activity linked to this issue https://github.com/pytorch/pytorch/issues/78341 Locally I see that 3.8 still uses the older numpy with the latest torch and that passes ok. Does it work under 3.12 with that version of numpy, is that something you can try locally, i.e. 1.26.4 - if so we could temporarily pin it to that for windows.

edoaltamura:
I found that 
- Numpy 1.26.4 passes on Python 3.8-11, but not 3.12 (in general)
- Numpy 2.x passes on Python 3.9-12, but not 3.8
Would this suggest we need version-specific requirements to test over the Python3.8-12 range?

woodsp-ibm:
> Numpy 2.x passes on Python 3.9-12, but not 3.8

There is no 3.8 distribution of Numpy 2.0 as that version is no longer supported. 3.8 EOL is end Oct this year at which point Qiskit will no longer support it so support for that would expected to be dropped here too. What do you mean passes for 3.12 - isn't it failing here in CI for 3.12 where its using numpy 2.0 (have no idea for 3.9 thru 3.11 as we only test min/max supported versions).

> Numpy 1.26.4 passes on Python 3.8-11, but not 3.12 (in general)

What does not in general here mean. CI is failing around the numpy() call in those detach statements for 3.12 where its also using numpy 2.0. Does general mean something else or the same thing? If the CI failure is not related to numpy what is it - other versions of some other libraries somehow interfering. The newest torch is working in 3.8 with 1.26.4 in CI and you indicate its working up to 3.11 so does some library or something change for 3.12? The torch issue I linked above seemed to have recent activity linked, i.e. after the latest release.

edoaltamura:
Restricting `numpy<2.0` in `constraints.txt` seems to have produced a suitable set of dependencies for all the tests.

edoaltamura:
Also closes https://github.com/qiskit-community/qiskit-machine-learning/issues/812.

Issue https://github.com/qiskit-community/qiskit-machine-learning/issues/818 was created as a reminder to revert the changes in this PR.

Issue #824, opened 2024-08-19
Remove `py3.8` tests because of approaching EOL


edoaltamura:
### What should we add?

The Python 3.8 release will reach its end of life in October 2024, as described in [PEP 569](https://peps.python.org/pep-0569/). Therefore, the CI tests configured with a Python 3.8 environment will be removed.

A new set of CI tests running on Python 3.13 will be introduced near the end of 2024, when Qiskit Machine Learning is projected to have reached version `0.8.0` and Python 3.13 will be released with the initial bug fixes.

PR #826, opened 2024-08-19
Remove Python 3.8 from CI (#824)

edoaltamura:
### Summary
This PR removed the CI test environment using Python 3.8, which is scheduled to reach EOL as in [PEP 569](https://peps.python.org/pep-0569/).

Closes https://github.com/qiskit-community/qiskit-machine-learning/issues/824
Closes https://github.com/OkuyanBoga/hc-qiskit-machine-learning/issues/44

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,7 +35,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8]
+        python-version: [3.9]
     steps:
       - name: Print Concurrency Group
         env:
@@ -112,14 +112,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8, 3.9, '3.10', 3.11, 3.12]
+        python-version: [3.9, '3.10', 3.11, 3.12]
         include:
           - os: macos-latest
-            python-version: 3.8
+            python-version: 3.9
           - os: macos-latest
             python-version: 3.12
           - os: windows-latest
-            python-version: 3.8
+            python-version: 3.9
           - os: windows-latest
             python-version: 3.12
           # macos-14 is an Arm64 image
@@ -165,7 +165,7 @@ jobs:
         run: |
           coverage3 combine
           mv .coverage ./ci-artifact-data/ml.dat
-        if: ${{ matrix.os == 'ubuntu-latest' && matrix.python-version == 3.8 }}
+        if: ${{ matrix.os == 'ubuntu-latest' && matrix.python-version == 3.9 }}
         shell: bash
       - uses: actions/upload-artifact@v4
         with:
@@ -188,7 +188,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8, 3.12]
+        python-version: [3.9, 3.12]
     steps:
       - uses: actions/checkout@v4
         with:
@@ -251,29 +251,25 @@ jobs:
 #          cd docs/_build/html
 #          mkdir artifacts
 #          tar -zcvf artifacts/tutorials.tar.gz --exclude=./artifacts .
-#        if: ${{ matrix.python-version == 3.8 && !startsWith(github.ref, 'refs/heads/stable') && !startsWith(github.base_ref, 'stable/') }}
+#        if: ${{ matrix.python-version == 3.9 && !startsWith(github.ref, 'refs/heads/stable') && !startsWith(github.base_ref, 'stable/') }}
 #        shell: bash
 #      - name: Run upload stable tutorials
 #        uses: actions/upload-artifact@v4
 #        with:
 #          name: tutorials-stable${{ matrix.python-version }}
 #          path: docs/_build/html/artifacts/tutorials.tar.gz
-#        if: ${{ matrix.python-version == 3.8 && !startsWith(github.ref, 'refs/heads/stable') && !startsWith(github.base_ref, 'stable/') }}
+#        if: ${{ matrix.python-version == 3.9 && !startsWith(github.ref, 'refs/heads/stable') && !startsWith(github.base_ref, 'stable/') }}
   Deprecation_Messages_and_Coverage:
     needs: [Checks, MachineLearning, Tutorials]
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: [3.9]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-      - uses: actions/download-artifact@v4
-        with:
-          name: ubuntu-latest-3.8
-          path: /tmp/u38
       - uses: actions/download-artifact@v4
         with:
           name: ubuntu-latest-3.9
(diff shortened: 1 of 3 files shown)


releasenotes/notes/py38_end_of_support-fa1fdea6ea02b502.yaml
```diff
@@ -0,0 +1,5 @@
+---
+deprecations:
```
woodsp-ibm, line 2: This feels more like it should be an `upgrade` section where if the user is still using 3.8 they need to upgrade their environment to now use a newer version, at a minimum 3.9. A deprecation means the user can continue to do what they were doing before but that way will issue a warning for some alternate that works now and where the former way will get removed in the future. Here the support is gone already hence the user needs to upgrade things.

edoaltamura: I changed this in the latest commit.

edoaltamura:
@woodsp-ibm for future reference, where in the code would you recommend adding a deprecation message ahead of a Python version removal? For individual features in the code it's straightforward, but what about Python versions?

woodsp-ibm:
>... for future reference, where in the code would you recommend adding a deprecation message ahead of a Python version removal? For individual features in the code it's straightforward, but what about Python versions?

Qiskit does it in the main `__init__..py` e.g. https://github.com/Qiskit/qiskit/blob/798449d1def74dceec470601b0c6a34495074520/qiskit/__init__.py#L24-L30
 
But.... as the apps are dependent on qiskit and since it deprecates support my recollection is that we simply removed support from the app release for whatever that version was in the release of the application pkg following on from when Qiskit did. The premise being the apps are dependent on qiskit, people install the latest of that, and having more than one deprecation notice may be annoying.
 
E..g if you search for `3.7` in this repo (or in the published release notes as its single html page) you will find two upgrade renos, one for dropping 3.6 and one for 3.7 - no deprecation notices that I saw.

constraints.txt
```diff
@@ -1,4 +1,4 @@
 numpy>=1.20,<2.0
-ipython<8.13;python_version<'3.9'
+ipython<8.13;python_version<'3.10'
```
woodsp-ibm, line 2: Is this change due to bumping the version we use from 3.8 to 3.9 and we just did not have the constraint right before in that it really applied to more than just 3.8

woodsp-ibm:
On thing to note, that at the point when we want to merge this PR we should update the  branch rules such that 3.9 jobs, that replaced the prior 3.8 ones, are Required to pass and those old 3.8 ones are not any more.

While looking at the jobs, as we have had the new M1 jobs for some while, do we want to make them passing as Required for the merge too...

constraints.txt
edoaltamura: My understanding of `<'3.9'` is that `ipython` will pick `3.8`, which is the only allowed version strictly lower than `3.9`. So, to allow `3.9` I raised `<'3.9'` to `<'3.10'`. Is this correct?

.github/workflows/main.yml
```diff
@@ -251,29 +251,25 @@ jobs:
...
 #          path: docs/_build/html/artifacts/tutorials.tar.gz
-#        if: ${{ matrix.python-version == 3.8 && !startsWith(github.ref, 'refs/heads/stable') && !startsWith(github.base_ref, 'stable/') }}
+#        if: ${{ matrix.python-version == 3.9 && !startsWith(github.ref, 'refs/heads/stable') && !startsWith(github.base_ref, 'stable/') }}
   Deprecation_Messages_and_Coverage:
     needs: [Checks, MachineLearning, Tutorials]
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: [3.9]
```
edoaltamura, line 267: > On thing to note, that at the point when we want to merge this PR we should update the branch rules such that 3.9 jobs, that replaced the prior 3.8 ones, are Required to pass and those old 3.8 ones are not any more.
> 
> While looking at the jobs, as we have had the new M1 jobs for some while, do we want to make them passing as Required for the merge too...

I was expecting changing python version here and in the "needs" jobs would run them with 3.9 directly, but they appear to pick up 3.8 instead. What would you suggest in this case?

constraints.txt
woodsp-ibm: This is a constraint (limits) over what is installed when it's needed to be installed. So that basically says if ipython is installed it should be less than version < 8.13 but with the caveat that this only for when the python version its being installed into is less than that value. So before it would have installed some version of ipython < 8.13 when using 3.8 and was not constrained in other cases so it would have used whatever version it got resolved to, which is likely much newer. Now with bumping the Python version there it will be constraining it to < 8.13 in Python 3.9, which it was not before. But then I do not know which jobs exactly end up needing/using it - I imagine its for the notebook ones - which we would have run on 3.8 (and 3.12) before but not 3.9. You could try and see without that constraint.

.github/workflows/main.yml
woodsp-ibm: How the rules work is you select from the jobs that are run which are required - its configured. The config still has the old jobs, but since they are no longer run they show up as required but of course they will never pass since they are not run. What needs to happen is that the new jobs that we require to pass are selected/configured into the rule and the old ones removed. You can see on PRs that do not change the jobs that things are still correct since the jobs configured in the rules are the ones that are run. As this changes things we will need to update the rules. Its something we end up doing from time to time is jobs are changed, such as dropping a Python version or adding a new one.

constraints.txt
edoaltamura: Ok, I'll try out the tests without `ipython<8.13;python_version<'3.10'` in the next commit.

.github/workflows/main.yml
edoaltamura: I've changed the required flags in `.github/actions/run-tests/action.yml`.

woodsp-ibm: The `required` flags in that file are not the branch rules I was talking about above. These should stay as they were - they indicate for each individual task CI whether it is required to pass or not and hence whether CI should should consider it failed or not. The branch rules exist in the github settings here - maybe you can view them but I doubt you can edit to update them. For each overall CI run e.g a Python 3.9 tests (i.e. each entry you see in the Build details) it configiures Github as to need the selected set to pass in order to be able to merge the PR.

constraints.txt
edoaltamura: Just pushed this update to remove the ipython constraint.

woodsp-ibm (dismissed):
LGTM

releasenotes/notes/py38_end_of_support-fa1fdea6ea02b502.yaml
```diff
@@ -0,0 +1,6 @@
+---
+upgrade:
+  - |
+    Removed support for using Qiskit Machine Learning with Python 3.8, reflecting
```
woodsp-ibm, line 4: I approved this already, and noticed the future tense aspect here for the EOL. Bbut we can of course edit the text later when doing the release as this talks about upcoming EOL which I guess is not far off now given that we are already into October so depending on how the release pans out that. Or we could just simply remove the word "upcoming" from the text and I think it would still be good and would read fine for anyone reading the notes later too.

edoaltamura: Thanks for noticing. I updated it now to avoid forgetting.

woodsp-ibm (approved):

PR #833, opened 2024-09-25
Fix lint errors due to Pylint 3.3.0 update in CI

edoaltamura:
## Summary

This PR adds `pylint : disable` comments to fix CI errors caused by the recent update to Pylint 3.3.0. These changes are needed for tests to pass on Python 3.9 and above, where stricter rules are now in place.

## Details

With Pylint 3.3.0, new checks flag some functions for having too many positional arguments, even when many of these are used as keyword arguments. This includes errors like `too-many-positional-arguments` (R0917), which have started showing up in several places in the code.

Since these issues aren't related to the core changes in this PR, I've added inline disables to silence the new warnings. This is in line with what was done in other Qiskit repositories, such as Qiskit Algorithms https://github.com/qiskit-community/qiskit-algorithms/pull/203#issuecomment-2369128595, when faced with similar lint rule updates.

This is a temporary solution. Later on, we may look at adjusting the methods to avoid the need for these disables.
Python 3.8, which still uses Pylint 3.2.7, is unaffected and passes tests. However, Python 3.8 support will soon be dropped: https://github.com/qiskit-community/qiskit-machine-learning/pull/826

## What’s next
Once this PR is merged, the CI pipeline should run successfully for all versions, including the nightly builds on Python 3.9+. Future PRs will also pass lint checks until a more permanent solution is implemented.


OkuyanBoga (approved):
A very short term temporary fix for unit tests. Can be fixed later.

woodsp-ibm (approved):
LGTM thx

PR #841, opened 2024-10-18
Remove redundant MacOS 14 image in CI

edoaltamura:
Removes duplicated MacOS 14 Arm image in CI and now keeps only `macos-latest`. This change was introduced in https://github.com/qiskit-community/qiskit-machine-learning/pull/779.

Closes https://github.com/qiskit-community/qiskit-machine-learning/issues/801
Closes https://github.com/OkuyanBoga/hc-qiskit-machine-learning/issues/50

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -114,6 +114,7 @@ jobs:
         os: [ubuntu-latest]
         python-version: [3.9, '3.10', 3.11, 3.12]
         include:
+          # macos-latest is an Arm64 image
           - os: macos-latest
             python-version: 3.9
           - os: macos-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -122,11 +122,6 @@ jobs:
             python-version: 3.9
           - os: windows-latest
             python-version: 3.12
-          # macos-14 is an Arm64 image
-          - os: macos-14
-            python-version: '3.10'
-          - os: macos-14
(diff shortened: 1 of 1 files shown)


woodsp-ibm:
Referring to this comment https://github.com/qiskit-community/qiskit-machine-learning/issues/801#issuecomment-2115178170  in the issue this closes

I guess you decided, at least for now, just to do CI on ARM rather than switch one of the pair over to X86 - we did both X86 and ARM in the past as we had seen issues on one platform not the other. But with things changing around the image naming both became ARM versions as they were named/given in CI, but as that comment notes there is still a X86 image that has `largest` as part of its name - see https://github.com/actions/runner-images?tab=readme-ov-file so if wanted that could be used.

Anyway its always something that could be added later if/as needed/wanted so I approved this.

woodsp-ibm (dismissed):

edoaltamura:
In this case, I think it makes sense to add the `large` and continue to test with it. I'll update the yml right away.

edoaltamura:
Oups! `macos-latest-large` returns this error:
image

edoaltamura:
Interesting! I'll undo the last commit https://github.com/qiskit-community/qiskit-machine-learning/pull/841/commits/14f945e3f0c8eaf39195fed81bc6e55ce077735f and only keep `macos-latest` then.

woodsp-ibm (approved):

PR #842, opened 2024-10-18
Release `nbconvert` constraint

edoaltamura:
### Summary
Releases constraint in the `nbconvert` version due to https://github.com/jupyter/nbconvert/issues/2092, which `nbconvert` v7.16.4 may have fixed. 

### Details and comments

Closes https://github.com/qiskit-community/qiskit-machine-learning/issues/735
Closes https://github.com/OkuyanBoga/hc-qiskit-machine-learning/issues/51

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,3 +1,1 @@
 numpy>=1.20
-nbconvert<7.14  # workaround https://github.com/jupyter/nbconvert/issues/2092
-


woodsp-ibm (approved):

edoaltamura:
Before merging, I just wanted to check how Nbconvert might behave as soon as we re-instate the stable tutorials in
https://github.com/qiskit-community/qiskit-machine-learning/blob/2bc61a3968e2fbdaecfed87d518947b65424ae06/.github/workflows/main.yml#L232-L261

These were commented out temporarily when Algorithms were merged (https://github.com/qiskit-community/qiskit-machine-learning/pull/817) because of the removed dependency. What do you think @woodsp-ibm?

woodsp-ibm:
This was done to make sure that when we went from version to version there were no breaking changes - at least as far as testing with the tutorials from the last (stable) minor version was concerned. The nbconvert change you did here does revert what was done earlier. I am not sure with things in this case - to some degree its a breaking change here since now we would expect users to use the parts that did come from algorithms from ML here now instead - so if you don't install qiskit-algorithms, since ML not longer depends/needs it then those stable tutorials are not going to work right. It would seem once a release was done with this self-contained ML, ie the next version, then that would be a point after which perhaps re-enabling that would make more sense?

edoaltamura:
Indeed, we should reinstate the stable tutorials after 0.8 is released. I was wondering if the `nbconvert>7.14` install could give errors on the stable tutorials as soon as they are reactivated, but I don't see issues with the tutorials running in this round of checks. So, stable tutorials in 0.8 should also run with `nbconvert>7.14` successfully.

I'll merge this for now, and we can revisit this after the 0.8 release if `nbconvert` breaks the stable tutorials.

PR #844, opened 2024-11-11
Post release 0.8

edoaltamura:
### Summary
Changes the mergify target branch to stable 0.8, and reactivates the tests for the stable tutorials, which were turned off in https://github.com/qiskit-community/qiskit-machine-learning/pull/817 to momentarily silence breaking changes from the migration of Qiskit Algorithms.


woodsp-ibm:
You need to bump the version on main to 0.9.0 now that 0.8.0 is released in [VERSION.txt](https://github.com/qiskit-community/qiskit-machine-learning/blob/main/qiskit_machine_learning/VERSION.txt) (Bug fix releases, if done, e.g. 0.8.1 the version is bumped direct on the stable branch so this always has the next release version for main)

You can see this update was done in the Post 0.7.0 release if you look back at the closed PRs. I think this is why the stable PRS are failing as its installing.

We do not backport these post release PRs. Normally they do not include anything but bumping version. So I removed the tag. If you want to backport the workflow I would suggest removing it from this PR, doing this first, then doing the workflow in a separate PR since we need to keep the versioning as it is on stable.

edoaltamura:
Thanks @woodsp-ibm, that's clear now. I just bumped`VERSION.txt`.

woodsp-ibm (approved):

PR #849, opened 2024-11-12
Prepare to sanitize docs

edoaltamura:
### Summary
Now that mergify target is updated, update numpy version to stable 0.8..


woodsp-ibm:
I am not sure about the mergify config change. This is normally updated on main following a release and does not get on stable until the subsequent release. But I imagine it does no harm as such here since we don;t label things on stable for backporting to the prior versions (look at stable/0.7 mergify which still points to 0.6 for instance). With it modified it would evidently be backporting to itself!

As to the stable tutorial change I guess thats fine. Its not used on stable - its more to check main code against the stable tutorials to ensure no unexpected code breakage with any new changes done on main.

Really I think all what was wanted/needed was the readme and the constraint change as was backported incorrectly in that other PR.

woodsp-ibm (approved):

PR #851, opened 2024-11-13
Reducing numpy version for deploy-docs.yml to fix numpy 2.0 bug

OkuyanBoga:
### Summary
An issue is occurring during deploying docs using the latest numpy version for pytorch. Using the latest stable version before version 2.x might temporarily sort the issue.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -38,6 +38,7 @@ jobs:
         run: |
           pip install jupyter qiskit[visualization] 'torchvision<0.10.0'
           sudo apt-get install -y pandoc graphviz
+          pip install -I numpy==1.26.4
         shell: bash
       - name: Build docs
         env:


woodsp-ibm:
That may work. I had also noted this, in a Slack conversation, in regards the torchvision pinning there

From what I can see that pin was added over 3 years ago by this #145
Which notes
> Pinned torchvision in the CI for now to fix Torch Connector Tutorial failure that is holding other PRs

Maybe we should remove that version pinning. Elsewhere where it builds the html for CI it just plain installs torchvision and that seems to work i.e. it runs/builds the notebooks/html.

I see you did this direct to stable 0.8. I guess whatever is done here should be done in the in main branch too so that the next stable build done off main this will work there too. But for now I guess this is the most expedient path to ensuring it works on stable

oscar-wallis (approved):
All good but this is a temporary fix, we should log these errors in appropriate place - Approved.

PR #853, opened 2024-11-15
Update deploy-docs.yml to fix version compatibility issues.

OkuyanBoga:
### Summary
- Removed version constraint for torchvision and numpy. 
- Updated python version from 3.9 to 3.10.


oscar-wallis (approved):
Approved

PR #855, opened 2024-11-16
Cleanup and bugfix to support different primitives. (#55)

OkuyanBoga:
### Summary
V2 primitives from different dependencies such as qiskit_aer etc. do not require isa_circuits, hence, removed and cleaned up the gradients. 
- PassManager object is not a requirement for V2 primitives anymore when using gradients and state_fidelities.
- Updated tests for gradients.
- Removed unsupported functionalities from gradients.


woodsp-ibm:
My concern with this is that for a stablebackport it looks like a lot of breaking changes if you were using this code, i.e. for any existing code you may have E.g. some classes are removed, some classes the parameter order is changed e.g. because passmanager was added or changed in order. One parameter has a default added which is different but ok as it being mandatory before any code would have had to state it so that code will still work unlike the other changes I see here.

OkuyanBoga:
Hi @woodsp-ibm, I preserved the old formatting as you suggested. The changes in this PR might seem major, but, there are minor bug fixes to allow other types of V2 primitives for different gradients, and, a few fixes for docs. I think we should consider this for the stable branch if people need these changes.

PR #878, opened 2024-12-10
Fix callback compatibility for trainable_model

### LINKED ISSUE #877: Incompatibility between ADAM and trainable_model
### Environment

- **Qiskit Machine Learning version**: 0.8.1

### What is happening?

When using Adam optimizer, there is a "missing required positional argument" error when the callback function is called. 

```
callback_graph(ite, weights, obj_func_eval):
```

### What should happen?

There should be no errors using VQC.

### Any suggestions?

_No response_

OkuyanBoga:
### Summary
Fixed a compatibility issue as mentioned in #877  for different `callback` functions when it is used in trainable_model based algorithms.

Fixes #877

### DIFF
--- a/releasenotes/notes/fix-callback-trainable_model-ca52060260a3e466.yaml
+++ b/releasenotes/notes/fix-callback-trainable_model-ca52060260a3e466.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+      Fixed a compatibility issue for different `callback` functions when it is 
+      used in :class:`~qiskit_machine_learning.algorithms.trainable_model` based algorithms.
--- a/releasenotes/notes/fix-callback-trainable_model-ca52060260a3e466.yaml
+++ b/releasenotes/notes/fix-callback-trainable_model-ca52060260a3e466.yaml
@@ -1,5 +1,5 @@
 ---
 fixes:
   - |
-      Fixed a compatibility issue for different `callback` functions when it is 
+      Fixed a compatibility issue for different `callback` functions when they are 
       used in :class:`~qiskit_machine_learning.algorithms.trainable_model` based algorithms.
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,2 +1,1 @@
 numpy>=2
-qiskit<1.3
--- a/qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
+++ b/qiskit_machine_learning/algorithms/classifiers/neural_network_classifier.py
@@ -140,22 +140,75 @@ class NeuralNetworkClassifier(TrainableModel, ClassifierMixin):
         return function
 
     def predict(self, X: np.ndarray) -> np.ndarray:
-        self._check_fitted()
+        """
+        Perform classification on samples in X.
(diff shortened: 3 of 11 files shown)


woodsp-ibm:
I guess I am trying to understand the relationship of this callback to that of the optimizers. Yes it looks to be the same signature as a scipy optimizer one but otherwise this is just a callback, True if you are using SPSA and have a callback for that, since it has a different signature you could not use it for the callback here in this context. The test which is failing is using SPSA, which presumably worked before, so I am having a hard time figuring things.

OkuyanBoga:
> I guess I am trying to understand the relationship of this callback to that of the optimizers. Yes it looks to be the same signature as a scipy optimizer one but otherwise this is just a callback, True if you are using SPSA and have a callback for that, since it has a different signature you could not use it for the callback here in this context. The test which is failing is using SPSA, which presumably worked before, so I am having a hard time figuring things.

Maybe, changing optimizers such as SPSA and ADAM might be a better option

releasenotes/notes/fix-callback-trainable_model-ca52060260a3e466.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+      Fixed a compatibility issue for different `callback` functions when it is
```
edoaltamura, line 4:
```suggestion
      Fixed a compatibility issue for different `callback` functions when they are 
```

OkuyanBoga:
Hi Steve, 

I was not clear enough on the issue, maybe this might be useful:

SPSA has its own callback function since it is not based on `SciPyOptimizer` and it is called inside its own loop:
https://github.com/qiskit-community/qiskit-machine-learning/blob/94ccb0a70ab46f22c13e0eac411c5f21274b5530/qiskit_machine_learning/optimizers/spsa.py#L580-L588

But trainable_model makes calls itself, and I think it is only compatible with `SciPyOptimizer`:
https://github.com/qiskit-community/qiskit-machine-learning/blob/94ccb0a70ab46f22c13e0eac411c5f21274b5530/qiskit_machine_learning/algorithms/trainable_model.py#L270-L273

and my suggested change is:
```
  def objective(objective_weights):
      objective_value = function.objective(objective_weights)
      if isinstance(self._optimizer, SciPyOptimizer):
          self._callback(objective_weights, objective_value)
      return objective_value
```
This way we can allow custom callback functions and clear potential ambiguity when the function is called (inside or outside of the optimisation loop for custom callback functions).

Please note that SPSA is used in tutorials with [QKT](https://github.com/qiskit-community/qiskit-machine-learning/blob/main/qiskit_machine_learning/kernels/algorithms/quantum_kernel_trainer.py), not with trainable models:
https://github.com/qiskit-community/qiskit-machine-learning/blob/94ccb0a70ab46f22c13e0eac411c5f21274b5530/docs/tutorials/08_quantum_kernel_trainer.ipynb#L253

I think trainable models must be adaptive if `SciPyOptimizer` is not used.

edoaltamura (approved):

PR #896, opened 2025-03-01
Changed datasets/ad_hoc.py to work for n>3

RishiNandha:
### Summary

This is related to https://github.com/qiskit-community/qiskit-machine-learning/issues/330

I have rewritten the ad_hoc dataset to work for n>3. I have also made some of the operations more efficient.

I ran "make lint" and "make style" as suggested in the contributing guidelines, but I got an error I was not able to resolve. This is my first contribution, if there are improvements to be made in the PR, please let me know. 

I'm also working on bringing more dataset generators to the repository, so I hope ad_hoc.py can still be kept around as the simpler one in the family not that it's been rewritten

### Details and comments

1. The tensor products such as H^{(x)n} were written in n step loops, I've made them logn step loops. And similar other small optimizations such as precomputing Zi*Zj.

2. The old implementation did an n-dimensional grid for generating random phase vectors. While this brought uniformity with small number of qubits, it was too much computational overhead for larger numbers. I've have swapped this out for 2 different options of stratified sampling

3. The old implementation had labelling done with parity() and majority() in n=2 and n=3 respectively. I've used an n-qubit Z matrix directly, which aligns better with the reference paper attached in the documentation and also relieves the constraint an n being 2 or 3

4. Vectorized the labelling loop. This has given some performance gain


qiskit_machine_learning/datasets/ad_hoc.py
```diff
@@ -1,6 +1,6 @@
 # This code is part of a Qiskit project.
 #
-# (C) Copyright IBM 2018, 2024.
+# (C) Copyright IBM 2018, 2024, 2025.
```
woodsp-ibm, line 3:
```suggestion
# (C) Copyright IBM 2018, 2025.
```
The last changed date, i.e. 2024 just needs amending to 2025 rather than adding another date.

The copyright is explained here in the contributing guide https://github.com/qiskit-community/qiskit-machine-learning/blob/main/CONTRIBUTING.md in the Pull Request Checklist item point 5 (copied below)

> Ensure all code, including unit tests, has the copyright header. The copyright date will be checked by CI build. The format of the date(s) is year of creation, last year changed. So for example:
> 
>     # (C) Copyright IBM 2018, 2021.
> 
> If the year of creation is the same as last year changed then only one date is needed, for example:
> 
>     # (C) Copyright IBM 2021.
> 
> If code is changed in a file make sure the copyright includes the current year. If there is just one date and it's a prior year then add the current year as the 2nd date, otherwise simply change the 2nd date to the current year. The year of creation date is never changed.

qiskit_machine_learning/datasets/ad_hoc.py
```diff
@@ -11,29 +11,29 @@
...
 def ad_hoc_data(
-    training_size: int,
+    train_size: int,
     test_size: int,
     n: int,
     gap: int,
+    divisions: int = 0,
     plot_data: bool = False,
-    one_hot: bool = True,
+    one_hot: bool = False,
```
woodsp-ibm, line 36: In terms of the method signature any changes should be done in a backwards compatible way, i.e. if you already have code that uses this it should still work. To that end the change in name of training_size is problematic, as is the change in the one_hot default - they should really stay as they are. We do/can change things so they would break existing code but such changes are normally done via a deprecation. Having said that I have no idea if the data set produced by this new method, for values that were supported before, ends up the same or not with this new algorithm. Then the additional parameter, divisions, to maintain compatibility that should go at the end.

I guess another way to do this would be to keep the original method there as-is and add a complete new method, not quite sure of name, maybe ad_hoc_dataset, ad_hoc_data_sobol or something, that can be used instead.  The current method could then be deprecated and removed later leaving only the new one. The default values and order of the parameters in a new method would be a free choice not constrained by compatibility.

woodsp-ibm: As you can see from CI there are failures due to these changes being incompatible with existing code - in this case unit tests we have here for example where they should still work/pass unchanged,

RishiNandha: Okay, Ill fix the parameters to be backward compatible. Quick fix

About if it'll produce the same values as before. It will match in shape and format, but for the same seed it won't be the same values. Because the sampling method is different

So shall I fix the default parameters and order of arguments to support backward compatibility and edit the unit tests to assert the new expected values?

RishiNandha: About the depreciation of ad_hoc.py and the renaming of this new generator: I think we can do whatever you'd and the other reviewers would suggest. 

If we make this a separate dataset and won't have constraints to work backwards with ad_hoc codes, I can also further explore including hyperparameters such as:

1. Linear vs Circular vs Full Entanglements 
2. Labelling based on Expected Value vs Labelling based on Measurement

Reasons being:
1. When we say "Separable by ZZFeaturemap" the most readily available example of ZZFeaturemap is linear entangled while ad_hoc by default is supporting only full entanglement. So giving that option might make the inner workings of this generator more transparent

2. I'm thinking measurement would make it harder for a classical ML method to work well with while QML or VQA methods won't have a problem. If that is the case, we can demonstrate that Quantum Methods have an advantage in performing better in this dataset. But I'm not sure, I'm very new to this field of study. But I think including this as an option with one of hyperparameters can leave the choice up to the user

RishiNandha: I made a commit with the parameter order and default values fixed. Like speculated by you, the unit tests still continue to fail because the arrays generated by the new sampling technique are numerically different from the old. So I wanted to ask what you would suggest. 

Shall we proceed with modifying the unit tests? Or shall we proceed with making this an independent dataset generator?

woodsp-ibm: @edoaltamura Do you guys have a take on this as the prime maintainers now? I chipped in based on what I saw and for me I think a new method with this functionality, since it ends up different than the existing one, for the compatible inputs overlap, would be "better". Existing users code would not suddenly end up having a different outcome due to it giving back different data than before - such as we see here with the unit test failures. If we wanted the existing method could be deprecated and removed later giving users a chance to switch over to the new method and sort out whatever they need on their end in using it and the outcome. Also the author has suggested they can look at further hyperparameters more flexibly this way without being constrained.

edoaltamura: Thanks for your input @woodsp-ibm! I'll add my comments for @RishiNandha in a separate review in a moment.

qiskit_machine_learning/datasets/ad_hoc.py
```diff
@@ -84,140 +82,128 @@ def ad_hoc_data(
...
+    # Observable for labelling boundary
+    O = V.conj().T @ z_n @ V
+
+    # Loop for Data Acceptance & Regeneration
+    n_samples = training_size+test_size
+    features = np.empty((n_samples, n), dtype=float)
+    labels = np.empty(n_samples, dtype=int)
+    cur = 0
+
+    while n_samples > 0:
```
edoaltamura, line 145: Is there a particular reason why you're using a while loop instead of a for one, defining samples outside the loop. Is it because Sobol and LHC can give different `count` values every time, where `count` is not always a submultiple of `n_samples`?

qiskit_machine_learning/datasets/ad_hoc.py
```diff
@@ -84,140 +82,128 @@ def ad_hoc_data(
...
+        # Stratified Sampling for x vector
+        if divisions>0: x_vecs = _modified_LHC(n, n_samples, divisions)
```
edoaltamura, line 147: In my view one way to incorporate the back-compatible
```python
basis = algorithm_globals.random.random(
        (2**n, 2**n)
    ) + 1j * algorithm_globals.random.random((2**n, 2**n))
```
would be to introduce another function, say `_uniform_havlicek`, that implements it and is triggered in an elif logic like this. To make the current unit tests pass, then you'd select this sampling type.

If you were to go ahead with this, then I'd suggest introducing a function argument called eg `sampling_method`, that accept the Literal values "sobol", "hypercube", or "havlicek", and if the use selects "hypercube", then they must pass `divisions` too.

Then, you could raise an error in `_uniform_havlicek` if n>3, specifying that this (original) method only works for 2 and 3, or you could try to extend it to arbitrary n in a way that it reduces to the original form when n=2, 3.

qiskit_machine_learning/datasets/ad_hoc.py
```diff
@@ -230,41 +216,145 @@ def _plot_ad_hoc_data(x_total, y_total, training_size):
...
+        i (int): Dimension index.
+
+    Returns:
+        np.ndarray: Computed φ_i values.
+    """
+    return x_vecs[:, i].reshape((-1, 1))
+
+
+def _phi_ij(x_vecs: np.ndarray, i: int, j: int) -> np.ndarray:
+    """Compute the φ_ij term for given dimensions.
```
edoaltamura, line 347: Make sure you whitelist symbols like \phi and \Delta (somewhere else) in the dictionary used by Pylint: https://github.com/qiskit-community/qiskit-machine-learning/blob/main/.pylintdict.

This will avoid spell check errors to be raised when running GH actions workflows. You can find more info at https://github.com/qiskit-community/qiskit-machine-learning/blob/main/CONTRIBUTING.md.
The same goes for the code formatting with the commands `pip install black; make black` - this will reformat the code to match the repo standards.

edoaltamura (changes requested):
Finally, you can add a release note (reno), as in the contributing guide, describing the extension of the dataset to higher n. This description will then be included in the v.0.8.3 release notes.

qiskit_machine_learning/datasets/ad_hoc.py
RishiNandha: I've read the other comments. I'll incorporate them soon. About the while loop, so since we only want to accept samples with labels 1 or -1 (that is the |exp_val|>gap), we might end up with lesser samples accepted in the first pass than n_samples

RishiNandha:
I've made the edits to include both the old and new sampling methods. I've also included measurement based labelling and linear/circular/full entanglement as options too. I'm yet to add the whitelists to pylint, I'll do it once the unittests are fixed too.

The unittests will still fail because there is a unit test that checked if giving an n>3 raises a value error. @rogue-infinity is collaborating with me on the unittests. I think once that is done, it'll be good to go

PR #900, opened 2025-03-14
Modified test_qsvr.py and test_fidelity_quantum_kernel_qsvr.py files for regression tasks

### LINKED ISSUE #434: Improve QSVR unit tests

rockywick:
### Summary
This is regarding the issue #434

### Details and comments
I have replaced the dataset used to a regression dataset and have modified the code to work for regression testing.

### Result
Closes #434


edoaltamura (approved):
Thanks for the PR @rockywick! The code looks good to go. Let's keep the discussion about the two unit tests being almost identical open in the issue though.

PR #902, opened 2025-03-16
Adhoc now supports n>3, and new arguments

RishiNandha:
### Summary

My previous pull request became messy due to bad version control from my side. So I've created a new one

This is related to https://github.com/qiskit-community/qiskit-machine-learning/issues/330

I have rewritten the ad_hoc dataset to work for n>3. I have also made some of the operations more efficient.

I'm also working on bringing more dataset generators to the repository, so I hope ad_hoc.py can still be kept around as the simpler one in the family now that it's been rewritten

### Details and comments

1. The tensor products such as H^{(x)n} were written in n step loops, I've made them logn step loops. And similar other small optimizations such as precomputing Zi*Zj. This step also supports 3 types of entangelement now: linear/circular/full

2. The old implementation did an n-dimensional grid for generating random phase vectors. While this brought uniformity with small number of qubits, it was too much computational overhead for larger numbers. I've have added 2 more options of stratified sampling. The old method has been retained as the "grid" method as suggested so that old user codes don't behave differently now. The two other methods included are "hypercube" and "sobol". Docstring has been updated and can be referred to for more information on the methods

3. The old implementation had labelling done with parity() and majority() in n=2 and n=3 respectively. I've used an n-qubit Z matrix directly, which aligns better with the reference paper attached in the documentation and also relieves the constraint an n being 2 or 3.

4. Vectorized the labelling loop. This has given some performance gain. I have also given two options: the same "expectation" based labelling has been retained but also a "measurement" based labelling has been added

5. I've verified lint. @rogue-infinity has developed unittests for this revamped ad hoc dataset. They have been verified to pass

### DIFF
--- a/qiskit_machine_learning/datasets/ad_hoc.py
+++ b/qiskit_machine_learning/datasets/ad_hoc.py
@@ -17,7 +17,6 @@ from __future__ import annotations
 
 import warnings
 import itertools as it
-from typing import Tuple
 
 import numpy as np
 from scipy.stats.qmc import Sobol
@@ -41,8 +40,8 @@ def ad_hoc_data(
     labelling_method: str = "expectation",
     class_labels: list | None = None,
 ) -> (
-    Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]
-    | Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]
+    tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]
+    | tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]
 ):
     r"""
     Generates a dataset that can be fully separated by
--- a/qiskit_machine_learning/datasets/ad_hoc.py
+++ b/qiskit_machine_learning/datasets/ad_hoc.py
@@ -1,6 +1,6 @@
 # This code is part of a Qiskit project.
 #
-# (C) Copyright IBM 2018, 2024.
+# (C) Copyright IBM 2018, 2025.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
@@ -11,16 +11,16 @@
 # that they have been altered from the originals.
 
 """
-ad hoc dataset
+Ad Hoc Dataset
 """
 from __future__ import annotations
 
-from functools import reduce
+import warnings
 import itertools as it
-from typing import Tuple, Dict, List
-import numpy as np
-from sklearn import preprocessing
+from typing import Tuple
 
+import numpy as np
+from scipy.stats.qmc import Sobol
 from qiskit.utils import optionals
 
 from ..utils import algorithm_globals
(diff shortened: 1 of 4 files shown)


RishiNandha:
@edoaltamura I hope we can continue the discussion/review here. We had some version conflicts in the way we were collaborating and got a lot of messy commits on the previous Pull Request. So I've closed it and started a fresh one here. I've updated the summary above according to the latest changes too

qiskit_machine_learning/datasets/ad_hoc.py
```diff
@@ -31,240 +31,618 @@ def ad_hoc_data(
...
-            grid in addition to training and testing samples.
+    Parameters
+    ----------
+    training_size : int
+        Number of training samples **per class**.
+    test_size : int
+        Number of testing samples **per class**.
+    n : int
+        Number of qubits (dimension of the feature space).
+    gap : int, optional
```
edoaltamura, line 108:
```suggestion
    gap : int
```
'optional' can be misleading here. It's normally used when you have the syntax `gap: int | None  = None` in the declaration, and then something like `if not gap: gap = 0` in the function scope. Here, I'd just say that it defaults to 0 if not specified. Remember that this is an API change from the previous version, because `gap` goes from being an arg to a kwarg, so make sure you explain it in the release note.

edoaltamura: Same for the 'optional' in the other kwargs.

qiskit_machine_learning/datasets/ad_hoc.py
```diff
@@ -31,240 +31,618 @@ def ad_hoc_data(
...
+    if class_labels is None:
+        class_labels = [0, 1]
+
+    # Errors
+    if training_size < 0:
+        raise ValueError("Training size can't be less than 0")
+    if test_size < 0:
+        raise ValueError("Test size can't be less than 0")
+    if n <= 0:
+        raise ValueError("Number of qubits can't be less than 0")
```
edoaltamura, line 177:
```suggestion
        raise ValueError("Number of qubits can't be less than 1")
```
Because you correctly impose `n <= 0`.

edoaltamura (changes requested):
Great work @RishiNandha! I've made a few (very small) suggestions. To avoid the syncing trouble you experienced when in-PR commits are merged, make sure you git-pull your local repo before working on it so that incoming changes are merged with your local copy of the code.
The main change needed is in the syntax of the docstring in the ad hoc main function - it should use `Args:` like in the rest of the code so that the documentation builder can understand it and translate it to static HTML pages.

qiskit_machine_learning/datasets/ad_hoc.py
```diff
@@ -31,240 +31,597 @@ def ad_hoc_data(
...
-            if sample_total[draws] == label:
-                sample_list.append([xvals[d] for d in draws])
+    x_train = np.concatenate((a_features[:training_size], b_features[:training_size]), axis=0)
+    x_test = np.concatenate((a_features[training_size:], b_features[training_size:]), axis=0)
+    if one_hot:
+        y_train = np.array([[1, 0]] * training_size + [[0, 1]] * training_size)
+        y_test = np.array([[1, 0]] * test_size + [[0, 1]] * test_size)
+    else:
+        y_train = np.array([class_labels[0]] * training_size + [class_labels[1]] * training_size)
+        y_test = np.array([class_labels[0]] * test_size + [class_labels[1]] * test_size)
```
edoaltamura, line 267: Can I ask why you chose to compute the labels manually instead of using Sklearn as here?
```python
if one_hot:
        encoder = preprocessing.OneHotEncoder()
        encoder.fit(np.array(class_labels).reshape(-1, 1))
        labels = encoder.transform(np.array(raw_labels).reshape(-1, 1))
        if not isinstance(labels, np.ndarray):
            labels = np.array(labels.todense())
    else:
        encoder = preprocessing.LabelEncoder()
        encoder.fit(np.array(class_labels))
        labels = encoder.transform(np.array(raw_labels))
```
Your code works for repeating fixed patterns (either one-hot lists or raw class labels) based only on the training/test sizes. If you use `OneHotEncoder` or `LabelEncoder` though you can transform an arbitrary list of raw labels dynamically from the input data. 
CC @OkuyanBoga

qiskit_machine_learning/datasets/ad_hoc.py
```diff
@@ -11,16 +11,16 @@
...
 """
 from __future__ import annotations
 
-from functools import reduce
+import warnings
 import itertools as it
-from typing import Tuple, Dict, List
-import numpy as np
-from sklearn import preprocessing
+from typing import Tuple
```
edoaltamura, line 20: In the same way you define the `list` type in
```python
class_labels: list | None = None,
```
you can also use `tuple` recognised directly from Python core (since Python 3.9, see [PEP 585](https://peps.python.org/pep-0585/)), so you can avoid the import from `typing`. 
```suggestion
```
Then remember to change type hints like
```python
-> (
    Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]
    | Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]
```
into
```python
-> (
    tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]
    | tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]
```

edoaltamura (changes requested):

qiskit_machine_learning/datasets/ad_hoc.py
RishiNandha: Because the .fit() is not necessary here since the generator gives equal number of datapoints in each class. So the case of transforming an arbitrary list of raw labels never arises. 

And the grid sampling and loop sampling methods also are returning a_features and b_features which are features for classes A and B.  They are not returning (features, labels)

edoaltamura:
Re-running the MacOS job, which seems affected by #903.

edoaltamura (approved):

Issue #897, opened 2025-03-05
Migration to qiskit 2.0 and integration testing


ElePT:
### What should we add?

The Qiskit SDK CI includes an integration testing step of downstream packages through https://github.com/Qiskit/qiskit-neko, among which we include `qiskit-machine-learning`. This allows us to see that the removal of deprecated primitive V1 implementations (https://github.com/Qiskit/qiskit/pull/13877) affects the `qiskit-machine-learning` -package, and that it will need an update to stay compatible with Qiskit 2.0. 

You are free to decide if you want to migrate to 2.0 and when, but we recommend migrating for increased compatibility and support. On the qiskit-neko side,`qiskit-machine-learning` integration tests will be skipped until the package is updated. The first Qiskit 2.0 release candidate is planned for tomorrow, and the release is planned for ~March 20th.

Thank you for your effort maintaining Qiskit ML :) Let me know if you have any questions or need migration support.

edoaltamura:
Thanks for keeping us in the loop @ElePT! We'd be very happy to give Neko a go when adding support for Qiskit 2.0. For us, this will likely be a 2-step process - first deprecating V1 primitives (and Qiskit v.0.x support), and then updating to 2.0 - and we'll get back to this at the right time. Really looking forward to seeing the 2.0 release out!

woodsp-ibm:
As an FYI I would expect ML here to fail when/if Qiskit 2.0 is installed with it. I was just updating Algorithms to get CI running and you can see in the PR here [qiskit-algorithms#221](https://github.com/qiskit-community/qiskit-algorithms/pull/221) that the tests against qiskit main fail to even start, as the class it uses/depends on has now been deleted.

```
 File "/home/runner/work/qiskit-algorithms/qiskit-algorithms/qiskit_algorithms/optimizers/__init__.py", line 139, in <module>
    from .qnspsa import QNSPSA
  File "/home/runner/work/qiskit-algorithms/qiskit-algorithms/qiskit_algorithms/optimizers/qnspsa.py", line 23, in <module>
    from qiskit.primitives import BaseSampler
ImportError: cannot import name 'BaseSampler' from 'qiskit.primitives' (/opt/hostedtoolcache/Python/3.9.21/x64/lib/python3.9/site-packages/qiskit/primitives/__init__.py)
```

[...]

PR #904, opened 2025-03-18
Pin Qiskit to `<2` in next release

edoaltamura:
### Summary
Pinning Qiskit<2 will avoid breaking changes to affect current installations. This potential effect was raised in #897. The constraint can be removed when QML achieves full compatibility with Qiskit 2.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,2 +1,1 @@
-numpy>=2
-qiskit<2.0
+
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,1 +1,2 @@
 numpy>=2
+qiskit<2.0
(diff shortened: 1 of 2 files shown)


woodsp-ibm:
I am wondering - in the requirements.txt already exists a version limit for qiskit
```
qiskit>=1.0
scipy>=1.4
numpy>=1.17
scikit-learn>=1.2
setuptools>=40.1
dill>=0.3.4
```
Would it be better to keep the version restrictions in one place especially since its a direct dependent i.e. `qiskit>=1.0,<2.0`. I also notice that there is a version limiting on numpy while the constraints.txt has something different.

woodsp-ibm (approved):
Just a comment to this:

> The constraint can be removed when QML achieves full compatibility with Qiskit 2.

Before this it has a constraint >=1.0 so that rather than being removed (depending in what removed is exactly) the upper bound could be removed but that's likely then to become the lower bound i.e. >=2.0 I would think.

PR #836, opened 2024-10-03
Add support for Python 3.13

edoaltamura:
This update introduces support for Python 3.13, scheduled for final release on 7 October 2024.

Closes https://github.com/qiskit-community/qiskit-machine-learning/issues/825
Closes https://github.com/OkuyanBoga/hc-qiskit-machine-learning/issues/45

PR #910, opened 2025-04-16
Revert "Add support for Python 3.13"

edoaltamura:
Reverts qiskit-community/qiskit-machine-learning#836

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -112,17 +112,17 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.9, '3.10', 3.11, 3.12, 3.13]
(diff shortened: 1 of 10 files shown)


ksk0629:
Hello, just out of my curiosity, why did you revert the support?

woodsp-ibm:
Is there a reason for reverting this 3.13 support?

edoaltamura:
Hi! The release note updates in #836 generated a clash in the git history which is causing the CI tests to fail. This doesn't have to do with the python 3.13 support, which appears to work well. However, the CI can only be unblocked by reverting the last commit or hard-resetting the head to a previous state. Then, the python 3.13 support can be reinstated without the reno updates. 
@woodsp-ibm do you know anyone who could help with the hard reset? It seems I don't have the privileges to do it.

Issue #903, opened 2025-03-17
Random CI failures


woodsp-ibm:
CI had a seeming random failure in #902 where re-running the failed job passed. In looking at the nightly CI Actions I see a few failures, such as below. The 3rd item below is what failed in #902.

Investigating if the code can be improved whereby the possibility of failure is understood and updating either the unit test and/or the mainline code would be good to avoid these,

```
test.gradients.test_estimator_gradient.TestEstimatorGradientV2.test_gradient_u_2_LinCombEstimatorGradient
---------------------------------------------------------------------------------------------------------

Captured traceback:
~~~~~~~~~~~~~~~~~~~
    Traceback (most recent call last):

      File "/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/ddt.py", line 221, in wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^

      File "/home/runner/work/qiskit-machine-learning/qiskit-machine-learning/test/gradients/test_estimator_gradient.py", line 540, in test_gradient_u
    self.assertAlmostEqual(value, correct_results[i][j], 1)

      File "/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/unittest/case.py", line 939, in assertAlmostEqual
    raise self.failureException(msg)

    AssertionError: np.float64(0.05078125) != 0.0 within 1 places (np.float64(0.05078125) difference)
```

```
test.algorithms.inference.test_qbayesian.TestQBayesianInference.test_trivial_circuit_V2
---------------------------------------------------------------------------------------

Captured traceback:
~~~~~~~~~~~~~~~~~~~
    Traceback (most recent call last):

      File "D:\a\qiskit-machine-learning\qiskit-machine-learning\test\algorithms\inference\test_qbayesian.py", line 243, in test_trivial_circuit_V2
    self.assertTrue(

      File "C:\hostedtoolcache\windows\Python\3.12.9\x64\Lib\unittest\case.py", line 727, in assertTrue
    raise self.failureException(msg)

    AssertionError: np.False_ is not true
```

```
test.gradients.test_estimator_gradient.TestEstimatorGradientV2.test_gradient_parameter_coefficient_2_LinCombEstimatorGradient
-----------------------------------------------------------------------------------------------------------------------------

Captured traceback:
~~~~~~~~~~~~~~~~~~~
    Traceback (most recent call last):

[...]

PR #915, opened 2025-05-09
Added Entanglement Concentration Dataset for 3 and 4 qubits for Benchmarking Binary Classifiers

RishiNandha:
### Summary

Classification dataset for 3 and 4 qubits based on the concentration of entanglement (CE) in Quantum States. Two pre-trained circuits are used to generate states of a given amount of CE. Users can use this dataset to benchmark their binary classification pipelines.

Pre-trained weights courtesy to https://github.com/LSchatzki/NTangled_Datasets. The CE values claimed in the above repository had a mismatch for 8 qubits, hence we've left other number of qubits for future development

I've verified mypy, spell, lint and black. For some reason, make html breaks the other modules. Need some help with resolving that

### Details and comments

- We have made the order and default values of parameters match the existing ad hoc data generator for consistency. Hence equal number of datapoints in each class are generated. 
- There are two sampling options: the input states given to the circuit before it's action can either be sampled by setting each qubit's state as one of the axes of the bloch sphere ("cardinal") or can be sampled randomly ("isotropic")
- Each qubit has an easy and a hard mode. Easy has a larger difference in CE values than hard. This is to make benchmarking of pipelines more standardizable. Easy can be used to verify the working of algorithms, while hard can be used to test the maximum the algorithm can achieve. 
- There are two formatting options. The x_train and x_test can either be a numpy array or a list of quantum states
- Reference for confirming the relevance of such a dataset: https://arxiv.org/abs/2109.03400. Authors have shown that QCNNs can learn from these datasets effectively


RishiNandha:
Not sure why the 3.9 tests are passing but the 3.11 and 3.12 ones are failing. And it seems like the routine is raising errors majorly only on files I've left untouched. Any inputs of what might be happening?

woodsp-ibm:
If you look at Actions tab at the top of the page i.e. here  https://github.com/qiskit-community/qiskit-machine-learning/actions you will see Machine Learning Tests that are Scheduled. These are the same tests but are run nightly so any changes in dependents that may cause problems/failures etc are caught. The scheduled tests have been failing for a while and the main branch code needs updating in some way (maybe pinning to an earlier dependent or changing code to suit etc) so things pass again. With that done, and then merged with the code in your PR, it would then only be changes done by your PR that could cause failures - but with the base (i.e. main) failing its not.

woodsp-ibm:
The CI issues have been fixed so I updated the branch (via the update button that was here) so any issues now would be just down to this PR - unless it has a random failure for which there is an issue #903 around that,

qiskit_machine_learning/datasets/entanglement_concentration.py
```diff
@@ -0,0 +1,361 @@
...
+
+from ..utils import algorithm_globals
+
+
+# pylint: disable=too-many-positional-arguments
+def entanglement_concentration_data(
+    training_size: int,
+    test_size: int,
+    n: int,
+    mode: str = "easy",
```
woodsp-ibm, line 35: My take, instead of suppressing the warning, would be to add `*,` after the `n` parameter which would allow the first 3 arguments to be positional but require the following, which do have defaults so they do not need to be provided, be provided when doing so as keyword arguments only.

woodsp-ibm:
The equivalent tests that are run as nightly actions are failing the same tutorials tests under 3.13 so I raised #969 for this. It appears to have come about in scikit-learn 1.7.1 (which does not get installed for the same tests under 3.9, which pass, as it does not support 3,9 so an earlier version that does gets used).

PR #918, opened 2025-05-13
VQE-based Quantum Data Generator and Hamiltonian model suite

rogue-infinity:
Summary

This is a Quantum Data Generator for a variety of Hamiltonians defined in literature https://arxiv.org/abs/2408.12739. We utilize VQE to estimate the ground state energies and sample a range of parameters to define the Phase of matter

Details and comments

**Hamiltonians**

- Introduced HamiltonianModel base class defining get_hamiltonian() and sample_parameters() interfaces
- Each model includes a .sample_parameters() factory and .get_phase() phase‐labeling method for dataset annotation

**VQE**
- Implemented VQE from scratch for estimating the ground state energy of these hamiltonians to be used for the dataset generation and subsequent usage to benchmark QNNs


rogue-infinity:
Will commit the unit tests shortly

woodsp-ibm:
The CI issues have been fixed by #919 but it seems that that is no update button provided here just the message about it being out-of-date) - maybe your branch has some protection that disallows this (another 2 PRs here the update button was present). As such it seems you will have to manually merge the new changes from main into your PR branch so as CI will not fail because of those pre-existing issues.

PR #925, opened 2025-05-15
Attempt to remedy random CI failures

woodsp-ibm:
### Summary

#903 lists CI failures that are seen randomly to occur here. I tried running the tests locally a few times and had a couple of failures. After changing the code I did not see any failures. Trying again though on the unaltered code I could not seem to get any failures again.

As such I am not sure the extent, if any, that this will remedy the situation. I added the simuator seed, via options, just in case one of the others, which was not set before, was what ends up getting used so hopefully the extra seeds help. The tests still pass with them set and so they seem to do no harm as such even if they end up not really being used. But lets see how things go. I did not mark this PR to close the random error one i.e. #903 but rather figured, if this seems ok and gets merged, to see how CI goes over time and if we see no re-occurence then close. If I see any random failure as this goes through CI then I think I will close it as clearly it would not be remedying things effectively.

### Details and comments

Note: EstimatorOptions does have a seed_estimator and I had initially set that too. But I removed it as it generated a warning when running the tests that it was not used with the local simulator.

### DIFF
--- a/test/algorithms/inference/test_qbayesian.py
+++ b/test/algorithms/inference/test_qbayesian.py
@@ -1,6 +1,6 @@
 # This code is part of a Qiskit project.
 #
-# (C) Copyright IBM 2023, 2024.
+# (C) Copyright IBM 2023, 2025.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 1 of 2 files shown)


edoaltamura (approved):
Let's see how the CI runs go over the next few days with the seed fixed. I don't remember it making a big difference when I tried in the past, but I had fixed the seed in `.run()` rather.
It'd be great if a new noiseless default for `GenericBackendV2` isn't needed (see https://github.com/qiskit-community/qiskit-machine-learning/issues/903#issuecomment-2731096336).

PR #927, opened 2025-05-19
Constrain qiskit versions in tox.ini

edoaltamura:
### Summary
This change constrains the Qiskit and Qiskit-Aer versions in `tox.ini` for local testing to reflect the constraints in requirements.

### Details and comments
Fixes https://github.com/qiskit-community/qiskit-machine-learning/issues/912

Note that in local testing, several warnings about primitives deprecations are produced. This is expected as we still support primitives V1 until the hard deprecation with Qiskit 2.0.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,9 +13,7 @@ setenv =
   LANGUAGE=en_US
   LC_ALL=en_US.utf-8
   ARGS="-V"
-deps = qiskit>=1.0,<2.0
-       qiskit-aer>=0.11.2
-       -r{toxinidir}/requirements.txt
+deps = -r{toxinidir}/requirements.txt
        -r{toxinidir}/requirements-dev.txt
        torch
        sparse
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -13,8 +13,8 @@ setenv =
   LANGUAGE=en_US
   LC_ALL=en_US.utf-8
   ARGS="-V"
-deps = git+https://github.com/Qiskit/qiskit.git
-       git+https://github.com/Qiskit/qiskit-aer.git
+deps = qiskit>=1.0,<2.0
```
woodsp-ibm, line 16: I guess this is fine. My recollection was that this was used to test with the dev versions long ago, hence what was there and first so by the time the requirements and requirements-dev deps are installed next they would already be there so the versions from there would not be needed as such. Since these are now the same as in the requirements(-dev) files it seems that they could be dropped completely such that only those two files are needed to be updated for versions going forwards

edoaltamura: Right! Just removed these 2 lines.

woodsp-ibm (approved):

PR #928, opened 2025-05-19
Update qnspsa.py's deprecated BaseSampler with BaseSamplerV2

visheshrwl:
### Summary
Changed the import statement for the BaseSampler from the deprecated to the BaseSamplerV2.

### Details and comments
No Details needed.

### DIFF
--- a/qiskit_machine_learning/optimizers/qnspsa.py
+++ b/qiskit_machine_learning/optimizers/qnspsa.py
@@ -20,7 +20,7 @@ from typing import Any, Callable
 import numpy as np
 from qiskit.circuit import QuantumCircuit
 
-from qiskit.primitives import BaseSampler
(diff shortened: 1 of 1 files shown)


edoaltamura:
Thanks for the PR. Note that BaseSampler (or its V2 equivalent) is not used explicitly anywhere in that optimiser. Changing the import would change one optional type hint, i.e. the *recommendation of the default* class to use.

For this version and the next (0.8.3), we plan on keeping V1 primitives as the default and changing the default to V2 from 0.9.0 onwards. The edit you suggested will need to be applied at that moment, along with harmonising this change of defaults across the library.

Note however that the current version **does** support V2 primitives: in this optimiser, you'd need to initialise SamplerV2 and then pass it as an argument. Doing so will silence the deprecation warnings (and give you the option to use quantum hardware).

PR #933, opened 2025-05-21
Update intersphinx mapping and other URLs pointing to IQP Classic

arnaucasau:
The documentation source of truth is moving to https://quantum.cloud.ibm.com/.


woodsp-ibm:
Thanks for updating the links. We'll have to deal with the old primitive references separately that either way now end up in Page Not Exists since they do not exist in latest 2.0 version.

Its failing the Checks here as for the Python .py files it does a copyright check to make sure its updated when the file itself is updated so it reflects the right years. So conf.py needs a small correction as below

> Wrong Copyright Year:'docs/conf.py':  Current:'# (C) Copyright IBM 2021, 2024.' Correct:'# (C) Copyright IBM 2021, 2025.'

arnaucasau:
I see! I'll update all the py files. Thanks Steve!

woodsp-ibm (approved):
LGTM. Thanks for updating these links.

edoaltamura (dismissed):

arnaucasau:
Hi @edoaltamura, the link you mentioned 404s because `qiskit.primitives.BaseEstimator` was removed in qiskit 2.0. In that page, we link to the migration guide for more information about it.

You can also see that we were getting the same 404 in the old platform until now, so this PR should preserve the status quo:

previous link -> https://docs.quantum.ibm.com/api/qiskit/qiskit.primitives.BaseEstimator
new link -> https://quantum.cloud.ibm.com/docs/api/qiskit/qiskit.primitives.BaseEstimator

The new platform contains exactly the same API docs than the old one, so you could still visit the `BaseEstimator` page by going to the docs of any version other than the latest. [This](https://quantum.cloud.ibm.com/docs/en/api/qiskit/1.4/qiskit.primitives.BaseEstimator) is the link for qiskit 1.4.

I think it'd be good to update the links in this repo to avoid all the 404s, but I wasn't sure where they should point given that in the requirements, qiskit is pinned to <2.0. Should we use the 1.4 ones?

woodsp-ibm:
I made the observation above

> Thanks for updating the links. We'll have to deal with the old primitive references separately that either way now end up in Page Not Exists since they do not exist in latest 2.0 version.

My thinking being that when V1 primitives are removed from main for the 2.0 support all the text/links around those V1 primitives will need to change.

Now presently the the V1 classes still exist but the failing urls would need to add a specific version in them now to get to docs where they still exist e.g.

>  https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.primitives.BaseSampler
>  becomes 
> https://quantum.cloud.ibm.com/docs/en/api/qiskit/1.4/qiskit.primitives.BaseSampler

So the change is relatively small but its only needed on a subset of the changed links here. The docs are there by version so other than going to the latest by default one has to pick a specific version. As far as I know the docs are done similarly as they are from here - rewriting would mean changing all the docs bits in all code. The main difference with qiskit is that is keeps the prior versions when it publishes a new version of docs - which given where the docs are published here was not something that was done, rather just the latest exists.

As to changing the urls instead of leaving them broken as they have been - it could be done in this PR since arguably main still supports the V1 primitives here - if the author is open to doing that.

woodsp-ibm:
> I see an update in a reno; this shouldn't lead to the same issue we recently experienced because the PR points to main - correct?

I am pretty sure we have done updates to renos in the past - so hopefully not! I merged a similar links change PR to Qiskit Algorithms yesterday and its nightly CI last night afterwards went through ok.

woodsp-ibm (approved):
LGTM. Thanks for updating the links.

PR #945, opened 2025-06-05
Deprecate BlueprintCircuit based circuit usage

### LINKED ISSUE #931: Qiskit Circuit Library deprecations
In issue #897 around Qiskit 2.0 support I wrote a comment that there are also circuit library deprecations that should be addressed https://github.com/qiskit-community/qiskit-machine-learning/issues/897#issuecomment-2887369112

Qiskit ML both uses library circuits like this and also has its own library with BlueprintCircuit based circuits (QNNCircuit and RawFeatureVector). Now BlueprintCircuit is not yet deprecated (is expected to be) but classes in Qiskit derived from this that are used here, such as ZZFeatureMap have already been deprecated since Qiskit 1.3 and their replacement is a function creating an equivalent circuit. Logically it makes sense to do the same with circuit library instances in ML. Now I will note that an aspect of BlueprintCircuit was it allowed a circuit template/design to be specified but without knowing the number of qubits at that point. I.e you might have a high level application input that gets mapped down and without running the mapping the number is not known, but nevertheless an appropriate circuit template for the problem can still be chosen with the circuit being actually built later when known. Now I do not think that aspect is used in ML here - the functions require the number of qubits to be given so the circuit is created but I think this should therefore not be an issue.

So this has two aspects;

1. Rework ML code/test/tutorials etc using BlueprintCircuit based Qiskit circuit library classes, such as ZZFeatureMap etc, to use the replacement functions.
2. Update the ML circuit library classes, and their usage here, similarly.

woodsp-ibm:
### Summary

Closes #931

Move away from BlueprintCircuit based circuit usage

**_Todo_**:

- [ ] Add a release note
- [ ] Update notebooks

I figured I'd check in the code changes ahead of doing the above just in case any alterations to what was done were wanted.

### Details and comments

This updates machine learning to move away from BlueprintCircuit based classes and use instead the circuit library functions which create similar circuits but which cannot be updated the same way once created e.g num_qubits is fixed at creation time.

In ML here were `RawFeatureVector` and `QNNCircuit` both of which are subclasses of BlueprintCircuit, These I deprecated but they can still be used and algorithms still accept the. Their "replacements" are `raw_feature_vector` and `qnn_circuit` methods which cannot be updated for num_bits etc similarly now to the qiskit circuit library functions. 

raw_feature_vector just returns a circuit - RawFeatureVector had a num_qubits getter, which by virtue of being a QusntumCircuit it still has, but the getter to get feature_dimension no longer has any way to specifically get this but its easy to compute as its 2 ** num_qhbuts.

qnn_circuit, this returns a circuit and the parameters for the underlying feature map and ansatz so they can be passed into a QNN, eg SamplerQNN directly. It performs the compose but also creates circuits based on number of qubits etc as before so since it was a bit more that a simple composition function I kept it. You cannot get at the underlying anstaz or feature map circuits like you could with the QNNCircuit getters and while I could have passed those back as well I kept it to what the xxxQNN needs which is th circuit and which parameters are used for input and weights. If someone were to need those for any reason they could make their own composition like in fact the example in the docs for SamplerQNN & EstimatorQNN show.

The other aspect I updated is the util method `derive_num_qubits_feature_map_ansatz` in adjust_num_qubits.py. Since it seems to be a public method I altered it so it does either BlueprintCircuit based circuits like it did before, the default, or if you set a new argument to True then it uses the library function equivalents,

Evidently any BluePrintCircuit based/using code will need to be removed later but for now it all should work as it did before but code will use the functions where it did not seem to be a compatibility concern as such.

### DIFF
--- a/qiskit_machine_learning/kernels/base_kernel.py
+++ b/qiskit_machine_learning/kernels/base_kernel.py
@@ -62,7 +62,7 @@ class BaseKernel(ABC):
                 msg="Passing None as a feature_map is deprecated",
                 version="0.9.0",
                 remedy="Pass a feature map with the required number of qubits to match "
-                "the features. Later any adjustment of the number of qubits will be "
+                "the features. Adjusting the number of qubits after instantiation will be "
                 "removed from Qiskit as circuits based on BlueprintCircuit, "
                 "like ZZFeatureMap to which this defaults, which could do this, "
                 "have been deprecated.",


qiskit_machine_learning/circuit/library/raw_feature_vector.py
```diff
@@ -81,6 +149,15 @@ def __init__(self, feature_dimension: Optional[int]) -> None:
         """
         super().__init__()
 
+        issue_deprecation_msg(
+            msg="RawFeatureVector, a BlueprintCircuit based class, is deprecated",
+            version="0.9.0",
+            remedy="Use raw_feature_vector instead but note that later "
+            "adjustment of the feature dimension (i,e updating num qubits) is not "
```
edoaltamura, line 156:
```suggestion
            "adjustment of the feature dimension (i,e updating num_qubits) is not "
```

qiskit_machine_learning/kernels/base_kernel.py
```diff
@@ -54,6 +56,18 @@ def __init__(self, *, feature_map: QuantumCircuit = None, enforce_psd: bool = Tr
                 Default ``True``.
         """
         if feature_map is None:
+            # Note: when removing None it should be done in all the derived classes as well
+            # along with an appropriate update to the docstring in each case
+            issue_deprecation_msg(
+                msg="Passing None as a feature_map is deprecated",
+                version="0.9.0",
+                remedy="Pass a feature map with the required number of qubits to match "
+                "the features. Later any adjustment of the number of qubits will be "
```
edoaltamura, line 65:
```suggestion
                "the features. Adjusting the number of qubits after instantiation will be "
```

qiskit_machine_learning/kernels/base_kernel.py
```diff
@@ -20,6 +20,8 @@
 from qiskit import QuantumCircuit
 from qiskit.circuit.library import ZZFeatureMap
```
edoaltamura, line 21:
```suggestion
from qiskit.circuit.library import zz_feature_map
```

qiskit_machine_learning/kernels/base_kernel.py
```diff
@@ -54,6 +56,18 @@ def __init__(self, *, feature_map: QuantumCircuit = None, enforce_psd: bool = Tr
...
+                msg="Passing None as a feature_map is deprecated",
+                version="0.9.0",
+                remedy="Pass a feature map with the required number of qubits to match "
+                "the features. Later any adjustment of the number of qubits will be "
+                "removed from Qiskit as circuits based on BlueprintCircuit, "
+                "like ZZFeatureMap to which this defaults, which could do this, "
+                "have been deprecated.",
+                period="4 months",
+            )
             feature_map = ZZFeatureMap(2)
```
edoaltamura, line 71:
```suggestion
            feature_map = zz_feature_map(2)
```

edoaltamura (changes requested):
Thanks very much for this update Steve! I made some suggestions in the code. The plan seems to remove the class versions of the circuit library instances in 0.9.

qiskit_machine_learning/kernels/base_kernel.py
woodsp-ibm: Now in base_kernel I kept the default, i.e. the default behavior where number of qubits could be adjusted, and noted that while this is the default it is now deprecated. 

Layer I think it probably  needs the default of None removing too so as only a feature map of right size is passed upfront - sure we can maintain the check in there, which is done later to ensure sizes match, but of course adjusting the size, like it tries to do now, which is still possible with such BlueprintCircuits will fail. Removing None will need to be done in all the derived kernel classes too evidently along with updating their docstrings too which all refer to this being the default.
(Now it might be possible to keep None, whereby the user can create the kernel upfront without knowledge of dataset size, and later set the feature_map in, but presently it has no setter and I do not know what the impact of that might be over the set of derived classes - maybe its nothing much, but my assumption was that after deprecation period the datasize would be know, so a correctly sized feature map can be created and passed in to create the chosen kernel)

Does that make sense - main thing for me was to not break peoples code that might rely on this behavior.

woodsp-ibm:
> Thanks very much for this update Steve! I made some suggestions in the code. The plan seems to remove the class versions of the circuit library instances in 0.9.

I wold not say remove, rather deprecate their usage, In examples, tests, notebooks the intent was to use the functions and only the class based ones to make sure things still work. So code still works with both but one should be moving away from the class based ones over to the functions - whether its the Qiskit library ones or the functions I created for the class based library circuits in ML here.

At some point down the road then yes they would be removed, but that's some future version in my mind, not 0.9

qiskit_machine_learning/kernels/base_kernel.py
edoaltamura: I see what you mean! Let's keep in mind for later on.

qiskit_machine_learning/kernels/base_kernel.py
edoaltamura: Same as above. We'll also need to potentially change it in the tutorials when it's the right time to change it here.

edoaltamura (approved):

woodsp-ibm:
I see you merged this, My intent, as per the todo in the PR at the start was to add a release note and update the tutorials to use functions where needed after the fundamental code changes I did had been seen, and agreed with, just in case. I can always do another PR of course, we do need a release note to talk about the deprecations here and it would be better if the tutorials were not using deprecated code. Maybe I'll do the release note one in a PR so as that is at least here too and tutorials can be looked at after.

PR #954, opened 2025-06-16
Pre-minor release - Update VERSION.txt

edoaltamura:
### Summary
Prepares for a minor bug-fix release.


woodsp-ibm:
Looks like #944 should be backported (can still add StableBackport tag to it even though its now merged and it should trigger a backport)

PR #961, opened 2025-07-04
Fix tests to accomodate SciPy 1.16.0 changes

### LINKED ISSUE #957: Nightly actions failing due to Scipy 1.16
It appears nightly actions have been failing for the past couple of nights. First is an error, that appears in multiple tests causing them to fail
```
 _slsqplib.error: Input array gradx must be 1D.
```
and then a mismatched result compared to expected
```
Arrays are not almost equal to 2 decimals

Mismatched elements: 5 / 5 (100%)
Max absolute difference among violations: 1.96224533
Max relative difference among violations: 1.96224533
 ACTUAL: array([-0.96,  0.94,  0.88,  0.78,  0.61])
 DESIRED: array([1., 1., 1., 1., 1.])
```
From an initial quick look it seems scipy 1.16.0 released when the tests started failing (Jun 22). Note the same tests that are presently passing are under Python 3.10. which is still using the earlier 1.15.3 as it seems 1.16.0 is now only for 3.11 and above

edoaltamura:
### Summary
Updates the argument definitions for the optimizers to match those in SciPy 1.16, similarly to https://github.com/qiskit-community/qiskit-algorithms/pull/234

Fixes #957

### DIFF
--- a/docs/lowercase_filter.py
+++ b/docs/lowercase_filter.py
@@ -10,7 +10,7 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
-""" Implements a Lower Case Filter for Sphinx spelling """
+"""Implements a Lower Case Filter for Sphinx spelling"""
 
 from enchant import tokenize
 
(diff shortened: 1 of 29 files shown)


edoaltamura:
I think the changes that need to be made here go beyond those needed for Algorithms, because I'm getting errors like ` _slsqplib.error: Input array gradx must be 1D.` which appears to come from Numpy.

PR #966, opened 2025-07-18
Make tests compatible with SciPy 1.16

edoaltamura:
### Summary

Updates the argument definitions for the optimisers to match those in SciPy 1.16, similarly to https://github.com/qiskit-community/qiskit-algorithms/pull/234

Fixes https://github.com/qiskit-community/qiskit-machine-learning/issues/957


requirements.txt
```diff
@@ -1,6 +1,6 @@
 qiskit>=1.0,<2.0
 numpy>=2.0
-scipy>=1.4,<1.16
+scipy>=1.16
```
woodsp-ibm, line 3: This cannot stay like this, > 1.4 I think was ok as in Python 3.9 1.16 does not exist. If you want to bump the min to say the prior version that still works with 3.9 that should be fine.

PR #968, opened 2025-07-18
Update tutorials

woodsp-ibm:
### Summary

Updates the tutorials to use the newer circuit library functions rather than the former, now deprecated, (BluePrint based) classes.
Addresses the checklist item to do this as found in #945

I will note that the output from these does not do so well - e.g. 02a Training on a Real Dataset where the scores are not as good and 11 Quantum Convolutional Neural Net where it classifies incorrectly one of the horizontal lines. Now these comparisons are against the tutorials as they are checked in where they were last run some time ago. I re-ran them to make sure they all worked so this is updating the output of these notebooks too. Now you can compare these to what is published out in the docs at https://qiskit-community.github.io/qiskit-machine-learning/tutorials/index.html where the notebooks would have been executed to produce the html. You can see they more compare to what I have here rather than the execution outcomes of what is checked in.

### Details and comments

Note: I left this PR as just addressing the circuit library changes over to using the function replacements. I did not attempt to alter the tutorials in any other way, and do not intend to, e.g. trying to improve the outcomes where it seems it did not do as well as the outcomes in the notebooks that are presently checked in. If this is done it should be in separate PR(s).


OkuyanBoga:
Thank you, Steve. It all looks good to me.

OkuyanBoga (approved):

PR #1021, opened 2026-01-12
feat: Add regression-based kernel loss functions (SVRLoss, MSRLoss, M…

srivarenya01:
# Add regression-based kernel loss functions

## Summary
This PR adds regression-based `KernelLoss` implementations to allow `QuantumKernelTrainer` to be used with regression algorithms like `QSVR`.

**Closes #960**

## Changes
- Added `SVRLoss`: Dual objective-based loss for Support Vector Regression
- Added `MSRLoss`: Mean squared error loss  
- Added `MARLoss`: Mean absolute error loss
- Added `HuberLoss`: Robust loss combining MSE and MAE (with configurable `delta`)

## Files Changed
- `qiskit_machine_learning/utils/loss_functions/kernel_loss_functions.py`
- `qiskit_machine_learning/utils/loss_functions/__init__.py`
- `test/utils/loss_functions/test_regression_kernel_loss.py` (new)
- `releasenotes/notes/0.9/add-regression-kernel-loss-functions-960.yaml` (new)

## Testing
- Added unit tests for all four loss functions
- All existing tests pass without regressions

Issue #960, opened 2025-07-03
Required regression-based loss functions for kernel training.


SaashaJoshi:
### What should we add?

I have found that a TrainableFidelityQuantumKernel is incompatible with regression algorithms, like QSVR. The reason seems to be that TrainableFidelityQuantumKernel utilizes QuantumKernelTrainer class that only supports SVCLoss (this class uses SVC internally) for optimizing the quantum kernel parameters. 

Is there a specific reason why Qiskit Machine Learning supports only SVCLoss and no other alternative, regression-based loss functions? Is this a known limitation?

TO-ADD: Regression-based `KernelLoss` functions.

edoaltamura:
@srivarenya01 happy for you to take this up!

srivarenya01:
@edoaltamura , added few loss functions, please review on your free time and let me know if the requirement is satisfied, or any additional changes are required. Thanks for your time and support.

Issue #1011, opened 2025-12-20
Coveralls inactive/disconnected


edoaltamura:
Coverall checks in CI are not being executed and are stuck with the message _"Waiting for status to be reported"_. They should be restored to make sure the unit tests have 90%+ coverage over the modules.

edoaltamura:
Coveralls appears to be working on recent PRs now.

PR #1022, opened 2026-01-29
feat: Add string aliases for regression loss functions

srivarenya01:
- Updates QuantumKernelTrainer to support string aliases ('svr_loss', 'msr_loss', 'mar_loss', 'huber_loss') for the newly added regression loss functions. (Merged in #1021 for #960 )
- updates release notes with usage examples (As Discussed at #1021 )
- adds unit tests for the string aliases

### DIFF
--- a/.pylintdict
+++ b/.pylintdict
@@ -348,6 +348,7 @@ mosca
 mpl
 mprev
 msg
+msr
 multiclass
 multinomial
 multioutput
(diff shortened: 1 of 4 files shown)


srivarenya01:
One of the check (coverage/coveralls) is still pending (#1011 ). Is there any additional step that i have missed, as I dont see any option to merge this PR in the current state.

PR #1037, opened 2026-03-25
Add UKRI-STFC (Hartree Centre) copyright checks

JohnKerf:
### Summary
- Added checks for '(C) Copyright UKRI-STFC (Hartree Centre)' copyright.
- Updated headers check to account for the STFC copyright

#### check_copyright.py
 - STFC copyright year started in 2024 so earliest this copyright can be added is 2024. If there is an IBM copyright that is less than this year then it will not require an STFC copyright until this file has been updated and the latest year would be >= 2024. 
 - Since we need the ability to both add a new STFC copyright line as well as check existing ones there is checks for both inserting a new line and replacing the current line.
 - To avoid duplicating code I created the `get_years` and `new_line` functions.

#### verify_headers.py
- pylint didn't like the large number of return statements within verify_header.py so I had to change the function to reduce the number of returns.
- Again it checks if the file copyright is >=2024 otherwise the STFC header is not necassary until the file has been updated.
- Because some headers can inlcude the STFC copyright while others don't, there needs to be the seperate Apache lines check since the line numbers will be different in each case.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,6 +1,7 @@
 # This code is part of a Qiskit project.
 #
 # (C) Copyright IBM 2021, 2026.
+# (C) Copyright UKRI-STFC (Hartree Centre) 2024, 2026.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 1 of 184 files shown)


OkuyanBoga (approved):
Looks good to me, great work thank you.

PR #1041, opened 2026-04-29
Fix SamplerQNN._postprocess to handle transpiled circuit layouts

### LINKED ISSUE #1040: Unexpected results from SamplerQNN when using pre-transpiled circuits with non-trivial qubit layouts
### Environment

- **Qiskit Machine Learning version**: 0.8.4 **(also verified in 0.9.0 and main branch 1.0.0-dev — same code path)**
- **Python version**:  3.12
- **Operating system**:  Google Colab (Linux)

- **qiskit**: 1.4.5
- **qiskit-ibm-runtime**: 0.43.1
- **qiskit-aer**: 0.17.2
- **Python**: 3.12
- **Hardware**: IBM backends (ibm_kingston) via Qiskit Runtime

**Note:** I checked and the same filter (`if ki < 2**self.num_virtual_qubits`, comment `# keys -> ints, filter to valid range`) is unchanged in 0.9.0 (latest PyPI) and on the main branch (1.0.0-dev).

## Summary

Hi, I'm a QML Intern at the Edge of AI Hub. I'm working on a research paper on cancer classification with QNNs, running variational quantum circuits on real IBM hardware. I ran into some inconsistencies in my results that I spent a while investigating, and I think the issue might be in `SamplerQNN._postprocess`. I could be wrong — I'm still learning the codebase — but I wanted to share what I found in case it's useful. Any clarification would also really help me with my research paper.

## What I observed

I trained 5 QML models (ClassifierSampler and VQC variants, 4 and 7 qubits) on a NISQ simulator, then ran inference on `ibm_kingston`. The workflow for each:

1. Transpile the trained circuit: `transpile(circuit, backend, optimization_level=1)`
2. Create a `SamplerQNN` with the transpiled circuit
3. `qnn.forward(X_test, trained_weights)` → `argmax` → predictions

Some experiments gave good accuracy, others gave near-random results. The weird thing is they all trained well on the simulator. After looking into it, I noticed the results depend on **where the transpiler places the logical qubits**:

| Model | Physical qubit positions | Accuracy |
|---|---|---|
| ClassifierSampler 4Q | [0, 1, 2, 3] | 0.94 |
| ClassifierSampler 7Q | [0, 1, 2, 3, 4, 5, 6] | 0.89 |
| ClassifierSampler 4Q (SI+FE) | [108, 109, 110, 118] | 0.39 |
| VQC 4Q (ZZFeatureMap) | [54, 60, 61, 62] | 0.39 |
| VQC 7Q (ZFeatureMap) | [0, 1, 2, 3, 4, 5, 6] | 0.61 |

Low positions (0–6) → reasonable results. High positions (54+, 108+) → near-random.

## What I found

I recovered the raw measurement counts from my hardware jobs (via `service.jobs()`) and computed predictions from the **same raw counts** in three different ways:

- **Method A**: Parity over all physical qubits (naive baseline)
- **Method B**: Parity over only the logical qubit bits, extracted by their physical positions (proper marginalisation)
- **Method C**: Simulate what `_postprocess` does — convert bitstring to int, keep only if `int(bitstring) < 2**num_virtual_qubits`

| Model | Positions | Method B | Method C | Original | Shots surviving |
|---|---|---|---|---|---|
| CS 4Q | [0,1,2,3] | 0.94 | 0.94 | 0.94 | 622/4096 (15.2%) |
| CS 7Q | [0,1,2,3,4,5,6] | 0.92 | 0.89 | 0.89 | 168/4096 (4.1%) |
| CS 4Q SI+FE | [108,109,110,118] | **0.94** | **0.39** | **0.39** | 16/4096 (0.4%) |
| VQC 4Q ZZ | [54,60,61,62] | **0.54** | **0.39** | **0.39** | 8/4096 (0.2%) |
| VQC 7Q Z | [0,1,2,3,4,5,6] | 0.59 | 0.61 | 0.61 | 211/4096 (5.2%) |

What stands out:

1. **Method C matches the original SamplerQNN results exactly** — not just accuracy, but all precision/recall/F1 per class (35/35 metrics match). So unless I'm misunderstanding the code, Method C should be doing the same thing as `_postprocess`.

2. **Method B recovers much better accuracy** when qubits are at high positions. For SI+FE: 0.94 vs 0.39 from the same raw data.

3. **Even at low positions, most shots get discarded** (85% for 4Q, 96% for 7Q). The code comment says `# keys -> ints, filter to valid range` and the probabilities aren't renormalised after filtering (they're divided by `total_shots` before filtering), so I don't think this is meant to be a noise filter. But even if it were — when qubits are at high positions, the filter ends up discarding the logical qubit signal itself, not just noise from ancillas.

To make sure the Method C match wasn't a coincidence, I ran full classification reports. Here's the Small initialisation + full entanglement experiment (biggest discrepancy):

**Original SamplerQNN result:**
```
              precision    recall  f1-score   support
           0       0.37      0.54      0.44        28
           1       0.42      0.27      0.33        33
    accuracy                           0.39        61
```

**Method C (simulating the filter):**
```
              precision    recall  f1-score   support
           0       0.37      0.54      0.44        28
           1       0.42      0.27      0.33        33
    accuracy                           0.39        61
```

**Method B (marginalising over logical qubits):**
```
              precision    recall  f1-score   support
           0       0.94      0.93      0.93        28
           1       0.94      0.94      0.94        33
    accuracy                           0.94        61
```

I have the full reports for all 5 experiments if that's useful.

### The filter in `_postprocess`

```python
# keys -> ints, filter to valid range
for k, v in counts_i.items():
    ki = _key_to_int(k)
    if ki < 2**self.num_virtual_qubits:
        probs_i[ki] = v / total_shots
```

This keeps only bitstrings whose integer value is below `2**num_virtual_qubits`. That works if logical qubits are at positions [0, 1, ..., n-1], because then the "valid range" [0, 2^n) actually corresponds to the logical qubit subspace. But when the transpiler maps them to higher positions (like [108, 109, 110, 118] in my case), a measurement with a `1` at position 108 gives an integer ≥ 2^108, which is way above the threshold of 2^4 = 16. So valid logical qubit outcomes get filtered outt. Additionally, even when logical qubits are at low positions [0, ..., n-1], any ancilla qubit flipping to 1 due to hardware noise pushes the integer value above the threshold — this is why only 15.2% of shots survive for CS 4Q and 4.1% for CS 7Q in the table above, despite their qubits being at low positions. Since `total_shots` is computed before filtering and there is no renormalization after, the output probabilities do not sum to 1.0.

### Is this handled somewhere else in the code?

I'm not super familiar with the full codebase, so I tried to check whether something else handles this before reaching `_postprocess`:
- `num_virtual_qubits` is correctly set to the logical qubit count (via `circuit.layout._input_qubit_count`), but the filter still uses the integer value, not the positions
- I traced `forward()` → `_forward()` → `sampler.run()` → `_postprocess()` and couldn't find any remapping of bitstrings from physical to logical ordering

### Why this might not show up in tests

The default `QMLSampler` (statevector-based) returns bitstrings already in logical qubit space, so the filter works fine with it. The issue only shows up with a real backend's SamplerV2 where bitstrings reflect physical qubit measurements.

### How I verified the qubit positions

I extracted the physical positions using `qc_transpiled.layout.final_index_layout(filter_ancillas=True)` and verified them against the transpiled circuit structure.

## Related issues

- **#674** and **#819** (closed): about output *shape* after transpilation, not the content of the bitstrings

## Steps to reproduce

```python
from qiskit.circuit import QuantumCircuit, ParameterVector
from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
from qiskit_ibm_runtime import QiskitRuntimeService, SamplerV2
from qiskit_machine_learning.neural_networks import SamplerQNN
import numpy as np

# Simple parameterized circuit
n_qubits = 4
params = ParameterVector('x', n_qubits)
qc = QuantumCircuit(n_qubits)
for i in range(n_qubits):
    qc.ry(params[i], i)
for i in range(n_qubits - 1):
    qc.cx(i, i + 1)

# Transpile to a real backend
service = QiskitRuntimeService()
backend = service.backend("ibm_kingston")
pm = generate_preset_pass_manager(optimization_level=1, backend=backend)
qc_transpiled = pm.run(qc)

# Check where qubits ended up
layout = qc_transpiled.layout.final_index_layout(filter_ancillas=True)
print(f"Logical qubits mapped to: {layout}")
# If any position >= n_qubits, the issue should show up

def parity(x):
    return "{:b}".format(x).count("1") % 2

sampler = SamplerV2(backend=backend)
qnn = SamplerQNN(circuit=qc_transpiled, interpret=parity, output_shape=2, sampler=sampler)

test_input = np.random.rand(1, n_qubits)
result = qnn.forward(test_input, np.array([]))
print(f"Result: {result}")
```

## Diagnostic code

This is the function I used to compare the three methods on raw counts:

```python
from sklearn.metrics import accuracy_score, classification_report
import numpy as np

def diagnostic_3_methods(counts_list, physical_qubits, n_logical_qubits, y_true):
    """
    Compare 3 ways to compute predictions from the same raw counts.
    Method A: parity over ALL physical qubits
    Method B: parity over LOGICAL qubits only (marginalisation by bit position)
    Method C: simulate the SamplerQNN integer-value filter
    """
    preds_A, preds_B, preds_C = [], [], []
    total_shots_all, surviving_shots_all = 0, 0

    for counts in counts_list:
        total_shots = sum(counts.values())
        total_shots_all += total_shots

        prob_A = [0.0, 0.0]
        prob_B = [0.0, 0.0]
        prob_C = [0.0, 0.0]
        surviving = 0

        for bitstring, count in counts.items():
            bs = bitstring.replace(' ', '')

            # Method A: parity on all qubits
            prob_A[bs.count('1') % 2] += count / total_shots

            # Method B: parity on logical qubits only (little-endian)
            n_ones = sum(int(bs[len(bs) - 1 - q]) for q in physical_qubits)
            prob_B[n_ones % 2] += count / total_shots

            # Method C: SamplerQNN filter
            ki = int(bs, 2)
            if ki < 2**n_logical_qubits:
                surviving += count
                prob_C[bin(ki).count('1') % 2] += count / total_shots

        surviving_shots_all += surviving
        preds_A.append(np.argmax(prob_A))
        preds_B.append(np.argmax(prob_B))
        preds_C.append(np.argmax(prob_C))

    survival = surviving_shots_all / total_shots_all * 100
    print(f"Method A (all qubits):        acc = {accuracy_score(y_true, preds_A):.4f}")
    print(f"Method B (logical qubits):    acc = {accuracy_score(y_true, preds_B):.4f}")
    print(f"Method C (SamplerQNN filter): acc = {accuracy_score(y_true, preds_C):.4f}")
    print(f"Shots surviving: {surviving_shots_all}/{total_shots_all} ({survival:.1f}%)")
```

## What I'd expect

If I'm understanding the problem correctly, I think `_postprocess` could marginalise over ancilla qubits using the circuit layout instead of filtering by integer value. I'm not sure this is the right fix, but something like:

```python
physical_qubits = self._circuit.layout.final_index_layout(filter_ancillas=True)
marginalised = {}
for k, v in counts.items():
    ki = int(k) if isinstance(k, int) else int(k, 2)
    logical_bits = tuple((ki >> q) & 1 for q in physical_qubits)
    logical_key = sum(b << i for i, b in enumerate(logical_bits))
    marginalised[logical_key] = marginalised.get(logical_key, 0) + v
counts = marginalised
```

Happy to provide the full raw data or any other info if it helps. And if I'm misunderstanding something about how `SamplerQNN` is supposed to be used with transpiled circuits, I'd really appreciate the clarification — it would help a lot with my research. Thanks!

## What I'd expect

If I'm understanding the problem correctly, I think `_postprocess` could marginalise over ancilla qubits using the circuit layout instead of filtering by integer value. I'm not sure this is the right fix, but something like:

```python
physical_qubits = self._circuit.layout.final_index_layout(filter_ancillas=True)
marginalised = {}
for k, v in counts.items():
    ki = int(k) if isinstance(k, int) else int(k, 2)
    logical_bits = tuple((ki >> q) & 1 for q in physical_qubits)
    logical_key = sum(b << i for i, b in enumerate(logical_bits))
    marginalised[logical_key] = marginalised.get(logical_key, 0) + v
counts = marginalised
```

Happy to provide the full raw data or any other info if it helps. And if I'm misunderstanding something about how `SamplerQNN` is supposed to be used with transpiled circuits, I'd really appreciate the clarification — it would help a lot with my research. Thanks!

sorayapnblm:
### Summary

Fixes [#1040](https://github.com/qiskit-community/qiskit-machine-learning/issues/1040)

`SamplerQNN._postprocess` silently discarded valid measurements and produced unnormalized probabilities when processing results from transpiled circuits. This PR replaces the integer value filter with layout-based marginalization over ancilla qubits.

### Details and comments

**The bug:**
The previous code filtered measurements by integer value (`ki < 2**num_virtual_qubits`) and divided each count by the total number of shots (computed before filtering). This causes three related problems:

- High qubit positions: When the transpiler maps logical qubits to higher positions (e.g. [5, 7] on an 8-qubit backend), valid bitstrings have integer values that exceed the threshold and are silently discarded, producing near-zero probabilities.
- Ancilla noise at any position: Even when logical qubits occupy low positions [0, 1, ..., n-1], any ancilla qubit flipping to 1 due to hardware noise pushes the bitstring's integer value above the threshold. In our experiments (see [#1040](https://github.com/qiskit-community/qiskit-machine-learning/issues/1040)), this discarded  85–96% of shots that carried valid logical qubit information.
- As a result, the output probabilities do not sum to 1.0 — because `total_shots` is computed before filtering but only surviving shots contribute to the numerator, and there is no renormalization step after filtering.

**The fix:**
Uses `circuit.layout.final_index_layout(filter_ancillas=True)` to identify the physical positions of logical qubits, then extracts and recombines only those bits — marginalizing over all ancilla qubits. Every shot contributes to the output regardless of ancilla state, so probabilities naturally sum to 1.0. When no layout is present (circuit was not transpiled), the original behavior is preserved as no ancilla qubits exist in that case.

**Changes:**
- `qiskit_machine_learning/neural_networks/sampler_qnn.py`: layout-aware marginalization in `_postprocess`
- `test/neural_networks/test_sampler_qnn.py`: regression test forcing logical qubits to positions [5, 7] with `SamplerV2` to reproduce the bug
- Release note added via reno

**Checks:**
- [x] `make test` — 955 passed, 0 failed
- [x] `make lint` — 10/10
- [x] `make style` — passes
- [x] Tests added to cover changes
- [x] Documentation updated accordingly
- [x] Read the CONTRIBUTING document
- [x] Copyright headers include 2026


sorayapnblm:
No code changes in this update — just improved documentation (PR description, release note, and test docstring) to better explain the bug.

While reviewing my own PR, I realized I had only described one aspect of the bug: high qubit positions. The integer value filter (`ki < 2**num_virtual_qubits`) actually causes two problems:

1. High qubit positions: When the transpiler maps logical qubits to high physical positions, valid bitstrings exceed the threshold and are discarded.
2. Ancilla noise: Even when logical qubits are at low positions ([0, ..., n-1]), any ancilla qubit flipping to 1 due to hardware noise pushes the integer value above the threshold, discarding that shot. In our experiments, this discarded 85–96% of valid shots.

As a result, the output probabilities do not sum to 1.0, since `total_shots` is computed before filtering with no renormalization after.

The fix already handles all of this: marginalization over ancilla qubits means every shot contributes regardless of ancilla state, so probabilities naturally sum to 1.0. I just hadn’t explained it clearly enough in the original PR description.

OkuyanBoga (changes requested):
The fix makes sense, and the new test covers the issue well. Thank you very much for your contribution.

There is only a small issue: I’d avoid using ki and adding it to pylint; a clearer name would be better. 

Something like key_int or measurement_int would be clearer. Apart from that, the fix and release note look good to me.

sorayapnblm:
@OkuyanBoga changed ki to key_int as requested

sorayapnblm:
@OkuyanBoga CI failure is the flaky test_VQC_33 (stochastic test, assertGreater(score, 0.5) got 0.2) — unrelated to this PR. Our test passes. Happy to re-run if needed.

OkuyanBoga (approved):

OkuyanBoga:
Thank you for your contribution

sorayapnblm:
@OkuyanBoga Happy to help

PR #1044, opened 2026-05-13
Add Phase of Matter quantum dataset generator

rogue-infinity:
## Summary

- Adds `phase_of_matter_data()` to `qiskit_machine_learning/datasets/` as a new sub-package `phase_of_matter/`
- Implements four spin-chain Hamiltonians from [Bermejo-Vega et al., arXiv:2408.12739](https://arxiv.org/abs/2408.12739):
  - `"heisenberg"` - Bond-alternating XXX Heisenberg: phases *trivial* / *topological*
  - `"haldane"` - Haldane chain: phases *antiferromagnetic* / *paramagnetic* / *spt*
  - `"annni"` - Axial Next-Nearest-Neighbor Ising: phases *ferromagnetic* / *paramagnetic* / *floating* / *antiphase*
  - `"cluster"` - Cluster Hamiltonian (periodic BC): phases *haldane* / *ferromagnetic* / *antiferromagnetic* / *trivial*
- Ground states computed via `scipy.sparse.linalg.eigsh` (exact diagonalization, default). Optional VQE pathway available via `backend=` argument for hardware-experiment workflows
- Follows the existing dataset API convention (`training_size`, `test_size`, `one_hot`, `formatting`, `class_labels`, `include_sample_total`, `seed`, `backend`)

## Files changed

| File | Description |
|---|---|
| `qiskit_machine_learning/datasets/phase_of_matter/__init__.py` | Package entry point |
| `qiskit_machine_learning/datasets/phase_of_matter/phase_of_matter.py` | Public `phase_of_matter_data()` API |
| `qiskit_machine_learning/datasets/phase_of_matter/_base.py` | Shared utilities: `pauli_term`, exact diag, VQE |
| `qiskit_machine_learning/datasets/phase_of_matter/_heisenberg.py` | Heisenberg Hamiltonian + phase sampler |
| `qiskit_machine_learning/datasets/phase_of_matter/_haldane.py` | Haldane Hamiltonian + phase sampler |
| `qiskit_machine_learning/datasets/phase_of_matter/_annni.py` | ANNNI Hamiltonian + phase sampler |
| `qiskit_machine_learning/datasets/phase_of_matter/_cluster.py` | Cluster Hamiltonian + phase sampler |
| `qiskit_machine_learning/datasets/__init__.py` | Export `phase_of_matter_data` |
| `test/datasets/test_phase_of_matter.py` | 55-test suite |

## Test plan

- [x] `pytest test/datasets/test_phase_of_matter.py`, 55 tests covering Hermiticity of all Hamiltonians, ground-state normalization, eigenstate residuals, phase label coverage for all models, API shape contracts, seed reproducibility, and error handling
- [x] Black / pylint / tox lint checks

## Usage example

```python
from qiskit_machine_learning.datasets import phase_of_matter_data

x_train, y_train, x_test, y_test = phase_of_matter_data(
    training_size=10, test_size=5, n=4, model="heisenberg", seed=0
)
# x_train.shape → (10, 16), y_train.shape → (10, 2)
```


rogue-infinity:
@OkuyanBoga @adekusar-drl @oscar-wallis 
Hi guys, could you please check the failing test case of ML Unit Tests on Windows 3.10. As far as I can see from the logs, it fails a VQC test where it achieves an accuracy of 0.2 on a dataset it creates on its own without referencing this new generator anywhere. Every other platform (Ubuntu × 4 Python versions, macOS × 2 Python versions) passed 

Awaiting your response on this!

rogue-infinity:
@OkuyanBoga, all the tests passed successfully, could you please go through the PR and let me know any modifications/ suggestions/ additions need to be made?

qiskit_machine_learning/datasets/phase_of_matter/_base.py
```diff
@@ -0,0 +1,120 @@
...
+        Complex numpy array of shape ``(2**n,)`` -- the normalized ground state.
+    """
+    mat = hamiltonian.to_matrix(sparse=True).astype(complex)
+    _, vecs = scipy.sparse.linalg.eigsh(mat, k=1, which="SA")
+    return _canonicalize_phase(vecs[:, 0])
+
+
+def get_ground_state_vqe(
+    hamiltonian: SparsePauliOp,
+    backend,  # pylint: disable=unused-argument
```
OkuyanBoga, line 79: `backend` is currently ignored after selecting the VQE path. Since passing a backend makes users expect that backend to be used, either the implementation should pass it through or the public docs should clearly state that VQE currently always uses `StatevectorEstimator`.

qiskit_machine_learning/datasets/phase_of_matter/_heisenberg.py
```diff
@@ -0,0 +1,86 @@
...
+        j1: Coupling constant on even bonds (:math:`J_1 \geq 0`).
+        j2: Coupling constant on odd bonds (:math:`J_2 \geq 0`).
+
+    Returns:
+        SparsePauliOp for the Hamiltonian on *n* qubits.
+    """
+    terms: list[SparsePauliOp] = []
+    for i in range(n - 1):
+        # i is 0-indexed; the paper is 1-indexed, so even bond = i % 2 == 0
+        j = j1 if (i % 2 == 0) else j2
```
OkuyanBoga: Please ensore implementation and documentation to match, the docstring states that J1 should be used for even bonds and J2 for odd bonds, it assigns J1 when i % 2 == 0, which corresponds to the first, third, fifth, etc. bonds in 1-indexed terms.

qiskit_machine_learning/datasets/phase_of_matter/phase_of_matter.py
```diff
@@ -0,0 +1,271 @@
...
+    else:
+        label_names = list(default_labels)
+
+    rng = np.random.default_rng(seed)
+
+    # ceil ensures every class gets at least the requested count even when
+    # training_size / test_size are not divisible by n_classes.
+    n_per_class_train = math.ceil(training_size / n_classes)
+    n_per_class_test = math.ceil(test_size / n_classes)
+    n_per_class = n_per_class_train + n_per_class_test
```
OkuyanBoga: I think this can make returned data unbalanced when sizes are not divisible by n_classes

qiskit_machine_learning/datasets/phase_of_matter/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+* :mod:`._annni`      — Axial Next-Nearest-Neighbor Ising (ANNNI) model
+* :mod:`._cluster`    — Cluster Hamiltonian
+
+The :func:`phase_of_matter_data` function is the single public entry point.
+
+.. currentmodule:: phase_of_matter
+
+.. autosummary::
+   :toctree: ../stubs/
+   :nosignatures:
```
OkuyanBoga: autosummary options should be indented under the directive or Sphinx may not parse them correctly

OkuyanBoga (changes requested):
Thank you for this PR and apologies for a bit late review. I believe it is a great addition however there are few points I see as an issue. Maybe, in the future, we can extend the backend options for the researchers, it would be valuable.

qiskit_machine_learning/datasets/phase_of_matter/_base.py
rogue-infinity: Updated the `backend` parameter description in `phase_of_matter_data()` to : *"The current VQE implementation uses `StatevectorEstimator` from `qiskit.primitives` unconditionally, regardless of the backend object passed. The `backend` argument is accepted for API consistency and is reserved for future hardware integration."* Also added this note directly inside the `.. warning::` block so it's visible at the top of the docstring.

qiskit_machine_learning/datasets/phase_of_matter/_heisenberg.py
rogue-infinity: Updated the docstring to show in the code. The loop runs 0-indexed (`i = 0, 1, 2, ...`), so `not i % 2` is `True` for `i = 0, 2, 4, ...` which correspond to **odd-indexed** bonds in 1-indexed notation (bond 1, 3, 5, ...). The docstring is now: *J1 on odd-indexed bonds (1-indexed: 1, 3, 5, ...) and J2 on even-indexed bonds (1-indexed: 2, 4, 6, ...)*.

qiskit_machine_learning/datasets/phase_of_matter/phase_of_matter.py
rogue-infinity: Replaced the `math.ceil` + end-trim approach with exact per-class counts using a base + remainder distribution: each class gets `total // n_classes` samples, and the first `total % n_classes` classes receive one extra. This means `sum(counts) == total` exactly and `max - min <= 1`. The old trim (`train_states[:training_size]`) is removed since counts are now exact. Added a regression test (`test_balanced_classes_non_divisible`)

qiskit_machine_learning/datasets/phase_of_matter/__init__.py
rogue-infinity: removed the `.. currentmodule::` and `.. autosummary::` block from `phase_of_matter/__init__.py` entirely. `phase_of_matter_data` is already listed in the parent `datasets/__init__.py` autosummary (matching the pattern of `ad_hoc_data` and `entanglement_concentration_data`), so the subpackage `__init__.py` only needs a prose module docstring.

rogue-infinity:
@OkuyanBoga thanks for your feedback!, I have made the necessary changes. Please do have a look and let me know if anything else needs to be changed

OkuyanBoga (approved):
Looks good to me, thank you

PR #1045, opened 2026-05-16
fix: decompose BlueprintCircuits before re-parametrization to fix QPY…

### LINKED ISSUE #1042: ComputeUncompute + FidelityQuantumKernel + SamplerV2 fails on IBM hardware with QPY parameter binding error
### Environment

- **Qiskit Machine Learning version**: 0.8.3
- **Python version**: 3.11
- **Operating system**: macOS Tahoe 26.1

Additional relevant packages:

- **qiskit**: 1.4.3
- **qiskit-ibm-runtime**: 0.43.1
- **qiskit-algorithms**: 0.4.0
- **qiskit-aer**: 0.17.1
- **numpy**: 2.1.2
- **scipy**: 1.14.1
- **scikit-learn**: 1.5.2

### What is happening?

I am trying to run a standard quantum kernel workflow on real IBM hardware using:

`SamplerV2 -> ComputeUncompute -> FidelityQuantumKernel -> PegasosQSVC`

The workflow fails during `PegasosQSVC.fit()` with a QPY parameter binding error related to the parameters of a standard `ZZFeatureMap`.

This also happens with the default feature maps provided by Qiskit, for example:

- `ZZFeatureMap(..., entanglement="linear")`
- `ZZFeatureMap(..., entanglement="circular")`
- `ZZFeatureMap(..., entanglement="full")`

So it does not seem to be caused by a custom circuit.

The error is:

```text
RuntimeJobFailureError: Unable to retrieve job result.
QpyError: Cannot bind following parameters not present in expression:
{Symbol { name: "x", uuid: ..., index: Some(0), vector: Some(...) }}
```

The traceback goes through:

```text
PegasosQSVC.fit()
FidelityQuantumKernel.evaluate()
ComputeUncompute._call()
SamplerV2 job.result()
```

I have seen similar issues reported before, especially around `ComputeUncompute`, transpilation/pass manager support, and hardware execution with Runtime primitives. For example, issues such as `qiskit-algorithms#194` and `qiskit-ibm-runtime#1519` seem related.

My understanding is that the problems discussed in those issues should already be fixed or supported for the versions I am using, especially because `ComputeUncompute` now accepts pass manager / transpilation support. However, with the environment listed above, the error still appears.

A minimal example is:

```python
import numpy as np

from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
from qiskit_ibm_runtime import QiskitRuntimeService, SamplerV2
from qiskit_machine_learning.state_fidelities import ComputeUncompute
from qiskit_machine_learning.kernels import FidelityQuantumKernel
from qiskit_machine_learning.algorithms import PegasosQSVC

service = QiskitRuntimeService()
backend = service.backend("ibm_kingston")

X_train = np.array([
    [0, 0, 0, 0, 1, 0],
    [0, 1, 0, 1, 1, 0],
    [1, 0, 1, 0, 0, 1],
    [1, 1, 1, 1, 0, 1],
], dtype=float)

y_train = np.array([0, 0, 1, 1])

feature_map = ZZFeatureMap(
    feature_dimension=X_train.shape[1],
    reps=1,
    entanglement="linear",
)

pass_manager = generate_preset_pass_manager(
    backend=backend,
    optimization_level=0,
)

sampler = SamplerV2(backend)
sampler.options.default_shots = 128

fidelity = ComputeUncompute(
    sampler=sampler,
    pass_manager=pass_manager,
)

qkernel = FidelityQuantumKernel(
    feature_map=feature_map,
    fidelity=fidelity,
)

clf = PegasosQSVC(
    quantum_kernel=qkernel,
    C=1000,
    num_steps=10,
    seed=12345,
)

clf.fit(X_train, y_train)
```

The output is:

```text
RuntimeJobFailureError: Unable to retrieve job result.
QpyError: Cannot bind following parameters not present in expression:
{Symbol { name: "x", uuid: ..., index: Some(0), vector: Some(...) }}

qiskit_machine_learning.exceptions.AlgorithmError: 'Sampler job failed!'
```

The same type of error also happens when changing the feature map to:

```python
ZZFeatureMap(feature_dimension=6, reps=1, entanglement="circular")
ZZFeatureMap(feature_dimension=6, reps=1, entanglement="full")
```

### What should happen?

The standard workflow

`SamplerV2 -> ComputeUncompute -> FidelityQuantumKernel -> PegasosQSVC`

should work on real IBM hardware with a standard parametrized `ZZFeatureMap`.

I would expect `clf.fit(X_train, y_train)` to complete successfully, submitting the required fidelity circuits through the Runtime Sampler and returning a trained `PegasosQSVC` model.

### Any suggestions?

It looks like the problem may be related to how `ComputeUncompute` builds, transpiles, or serializes the fidelity circuits internally when using parametrized feature maps with `SamplerV2`.

My guess is that when the feature map and its inverse are combined internally, the parameters named `x[0]`, `x[1]`, etc. may end up with different UUIDs or parameter objects that QPY cannot bind correctly before submitting the Runtime job.

I have seen previous issues that look related, but I believe those problems should theoretically already be solved for the versions I am using:

- `qiskit-machine-learning==0.8.3`
- `qiskit-ibm-runtime==0.43.1`
- `qiskit==1.4.3`
- `qiskit-algorithms==0.4.0`

As a workaround, I can manually build parameter-free fidelity circuits for each kernel entry and send them directly to `SamplerV2`, but this bypasses the standard `FidelityQuantumKernel` / `ComputeUncompute` API.

Could you clarify whether `FidelityQuantumKernel + ComputeUncompute + SamplerV2` is currently expected to work on IBM hardware with these versions? If yes, is there a recommended construction pattern for the feature map, pass manager, sampler, or session to avoid this QPY parameter binding error?

92meharali:
… UUID mismatch (#1042)

When a BlueprintCircuit (e.g. ZZFeatureMap) is passed to ComputeUncompute, its internal gate-level symbolic expressions may use the original parameter UUIDs from the blueprint's own ParameterView even after assign_parameters() is called on the outer circuit. This is because BlueprintCircuits use lazy evaluation via _build(), so the gate-level expressions are not fully materialized until the circuit is iterated or decomposed.

When IBM Runtime serializes the SamplerV2 PUB via QPY for remote execution on real hardware, the expression tree on the backend contains the original blueprint UUID while the BindingsArray keys hold the new re-parametrized UUID, causing:

    QpyError: Cannot bind following parameters not present in expression:
    {Symbol { name: "x", uuid: ..., index: Some(0), vector: Some(...) }}

Fix: call circuit.decompose() on both input circuits before assigning the new ParameterVectors. This forces full materialization of the circuit's gate-level instructions so that all parameter UUID references are consistent before re-parametrization, QPY serialization, and binding.

Fixes: #1042

## Summary
- Decompose BlueprintCircuits (e.g. `ZZFeatureMap`) before re-parametrization in `BaseStateFidelity._construct_circuits`
- Fixes `QpyError: Cannot bind following parameters not present in expression` when using `ComputeUncompute` + `SamplerV2` on IBM Runtime
- Adds regression test and release note

Fixes #1042

## Test plan
- [x] `pytest test/state_fidelities/test_compute_uncompute.py`
- [ ] CI green on PR
- [ ] (Optional) Reproduce original issue on IBM hardware with `ZZFeatureMap` + `PegasosQSVC`


oscar-wallis:
Hi @92meharali , could you check over our new [AI Policy](https://github.com/qiskit-community/qiskit-machine-learning/blob/main/AI_POLICY.md) and confirm you have read and understood it? In line with the new AI Policy, please also disclose whether or not you used AI in this PR and if so, which? A similar message is being sent to all open PRs.

oscar-wallis:
Hi @92meharali, unfortunately our testing of your code shows that it does not seem to fix the issue, additionally, this issue is fixed on qiskit>=2.2. As such we're creating our own quick fix for versions qiskit 2.0 and 2.1, and we will close this PR. Thanks for your work.

92meharali:
thanks @oscar-wallis I understand!

PR #1055, opened 2026-07-30
Address "QNN warning seems out-of-place/unnecessary/different" #967

### LINKED ISSUE #967: QNN warning seems out-of-place/unnecessary/different
I have just started changing tutorials to use the new circuit library functions to create library circuits rather than the deprecated (Blueprint based) classes. This was a work item in #945.

In changing things I notice this warning with QNN use

> No gradient function provided, creating a gradient function. If your Estimator requires transpilation, please provide a pass manager.

which you can see in the published docs as these are the tutorials re-run using the latest code. E.g here https://qiskit-community.github.io/qiskit-machine-learning/tutorials/02_neural_network_classifier_and_regressor.html#Classification-with-an-EstimatorQNN and in a few other places in same tutorial and elsewhere when using QNNs

2 observations/comments:

1. Why do we warn about no gradient function, i,e that it;s using its internally create default. It never warned about that before - and still does not in the present code if you use a V1 primitive.

2. Why the transpiler warning too. In the case of the tutorial its annoying to see these since they are irrelevant in this case. I do not see any other algorithms warning, when using an Estimator or Sampler, that pass manager might need to be set - so why do we do so in the QNNs?

m-meller:
### Summary
Addresses #967. 
QNNs were reporting warnings when no gradient function was provided and they resorted to defaults. This is inconsistent with other algorithms and the wider code-base. Additionally, a transpiler warning was emitted, which again should not be reported at this level.

The result was that users could see warnings in situations where these warnings are not relevant (e.g., tutorials).

If users pass a primitive that requires transpilation, it should be their responsibility to provide associated pass manager. This is part of EstimatorQNN and SamplerQNN documentation. As such, the warnings should be removed.

### Details and comments
- Removed the warnings in `estimator_qnn.py` and `sampler_qnn.py`.
- While the change does not affect the code logic, tests were run and passed locally.

### DIFF
--- a/qiskit_machine_learning/neural_networks/estimator_qnn.py
+++ b/qiskit_machine_learning/neural_networks/estimator_qnn.py
@@ -15,7 +15,6 @@
 
 from __future__ import annotations
 
-import logging
 from copy import copy
 from typing import Sequence
 
@@ -34,8 +33,6 @@ from ..gradients import (
     ParamShiftEstimatorGradient,
 )
 from .neural_network import NeuralNetwork
-
-logger = logging.getLogger(__name__)
 
 
 class EstimatorQNN(NeuralNetwork):
@@ -177,11 +174,6 @@ class EstimatorQNN(NeuralNetwork):
 
         # set gradient
         if gradient is None:
-            if pass_manager is None:
-                logger.warning(
-                    "No gradient function provided, creating a gradient function."
-                    " If your Estimator requires transpilation, please provide a pass manager."
-                )
             gradient = ParamShiftEstimatorGradient(
                 estimator=self.estimator, pass_manager=pass_manager
             )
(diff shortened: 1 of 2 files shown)


oscar-wallis (approved):
I remember these warnings being added when PrimitivesV2 were added and we wanted to make sure people were using transpilers etc. Changes made fix this issue as it isn't consistent with other classes and warnings don't make sense in tutorials.

oscar-wallis:
Hi @m-meller, could you check over our new [AI Policy](https://github.com/qiskit-community/qiskit-machine-learning/blob/main/AI_POLICY.md) and confirm you have read and understood it? In line with the new AI Policy, please also disclose whether or not you used AI in this PR and if so, which? A similar message is being sent to all open PRs.

m-meller:
Hi @oscar-wallis. I can confirm, I have reviewed and accept project's AI policy and no AI has been used for this PR and associated commits. 

Thanks

Issue #850, opened 2024-11-12
Merge IEEE2024 tutorials into `main`


edoaltamura:
### What should we add?

Merge the notebooks from the tutorial presented at IEEE QW2024 in Montreal. These are stored in the branch [IEEE2024](https://github.com/qiskit-community/qiskit-machine-learning/tree/IEEE2024) and should be updated to use V2 primitives in line with the 0.8 release.

oscar-wallis:
Hi, yeah happy to do this, they need polishing which is waiting on me to do.

PR #1059, opened 2026-08-10
Add sub-sampling quantum kernel logic and tutorial

JohnKerf:
### Summary

PR related to issue: #850 

I have implemented the sub-sampling quantum kernel method by introducing:
- ``BatchedSubKernelSVCLoss`` - handles the batching of smaller sub-kernels and computes average SVC loss.
- ``CovariantFeatureMap`` - quantum circuit used to demonstrate the sub-kernel method in the tutorial.
- I have also introduced new tests for these.
- A new tutorial ``14_subsampling_quantum_kernel`` and accompanying ``dataset_graph7.csv`` dataset.

The tutorial is based on the original IEEE tutorial https://github.com/qiskit-community/qiskit-machine-learning/tree/IEEE2024/IEEE2024/The_Sub-Kernel_Method with some corrections to the logic.

### Details and comments

The tutorial has been simplified where possible. In particular, ``the BatchedSVCLoss`` class has become the feature ``BatchedSubKernelSVCLoss`` and is imported instead of coded in the tutorial. ``DataBatcher`` has been removed from the tutorial and batching logic is implemented within the loss class. The tutorial uses SPSA throughout, rather than giving options for multiple optimisers. The callback function now only records the loss. Finally, plotting and other functions have been condensed/simplified to improve readability and flow.

While reproducing the original tutorial, I noticed that the runtime for a sub-kernel size of 32 was approximately the same as for 16. Comparing the tutorial logic against the accompanying https://arxiv.org/abs/2401.02879 paper identified two key issues:

1) The paper states:
"Note that each time a new subset is selected, it is composed of previously unselected points until the entire dataset is exhausted, at which point previously selected points may be used again.".  
The original DataBatcher did not implement this correctly and could discard remaining data points when the dataset size was not divisible by the sub-kernel size. 
2) The original logic could skip remaining sub-kernels when there were fewer than minibatch_size sub-kernels left in an epoch. This was because the index could be reset before those remaining sub-kernels were evaluated. Additionally, when this occured the loss was still being divided by the requested minibatch size resulting in inaccurate results.

To address these issues I removed ``DataBatcher`` completely and implemented ``BatchedSubKernelSVCLoss``.
BatchedSubKernelSVCLoss instead contains individual class sampling pools using self.class_idxs and self.unused_idxs. Batching is now handled by the ``_batch_subkernels`` method within the class. Samples are drawn according to the class ratios, and each class pool is only refreshed once that class has been exhausted.

Since SVC requires both classes to be represented by at least one sample, the number of samples per class is clipped to ensure that every sub-kernel contains at least one sample from each class.

**AI Disclosure** - *In this work ChatGPT was used as a tool to help debug issues and suggest potential fixes. I was a constant human in the loop and all code was implemented and checked myself and I fully understand how it works. ChatGPT was also used to help create tests but again I fully checked and implemented the tests myself and ChatGPT was simply used as a tool.*


JohnKerf:
The failed test isnt related to my code changes

PR #1060, opened 2026-08-12
Fix ibm runtime issue caused by reparameterization

JohnKerf:
### Summary
PR related to #1042 
I tested the original PR https://github.com/qiskit-community/qiskit-machine-learning/pull/1045 fix and it did not work so I have created this new PR.

- Issue is resolved for Qiskit versions >= 2.2
- PR related to Qiskit fix: https://github.com/Qiskit/qiskit/pull/13727
- Issue occurs when a parameter name in reparameterisation is identical to the parameter name used in the original circuit
- This causes a runtime error: `RuntimeJobFailureError: 'Unable to retrieve job result. Error code 3211; Job not valid. Circuit deserialization error. \'Cannot bind following parameters not present in expression:`
- Fix involves changing reparameterization names from x and y to x_fidelity and y_fidelity

### Details and comments

Although the qiskit release containing the PR for the fix is included in version 2.1.0 the compatible ibm-runtime version 0.45.0 still causes the error and therefore (at least in my testing) it is only fixed for qiskit version >= 2.2 which is compatible with the working qiskit_ibm_runtime 0.46.0. Since qiskit-machine-learning is for qiskit >= 2.0 then this workaround is necassary.

- In #1042 the ZZFeatureMap creates a parameter vector with names `x[0], x[1]....`
- When we pass this into the `ComputeUncompute`/`BaseStateFidelity`,` BaseStateFidelity._construct_circuits` previously constructed `parameters_1 = ParameterVector("x", circuit_1.num_parameters)`  which would create the same names as the `ZZFeatureMap` thereby causing the issue.
- By renaming the reparameterized names in ` BaseStateFidelity._construct_circuits` from x and y to x_fidelity and y_fidelity we circumvent this issue however it is still possible for a user to recreate the error if they name a parameter x_fidelity or y_fidelity. 
- Due to this only being an issue for versions < 2.2 and it only occuring in the unlikely event that the parameters are named the same, the decision was made to not enforce that the names are unique since this would introduce unnecessary  computational cost.  Instead we provide a more descriptive error message that triggers under the specific conditions for the error.

I have verified the code works and the appropiate error messages appear for the given conditions using:
device: ibm_marrakesh device
qiskit: >= 2.0
qiskit_machine_learning: 1.0.0
qiskit_ibm_runtime: >= 0.45

**AI Disclosure** - *Minimal use of ChatGPT was used in this work. All code and checks were done by myself and AI was mainly used for research around what the error message meant and helped with finding the qiskit PR related to the fix.*

### DIFF
--- a/qiskit_machine_learning/state_fidelities/compute_uncompute.py
+++ b/qiskit_machine_learning/state_fidelities/compute_uncompute.py
@@ -16,8 +16,11 @@ Compute-uncompute fidelity interface using primitives
 
 from __future__ import annotations
 
-from collections.abc import Sequence
 from copy import copy
+from collections.abc import Sequence
+
+from importlib.metadata import version
+from packaging.version import Version
 
 from qiskit import QuantumCircuit
 from qiskit.primitives import BaseSamplerV2, PrimitiveResult, SamplerPubResult
@@ -193,6 +196,24 @@ class ComputeUncompute(BaseStateFidelity):
         try:
             result = job.result()
         except Exception as exc:
+            message = str(exc)
+
+            backend = job.backend()
+            is_hardware = backend is not None and not backend.configuration().simulator
+
+            if (
+                "Cannot bind following parameters not present in expression" in message
+                and Version(version("qiskit")) < Version("2.2")
+                and is_hardware
+            ):
+                raise AlgorithmError(
+                    "Sampler job failed due to a circuit deserialization error. "
+                    "This could be caused by known issue. "
+                    "Ensure no circuit paramters are named x_fidelity or y_fidelity. "
+                    "For more information: "
+                    "https://github.com/qiskit-community/qiskit-machine-learning/pull/1060."
+                ) from exc
+
             raise AlgorithmError("Sampler job failed!") from exc
 
         quasi_dists = _post_process_v2(result, num_virtual_qubits)
--- a/qiskit_machine_learning/state_fidelities/compute_uncompute.py
+++ b/qiskit_machine_learning/state_fidelities/compute_uncompute.py
@@ -20,7 +20,6 @@ from copy import copy
 from collections.abc import Sequence
 
 from importlib.metadata import version
-from packaging.version import Version
 
 from qiskit import QuantumCircuit
 from qiskit.primitives import BaseSamplerV2, PrimitiveResult, SamplerPubResult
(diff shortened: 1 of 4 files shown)


qiskit_machine_learning/state_fidelities/compute_uncompute.py
```diff
@@ -16,8 +16,11 @@
 
 from __future__ import annotations
 
-from collections.abc import Sequence
 from copy import copy
+from collections.abc import Sequence
+
+from importlib.metadata import version
+from packaging.version import Version
```
oscar-wallis, line 23: I'd like to avoid adding the packaging dependency. I think this can be done with importlib alone.

qiskit_machine_learning/state_fidelities/compute_uncompute.py
```diff
@@ -193,6 +196,24 @@ def _call(
         try:
             result = job.result()
         except Exception as exc:
+            message = str(exc)
+
+            backend = job.backend()
```
oscar-wallis, line 201: job.backend() will fail for a generic BackendSamplerV2. I'd suggest using something like getattr(job, "backend", None). If this doesn't work, then we know the error isn't caused by this anyway and can through the regular algorithm error.

qiskit_machine_learning/state_fidelities/compute_uncompute.py
```diff
@@ -193,6 +196,24 @@ def _call(
...
+
+            if (
+                "Cannot bind following parameters not present in expression" in message
+                and Version(version("qiskit")) < Version("2.2")
+                and is_hardware
+            ):
+                raise AlgorithmError(
+                    "Sampler job failed due to a circuit deserialization error. "
+                    "This could be caused by known issue. "
+                    "Ensure no circuit paramters are named x_fidelity or y_fidelity. "
```
oscar-wallis, line 212:
```suggestion
                    "Ensure no circuit parameters are named x_fidelity or y_fidelity. "
```

qiskit_machine_learning/state_fidelities/compute_uncompute.py
```diff
@@ -193,6 +196,24 @@ def _call(
...
             result = job.result()
         except Exception as exc:
+            message = str(exc)
+
+            backend = job.backend()
+            is_hardware = backend is not None and not backend.configuration().simulator
+
+            if (
+                "Cannot bind following parameters not present in expression" in message
+                and Version(version("qiskit")) < Version("2.2")
```
oscar-wallis, line 206: Can the check be 2.0 <= version < 2.2. This would trigger for any qiskit below qiskit 2.2 and we only support 2.0 and above.

oscar-wallis (changes requested):
Added some comments here. This will need a release note for the bug fix / making the bug harder to fix + our rationale. Also, this could use a small test that does 1. make sure the internal params are called x/y_fidelity and that if you submit a param vector called x/y_fidelity, the correct error is thrown. That could be overkill - what do you think?

qiskit_machine_learning/state_fidelities/compute_uncompute.py
JohnKerf: @oscar-wallis I could do something like this? 
`tuple(map(int,version("qiskit").split("."))) < (2,2,0)`

